PR #2, opened 2015-04-20
Added Basic Item Documentation.

sixd:


LexManos:
Your tutorial is out of date. We are on 1.8.

cpw:
I have to agree with Lex. Explain, don't just dump a big code blob..

cpw:
This is more what I would expect on a wiki, than in curated documentation.

thejereman13:
As soon as I have some free time today, I'll try to write up some documentation on the necessary classes/functions used in creating and registering an item. Are you guys wanting something along the lines of "here's what functions to use, here's how they work"?

cpw:
Yes, explain how an item is structured, what you need to override and why. Remember that 1.8 is our official version too ;-)

sixd:
Oh I am so sorry I did not get the fact that this was for 1.8 I am so sorry.

cpw:
Its fine. Most documentation is applicable for 1.7.10 AND 1.8, and I'll be building something to discuss the update. But any sample code should start with 1.8..

PR #6, opened 2015-04-20
Add mkdocs.yml, fixes section display names

tterrag1098:
The yml file will need to be updated as pages are added, it can control the order and "proper" name of every file.


tterrag1098:
Hopefully the format is self-evident, you can nest pages using extra commas, e.g. `Networking, Home`

tterrag1098:
For future reference: http://www.mkdocs.org/user-guide/configuration/

cpw:
Thanks

PR #10, opened 2015-05-02
[WIP] Rendering documentation

bobtwinkles:
I'm creating this issue now so nobody does duplicate work, and more importantly to get some early feedback.

The following is the list of things I hope to cover, roughly in the order I plan on writing them.
  - [x] `ModelBlock` machinery: how the game turns JSON in to blocks
  - [ ] `TileEntitySpecialRenderer` machinery: how the game handles animated blocks
  - [ ] `entity.Render` machinery: how the game renders entities
  - [ ] Culling: how the game decides what to render


bobtwinkles:
Oops, completely forgot about particles (wow brain)... I think I'll write that after the `ModelBlock` and `TESR` stuff.

bobtwinkles:
`ModelBlock` 20k foot overview is more or less completed.

rendering/modelblock.md
```diff
@@ -0,0 +1,157 @@
...
+```java
+class BlockCustomGrass extends Block {
+    // snip a whole bunch of important stuff for brevity's sake, I'll assume you already have a basic block working.
+    public static final PropertyBool SNOWY = PropertyBool.create("snowy");
+    public static final PropertyBool POWERED = PropertyBool.create("powered");
+
+    public BlockCustomGrass() {
+        super(Material.grass);
+        this.setDefaultState(
+            this.blockState.getBaseState().withProperty(SNOWY, boolean.valueOf(false))
```
tterrag1098: These should be caps Boolean, but is that call really needed? Are the values not autoboxed?

rendering/modelblock.md
```diff
@@ -0,0 +1,157 @@
...
+                                          .withProperty(POWERED, boolean.valueOf(False))
```
tterrag1098: `false` not False

rendering/modelblock.md
```diff
@@ -0,0 +1,157 @@
...
+In order to describe our mapping, we'll need a client proxy.
+Using the client proxy, we'll create a custom `StateMap`, using `StateMap.Builder` that ignores our `POWERED` state.
+
+```java
+import net.minecraftforge.client.model.ModelLoader;
+public class ClientProxy extends CommonProxy {
+    @Override
+    public void preInit () {
+        ModelLoader.setCustomStateMappeR(BlockCustomGrass.class,
+            (new StateMap.Builder()).addPropertiesToIgnore(new IProperty[] {BlockCustomGrass.POWERED}).build())
```
tterrag1098: The array instantiation is completely unnecessary, it's decompiler noise.

rendering/modelblock.md
```diff
@@ -0,0 +1,157 @@
...
+```
+
+This isn't sufficient to get the nice `snowy=true` and `snowy=false` to work in the JSON however, we're going to have to write a bit of code to tell the game what properties our block has.
+Additionally, we need to tell the game how to convert our properties in to the 4-bit metadata value.
+Since we can compute whether or not the block is snowy at runtime, we're just going to tell the game that the metadata is always 0.
+
+```java
+    @Override
+    protected BlockState createBlockState() {
+        return new BlockState(this, new IProperty[] { SNOWY });
```
tterrag1098: Again, no array needed, that's what varargs are for

rendering/modelblock.md
tterrag1098: I just tested in eclipse, manual boxing is *not* required.

tterrag1098:
Your examples are good but they feature a lot of decompiler noise, which is a bad habit to teach. You do not need to instantiate arrays for varargs, and you do not need to manually box primitive values.

LexManos:
Remember, we're not here to give function/copy/paste code. Teach the concept. Copy/paste is bad.

bobtwinkles:
So to reduce the temptation to copy/paste things directly should I strip out essentially all the boilerplate? I.E. if I were to rewrite the `BlockCustomGrass` example, it would look something like this:

... So to start our custom grass block implementation, we're going to need a property that we want to keep track of. Snowyness is a good example, as it's pretty self contained:

```java
// somewhere, probably in your block class
public static final PropertyBool SNOWY = PropertyBool.create("snowy");
```

We'll need to initialize the default state of our property by calling `setDefaultState` on our block.
`setDefaultState` takes an `IBlockState` which we can create from our property by requesting a default state from the block. Usually this is all done in the block constructor, so something like

```java
this.blockState.getBaseState()
```

will give us a clean state object. To add our `SNOWY` property to it, all we'll need to do is invoke `withProperty` which expects an `IProperty` and a value of the type the property tracks. In this case,

```java
withProperty(SNOWY, false);
```
should do just fine.

.... etc.

tterrag1098:
You might want to have the full line, just to reinforce the purpose, and reduce confusion. What lex is saying is to avoid writing out the whole class. i.e.

```java
this.setDefaultState(this.blockState.getBaseState().withProperty(SNOWY, false));
```

Though the explanations of each part of that line are still welcome.

bobtwinkles:
Yep, putting everything in context is why I had the full listings to start with... I think the best approach is to do both: explanations of each part of the line, and then the complete line. I'll try to avoid writing out complete class listings in the future.

cpw:
@bobtwinkles this is looking good so far - it's nice to have a good resource on the new rendering system in MC 1.8/Forge 1.8. @Rainwarrior maybe you could give this a once over once bob is complete, as I am not as familiar with this new 1.8 stuff..

tterrag1098:
Is this going to be continued? Definitely a *very* important part of the Documentation project.

bobtwinkles:
Yep, I still plan on finishing this. IRL stuff sort of ate my life and I haven't had much time to work on it though... Fortunately most of that should be finished in about a week. Of course, if anyone wants to claim some part of this I'm totally fine with that. If you want me to squash this in to a somewhat mergable state this weekend I think I'll have time for that at least.

shadekiller666:
hey @bobtwinkles i probably know the most about the new model loader registry system behind @Rainwarrior himself, i can help you out, just let me know, i'll start writing some docs for the OBJ loader soon

bobtwinkles:
If you want to write documentation on that front, you should probably start from scratch. A lot has changed since (especially in the modelblock JSON area) since I started writing this documentation, and most of what I describe here isn't really best practices. I unfortunately don't have the time I need to really complete this PR any more, so I'm going to close it =(. However, I spent about 15 minutes rebasing and cleaning up the diffs just now, so if you want to look at those they're over [here](https://github.com/bobtwinkles/Documentation/tree/gfx-old).

bonii-xx:
Too bad, thanks for your work. It provides a good base to build upon, hope we can get it in. :)

PR #11, opened 2015-06-04
Adding Simple Tile Entity Syncing

nekosune:
Added instructions on simple tile entity syncing from server to client

### DIFF
--- a/networking/SimpleTileEntities.md
+++ b/networking/SimpleTileEntities.md
@@ -1,0 +1,58 @@
+Simple Tile Entity Server to Client Networking
+=============================
+
+If you want to keep the client side tile entity synced with the server, you do not need the full networking system
+
+You can split the nbt saving/loading for the syncable items into their own functions
+
+    @Override
+    public void writeToNBT(NBTTagCompound tag)
+    {
+        super.writeToNBT(tag);
+        writeCustomNBT(tag);
+    }
+    
+    public void writeCustomNBT(tag)
+    {
+        tag.setString("active","yes");
+    }
+    
+    @Override
+    public void readFromNBT(NBTTagCompound tag)
+    {
+        super.writeToNBT(tag);
+        writeCustomNBT(tag);
+    }
+    
+    public void readCustomNBT(tag)
+    {
+        active=tag.setString("active");
+    }
+    
+Then override the getDescriptionPacket and onDataPacket
+
+    @Override
+    public Packet getDescriptionPacket()
+    {
+        NBTTagCompound nbttagcompound = new NBTTagCompound();
+        writeCustomNBT(nbttagcompound);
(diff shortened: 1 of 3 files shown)


tterrag1098:
You need to add the page to the config in the main directory

nekosune:
Changes the commit to do so

tterrag1098:
Hm...to be consistent (and make your life easier) you should use "```" to designate code, not 4 space indents (which are a pain).

tterrag1098:
Also this allows you to do ```java to explicitly say what syntax highlighting to use.

networking/SimpleTileEntities.md
```diff
@@ -0,0 +1,58 @@
+Simple Tile Entity Server to Client Networking
+=============================
+
+If you want to keep the client side tile entity synced with the server, you do not need the full networking system
+
+You can split the nbt saving/loading for the syncable items into their own functions
+
+    @Override
+    public void writeToNBT(NBTTagCompound tag)
```
LexManos: Too much copy pasta code.

networking/SimpleTileEntities.md
```diff
@@ -0,0 +1,58 @@
...
+    {
+        super.writeToNBT(tag);
+        writeCustomNBT(tag);
+    }
+    
+    public void writeCustomNBT(tag)
```
LexManos: Why the redirection? No explanation and just complicates things you could just call read/write directly.

networking/SimpleTileEntities.md
```diff
@@ -0,0 +1,58 @@
...
+        active=tag.setString("active");
+    }
+    
+Then override the getDescriptionPacket and onDataPacket
+
+    @Override
+    public Packet getDescriptionPacket()
+    {
+        NBTTagCompound nbttagcompound = new NBTTagCompound();
+        writeCustomNBT(nbttagcompound);
```
LexManos: Why NBT? Why the exact same values from read/write NBT?
No explanation of what actually should be synced and why.
You're just giving copy pasta code which is BAD.

networking/SimpleTileEntities.md
```diff
@@ -0,0 +1,58 @@
...
+        super.onDataPacket(net,pkt);
+        readCustomNBT(pkt.getNbtCompound);
+    }
+  
+  
+Finally you need to override markDirty
+
+    @Override
+    public void markDirty()
+    {
```
LexManos: Why do you need to do this? This has nothing to do with networking it's called in a couple of placed, there is no explination of what this actually is.

LexManos:
'Put this code in your code' is not documention.

tterrag1098:
I have to say, I agree. Documentation explains *why and how*, this is a tutorial, which is not what we are going for.

diesieben07:
I would not call this a tutorial either.

PR #12, opened 2015-06-15
[WIP] Add conventions

unascribed:
Just some general conventions for mods, intended to be fairly generic.

I'm basing this on my own experiences and common 'de-facto' practices I've seen in actual mods.

Feedback welcome.

- [x] Introduction
- [x] Mod identity (name/modid/version)
- [x] Package and class naming
- [ ] Ore dictionary names
- [ ] Block/item naming
- [ ] Other?

PR #13, opened 2015-06-17
Versioning conventions

ThePsionic:
I know @unascribed (see #12) also has a version section in his pull request, but this is an idea we (with the help of @PaleoCrafter) came up with. It's basically a modified version of Semantic Versioning, which allows for easier MC version spotting as well as changing versions for API changes that don't break the world. Any feedback is appreciated.


ThePsionic:
Ugh, one second

PR #16, opened 2015-06-23
Add versioning file to index

tterrag1098:


tterrag1098:
BTW, I test all my branches on my own rtd site, this one is here: http://mcforge-tterrag.readthedocs.org/en/fix_index/

PR #17, opened 2015-06-23
Add Forge blockstate documentation

bonii-xx:
Explains what the forge blockstate json does and how to use it by example. Covers submodels and properties.

I don't have any way of seeing how it looks locally, so I don't know if there are any markup mistakes in it.


mkdocs.yml
```diff
@@ -4,6 +4,8 @@ pages:
   - [styleguide.md, Style Guide]
   - [gettingstarted/index.md, Getting Started, Home]
   - [gettingstarted/structuring.md, Getting Started, Structuring Your Mod]
+  - [blockstates/introduction.md, Models & Rendering, Blockstates]
+  - [blockstates/forgeBlockstates.md, Models & Rendering, Blockstates, Forges Blockstates]
```
tterrag1098: This is invalid, there are only 2 levels

tterrag1098: I take that back, I don't think this is what was causing the build to fail. It seems it's the link at the bottom...

tterrag1098: Ok, after further investigating it is using a number as the link label which causes the error....bizzare. Using a name like `forge` i.e. 

```
... [Forges Blockstate Json][forge] allows you to do so in 1.8.
[forge]: forgeBlockstates.md "Forges Blockstate Json"
```

seems to work fine.

tterrag1098:
Going to fix the minor issues myself since boni seems to have vanished.

PR #18, opened 2015-06-24
Blockstates: Add Lex' remarks and fix the enumeration

bonii-xx:
I don't have a local setup to test if the enumeration is actually fixed, if you please would, tterrag. Weirdly enough it worked fine in the online editor I tested it with.

For documentation purposes: The issue is that the enumeration is interpreted as *foo* and therefore converted into multiple italic parts. If this doesn't fix it, then the line-endings are probably incorrect.

### DIFF
--- a/blockstates/forgeBlockstates.md
+++ b/blockstates/forgeBlockstates.md
@@ -1,9 +1,9 @@
-Forges Blockstates
+Forge's Blockstates
 ===================
 
 Forge has its own blockstate json format to accommodate for modders needs. It introduces submodels, which allows you to build the final blockstate from different parts. You can build a models *normal* blockstate from multiple parts as well as create a complex variant depending on the blocks properties.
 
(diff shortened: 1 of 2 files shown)


blockstates/forgeBlockstates.md
```diff
@@ -3,7 +3,7 @@ Forges Blockstates
 
 Forge has its own blockstate json format to accommodate for modders needs. It introduces submodels, which allows you to build the final blockstate from different parts. You can build a models *normal* blockstate from multiple parts as well as create a complex variant depending on the blocks properties.
 
-*DISCLAIMER: Note that all models and textures referenced are from vanilla minecraft. For your own mod, you have to use the full location! For example: "mymod:blocks/blockTexture"*
+*DISCLAIMER: Note that all models and textures referenced are from vanilla minecraft. For your own mod, you have to use the full location! For example: "mymod:blocks/blockTexture". You don't have to use Forges blockstate format at all, you can also use the vanilla format!*
```
ThePsionic: Forges -> Forge's

bonii-xx: Thanks, was there several times even.

PR #29, opened 2015-12-14
Added a page about proxies

sham1:
A simple page explaining SidedProxies and what they can be used for. There's also a simple, yet useful example about item model registering demonstrating the necessity of them.


docs/gettingstarted/proxies.md
```diff
@@ -0,0 +1,90 @@
...
+
+What are sided proxies
+----------------------
+Sided proxies are a way to introduce side-specific things into your mod's code base in such a way that the other side will not crash when referencing classes intended for the other side.
+
+An example of this would be a piece of code meant to register item models for rendering:
+
+```
+ModelLoader.setCustomModelResourceLocation(yourItem, 0, new ModelResourceLocation("modid:model_name", "inventory"));```
+
```
tterrag1098: You missed a closing \`\`\` here.

sham1: Yeah, I already am fixing and squashing that fix so I can push it.

tterrag1098:
Going to leave this open for a while to get some discussion

DarkGuardsman:
I suggest mayber offering an alt design pattern. For example how I use preInit() init() and postInit() https://github.com/BuiltBrokenModding/VoltzEngine/blob/development/src/main/scala/com/builtbroken/mc/lib/mod/AbstractProxy.java

shadowfacts:
Looks good, I have a couple of suggestions though:

1. Early on, you should specify that when you are referring to "side" it means client/server as this could be misinterpreted by newcomers.
2. This is purely stylistic, but inline code such as @SidedProxy should be surrounded by single back ticks.

docs/gettingstarted/proxies.md
```diff
@@ -0,0 +1,129 @@
...
+----------------------
+Sided proxies are a way to introduce side-specific things into your mod's code base in such a way that the other side will not crash when referencing classes intended for the other side. As a quick recap, sides here refer to either client or dedicated servers.
+
+An example of this would be a piece of code meant to register item models for rendering:
+
+```
+ModelLoader.setCustomModelResourceLocation(yourItem, 0, new ModelResourceLocation("modid:model_name", "inventory"));
+```
+
+While in your `@Mod` file you may want to just check for an appropriate side in one of the lifecycle events, it still won't help as the will still run in the server-side, where ModelResourceLocation cannot be found, causing an error.
```
Unh0lyTigg: "it still won't help as the will still run in the server-side" might want to check your grammar there, might I suggest "it still won't help as it will still run on the server-side" instead.

docs/gettingstarted/proxies.md
```diff
@@ -0,0 +1,129 @@
...
+  ```
+  Then you just implement these in your client and server proxies, remember to call the super-class methods before doing anything side-specific so you do not forget to register items or whatever. And then call these inside your lifecycle-events in your `@Mod`-class.
+
+* Have CommonProxy be the proxy for server side
+
+  Example:
+
+  ```
+  public class CommonProxy {
+    // Put hatever methods here
```
Unh0lyTigg: Missing a w for whatever

RainWarrior:
Yes, there needs to be a big section on Sides in general first, what they are, where they come from, how are they different from client and server "threads", e.t.c.

sham1:
Yeah, did some changes to the document ordering so if the reader goes on chronological order they will know what sides are and have had gotten a bit of knowledge about proxies from the side page.

docs/concepts/proxies.md
```diff
@@ -0,0 +1,130 @@
+Using SidedProxies
+===================
+
+This page will go through all the benefits of using proxies and in what circumstance to use them in.
+
+What are sided proxies
+----------------------
+Sided proxies are a way to introduce side-specific things into your mod's code base in such a way that the other side will not crash when referencing classes intended for the other side. As a quick recap, sides here refer to either client or dedicated servers.
```
bonii-xx: This is not correct, since SSP also has client and server side.

docs/concepts/proxies.md
```diff
@@ -0,0 +1,130 @@
...
+
+While in your `@Mod` file you may want to just check for an appropriate side in one of the lifecycle events, it still won't help as all the classes referenced there will still be loaded, including ModelResourceLocation,
+which does not exist on the dedicated server, causing an error there.
+The solution is to use a proxy class, which only gets loaded on client, thus making sure that the above code is called on the right place.
+
+Using sided proxies
+-------------------
+
+The main meat of using proxies is an Forge-added annotation called `@SidedProxy`. You give the annotation two strings as paths to your proxy classes, both client and server.
+This annotation belongs above a field, whose type is a common superclass shared by both server proxy and client proxy classes.
```
bonii-xx: Should say 'public static' field.

docs/concepts/proxies.md
```diff
@@ -0,0 +1,130 @@
...
+  @Override
+  public void registerItemModels() {
+    // You can put whatever here, for instance a comment to explain why this is empty.
+    // The important part is that this is present.
+  }
+
+}
+```
+
+Now in your mod's preInit method you can call ```proxy.registerItemModels();``` and it won't crash with a ClassNotFoundException when you try to use your mod on a dedicated server, as Forge makes sure that the ClientProxy is never refered there, therefor eliminating any change of accidentially using that client side method implementation on server.
```
bonii-xx: referred

docs/concepts/proxies.md
```diff
@@ -0,0 +1,130 @@
...
+Some alternetive styles of proxies
+----------------------------------
+
+There are also alternative styles of writing these proxies that you can explore. Here are some of them:
+* Have preInit, init and postInit be inside the proxies
+
+  Example:
+
+  ```
+  * your package and possible includes *
```
bonii-xx: Should be a comment

docs/concepts/proxies.md
```diff
@@ -0,0 +1,130 @@
...
+  }
+  ```
+  Then you just implement these in your client and server proxies, remember to call the super-class methods before doing anything side-specific so you do not forget to register items or whatever. And then call these inside your lifecycle-events in your `@Mod`-class.
+
+* Have CommonProxy be the proxy for server side
+
+  Example:
+
+  ```
+  public class CommonProxy {
```
bonii-xx: Snippet should also contain the ClientProxy, otherwise it's pretty meaningless

bonii-xx:
The 'notice how X' should probably also be a comment in the code snippet rather than in the header.

LexManos:
It would probably be better to think of and refer to these things as Environments instead of 'Sides'

DarkGuardsman:
Environment may be more correct, but people most likely understand side better. At least, when it is prefixed with client or server.

williewillus:
Is this obsoleted by my sides doc?

tterrag1098:
Yes

PR #31, opened 2016-01-04
Added information about GC overhead limit exceeded error

swilgosz:
This is an error which occurs even if user goes step by step through current documentation so information about it should be included


bonii-xx:
Oh right, that stuff. Should probably on that page yes. Not sure if the formatting used is correct for the web-stuff though.

matthewprenger:
Yeah looks like the formatting is messed up. Also I would mention that `~` means the user home. I've found a lot of windows users that don't know what that means.

tterrag1098:
Adding the block there makes the numbers reset, see http://documentation-tterrag.readthedocs.org/en/latest/gettingstarted/

ThePsionic:
That's because [Markdown always starts at 1 for a new list](http://daringfireball.net/projects/markdown/syntax#list). You'd need to rewrite the second part with HTML code using `<ol start="5">`.

tterrag1098:
You can't use HTML, so it'll have to be done another way.

swilgosz:
I updated the code, fixing list numbering problem and added information about `~` - you can see running changes here: http://swilgosz-forge-doc.readthedocs.org/en/latest/gettingstarted/

tterrag1098:
Ok, this looks good.

@PaleoCrafter looks like multiline code blocks still look derpy inside admonitions. Might be worth some more CSS fixing :)

PR #33, opened 2016-01-17
Preliminary article on sides

williewillus:
Not sure if this is the clearest way this can be said, but I think it cleans up a lot of misconceptions about when to use the various sided mechanisms available to us.

### DIFF
--- a/docs/concepts/sides.md
+++ b/docs/concepts/sides.md
@@ -1,0 +1,69 @@
+Sides in Minecraft
+===================
+
+
+----------
+
+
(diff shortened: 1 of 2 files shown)


bonii-xx:
Might be a good place to also mention that networking is multithreaded. There are only very few cases (and probably no regular use cases) where it doesn't make sense to do stuff in the main thread.

LexManos:
Ive been wanting to change the name of @SideOnly to something more like @EnvironmentOnly to distinguis between networking sides and code environments.
But... Dont think that'll happen. But getting people to think of those two concepts as seperate would be good and something to address more in this tutorial.

PR #34, opened 2016-01-31
Explanations for the different data attaching/storage systems.

gigaherz:
As the title says, I added straightforward and to the point explanations of the IEEP and WorldSavedData systems, along with skeleton examples of both.

### DIFF
--- a/docs/datastorage/capabilities.md
+++ b/docs/datastorage/capabilities.md
@@ -1,0 +1,144 @@
+The Capability System
+==========================
+
+Capabilities allow exposing features in a dynamic and flexible way, without having to resort to directly implementing many interfaces.
+
+In general terms, each capability provides a feature in the form of an interface, alongside with a default implementation which can be requested, and a storage handler for at least this default implementation. The storage handler can support other implementations, but this is up to the capability implementor, so look it up in their documentation before trying to use the default storage with non-default implementations.
+
+Forge adds capability support to TileEntities, Entities, and ItemStacks, which can be exposed either by attaching them through an event or by overriding the capability methods in your own implementations of the objects. This will be explained in more detail in the following sections.
+
+Forge-provided Capabilities
+---------------------------
+
+As of this writing, Forge provides just one capability: IItemHandler.
+
+This capability exposes an interface for handling inventory slots. It can be applied to TileEntities (chests, machines, ...), Entities (extra player slots, mob/creature inventories/bags), or ItemStacks (portable backpacks and such). It replaces the old `IInventory` and `ISidedInventory` with with an automation-friendly system.
+
+Using an Existing Capability
+----------------------------
+
+As mentioned earlier, TileEntities, Entities, and ItemStacks implement the capability provider feature, through the `ICapabilityProvider` interface. This interface adds two methods, `hasCapability` and `getCapability`, which can be used to query the capabilities present in the objects.
+
+In order to obtain a capability, you will need to refer it by its unique instance. In the case of the Item Handler, this capability is primarily stored in `CapabilityItemHandler.ITEM_HANDLER_CAPABILITY`, but it is possible to get other instance references by using the `@CapabilityInject` annotation.
+
+```Java
+@CapabilityInject(IItemHandler.class)
+static Capability<IItemHandler> ITEM_HANDLER_CAPABILITY = null;
+```
+
+This annotation can be applied to fields and methods. When applied to a field, it will assign the instance of the capability (the same one gets assigned to all fields) upon registration of the capability, and left to the existing value (null), if the capabilitiy was never registered. Because local static field accesses are fast, it is a good idea to keep your own local copy of the reference for objects that work with capabilities. This annotation can also be used on a method, in order to get notified when a capability is registered, so that certain features can be enabled conditionally.
+
+Both the `hasCapability` and `getCapability` methods have a second parameter, of type EnumFacing, which can be used in the to request the specific instance for that one face. If passed `null`, it can be assumed that the request comes either from within the block, or from some place where the side has no meaning, such as a different dimension. In this case a general capability instance that does not care about sides will be requested instead. The return type of `getCapability` will correspond to the type declared in the capability passed to the method. For the Item Handler capability, this is indeed `IItemHandler`.
+
+Exposing a Capability
+---------------------
+
+In order to expose a capability, you will first need an instance of the underlying capability type. Note that you should assign a separate instance to each object that keeps the capability, since the capability will most probably be tied to the containing object.
+
+There's two ways to obtain such an instance, through the Capability itself, or by explicitly instantiating an implementation of it. The first method is designed to use a default implementation, if those default values are useful for you. In the case of the Item Handler capability, the default implementation will expose a single slot inventory, which is most probably not what you want.
+
+The second method can be used to provide custom implementations. In the case of `IItemHandler`, the default implementation uses the `ItemStackHandler` class, which has an optional argument in the constructor, to specify a number of slots. However, relying on the existance of these default implementations should be avoided, as the purpose of the capability system is to prevent loading errors in contexts where the capability is not present, so instantiation should be protected behind a check testing if the capability has been registered (see the remarks about `@CapabilityInject` in the previous section).
+
+Once you have your own instance of the capability interface, you will want to notify users of the capability system that you expose this capability. This is done by overriding the `hasCapability` method, and comparing the instance with the capability you are exposing. If your machine has different slots based on which side is being queried, you can test this with the `facing` parameter. For Entities and ItemStacks, this parameter can be ignored, but it is still possible to have side as a context, such as different armor slots on a player (top side => head slot?), or about the surrounding blocks in the inventory (west => slot on the left?). Don't forget to fall back to `super`, otherwise the attached capabilities will stop working.
+
+```Java
+@Override
+public boolean hasCapability(Capability<?> capability, EnumFacing facing)
+{
+  if (capability == CapabilityItemHandler.ITEM_HANDLER_CAPABILITY)
+    return true;
+  return super.hasCapability(capability, facing);
+}
+```
+
+Similarly, you will want to provide the interface reference to your capability instance, when requested. Again, don't forget to fall back to `super`.
+
+```Java
+@Override
+public <T> T getCapability(Capability<T> capability, EnumFacing facing)
+{
+  if (capability == CapabilityItemHandler.ITEM_HANDLER_CAPABILITY)
+    return (T)inventory;
+  return super.getCapability(capability, facing);
+}
+```
+
+It is strongly suggested that direct checks in code are used to test for capabilities instead of attempting to rely on maps other other data structures, since capability tests can be done by many objects every tick, and they need to be as fast as possible in order to avoid slowing down the game.
+
+Attaching Capabilities
+----------------------
+
+As mentioned, attaching capabilities to entities and itemstacks can be done through the AttachCapabilityEvent. This event has 3 sub-events, for extra granularity:
+* `AttachCapabilityEvent.Entity`: Fires only for entities.
+* `AttachCapabilityEvent.TileEntity`: Fires only for tile entities.
+* `AttachCapabilityEvent.Item`: Fires only for item stacks.
+
+In all cases, the event has a method `addCapability`, which can be used to attach capabilities to the target object. Instead of adding capabilities themselves to the list, you add capability providers, which have the chance to return capabilities only from certain sides. While the provider only needs to implement `ICapabilityProvider`, if the capability needs to store data persistently it is possible to implement `ICapabilitySerializable<T extends NBTBase>` which, on top of returning the capabilities, will allow providing NBT save/load functions.
+
+For information on how to implement `ICapabilityProvider`, refer to the [Exposing a Capability](#exposing-a-capability) section.
+
+Creating Your Own Capability
+----------------------------
+
+In general terms, a capability is declared and registered through a single method call to `CapabilityManager.INSTANCE.register()`. One possibility is to define a static `register()` method inside a dedicated class for the capability, but this is not required by the capability system. For the purpose of this documentation we will be describing each part as a separate named class, although anonymous classes are an option.
+
+```Java
+CapabilityManager.INSTANCE.register(capability interface class, storage, default implementation factory);
+```
+
+The first parameter to this method, is the type that describes the capability feature. In our example, this will be `IExampleCapability.class`.
+
+The second parameter is an instance of a class that implements `Capability.IStorage<T>`, where T is the same class we specified in the first parameter. This storage class will help manage saving and loading for the default implementation, and it can, optionally, also support other implementations.
+
+```Java
+private static class Storage
+    implements Capability.IStorage<IExampleCapability>
+{
+  @Override
+  public NBTBase writeNBT(Capability<IExampleCapability> capability, IExampleCapability instance, EnumFacing side)
+  {
+    // return an NBT tag
+  }
+
+  @Override
+  public void readNBT(Capability<IExampleCapability> capability, IExampleCapability instance, EnumFacing side, NBTBase nbt)
+  {
+    // load from the NBT tag
+  }
+}
+```
+
+The last parameter is a callable factory that will return new instances of the default implementation.
+
+```Java
+private static class Factory implements Callable<IExampleCapability>
+{
+  @Override
+  public IExampleCapability call() throws Exception
+  {
+    return new Implementation();
+  }
+}
+```
+
+Finally, we will need the default implementation itself, to be able to instantiate it in the factory. Designing this class is up to you, but it should at least provide a basic skeleton that people can use to test the capability, if it's not a fully usable implementation on itself.
+
+Synchronizing Data with Clients
+-------------------------------
+
+By default, Capability data is not sent to clients. In order to change this, the mods have to manage their own synchronization code using packets.
+
+There are three different situation in which you may want to send synchronization packets, all of them optional:
+
+1. When the entity spawns in the world, or the block is placed, you may want to share the initialization-assigned values with the clients.
+1. When the stored data changes, you may want to notify some or all of the watching clients.
+1. When a new client starts viewing the entity or block, you may want to notify it of the existing data.
+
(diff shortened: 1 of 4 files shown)


docs/datastorage/extendedentityproperties.md
```diff
@@ -0,0 +1,199 @@
+Extended Entity Properties
+==========================
+
+Extended entity properties allow attaching data to entities.
+
+Declaration
+-----------
+
+The basis of the EEPs is the ```IExtendedEntityProperties``` interface. This interface provides the basic methods required for managing the extended data:
```
tterrag1098: Just use single backticks, like `this`

docs/datastorage/extendedentityproperties.md
```diff
@@ -0,0 +1,199 @@
...
+In order to uniquely identify your property and avoid duplication, the method takes a string parameter with an identifier for the property. A good practice is to include the modid in this string, so that it will not collide with other mods.
+
+**WARNING**: If the same property identifier is added twice, Forge will append a number to it, and return this modified identifier from the ```registerExtendedProperties``` method. If you don't want that to happen, you can use ```Entity#getExtendedProperties``` to check if an IEEP with that name was already added.
+
+Usage
+-----
+
+To make use of the extended data, the instance of the IEEP implementation has to be obtained from the Entity, and because the entity could have been unloaded or may have changed dimensions, it is not safe to cache the references.
+
+To obtain the IEEP reference, one would use Entity#getExtendedProperties, with the same property ID specified on registration. THe return value, if not null, is the instance of ```IExtendedEntityProperties``` added during entity construction.
```
tterrag1098: THe -> The

docs/datastorage/extendedentityproperties.md
```diff
@@ -0,0 +1,199 @@
...
+
+A good idea is to create a static "get" method in your IEEP implementation, that will automatically obtain the instance, and cast it to your implementation class name.
```
tterrag1098: "name" is extraneous (and wrong, really)

docs/datastorage/extendedentityproperties.md
```diff
@@ -0,0 +1,199 @@
...
+
+This can be done by handling the ```TickEvent.PlayerTickEvent``` event if the entity is attached to players.
+
+Skeleton example
+----------------
+
+This skeleton can be used as a basis for new IEEP implementations. 
+
+```Java
+public class ExampleEntityProperty implements IExtendedEntityProperties
```
tterrag1098: We try to avoid providing complete implementations, as this just results in copypasta. Please try to explain the concepts rather than just pasting code.

gigaherz: I would normally agree with not providing a block of code that can be copypasted, but I bevelie in this case, a functional skeleton that people can easily dump into their code is good, since almost everything in it is just boilerplate code and people trying to implement it their own way may result in more issues than benefits.

Also, explaining "the concepts" is already done above the code? I can't think of anything else to explain that is either general Java programming knowledge, or out of the scope for that explanation (belonging on other documentation pages)...

LexManos:
Just as an overall, the IEEP system is/will be deprecated in favor of the Capability system that is now in place in 1.8.9+

docs/datastorage/extendedentityproperties.md
LexManos: No copy/pasta code. Period.

gigaherz:
I was planning to do a page on Capabilities at a later date, when i get to understand them correctly (intended usage, best practices and such). I didn't think the Capabilities system would deprecate IEEPs, so I wrote about those first.

Can this be merged meanwhile? Or would you rather I remove the IEEP page from the PR and keep only the WorldSavedData?

LexManos:
Ya i would say either dont have it, or have a 'This is a legacy page, the replacement is Capabilities' banner at the top.

gigaherz:
Took the time to add a page on Capabilities, and link it from the IEEP page. Hopefully I understood them correctly.

tterrag1098:
I'll read through this, but not being familiar with capabilities I can't offer much there. Would need @LexManos or some modder who has used them to chime in.

docs/datastorage/capabilities.md
```diff
@@ -0,0 +1,176 @@
+The Capability System
+==========================
+
+Capabilities allow exposing features in a dynamic and flexible way, without having to resort to directly implementing many interfaces.
+
+In general terms, each capability provides a feature in the form of a class or interface, alongside with a default implementation which can be requested, and a storage handler for at least this default implementation. The storage handler can support other implementations, but this is up to the capability implementor, so look it up in their documentation before trying to use the default storage with non-default implementations.
+
+Forge adds capability support to TileEntities, Entities, and ItemStacks, which can be exposed either by attaching them through an event or, with the exception of ItemStacks, by overriding he capability methods in your own implementations of the entities. This will be explained in more detail in the following sections.
```
LexManos: "overriding the", and Items can overide their own methods to implement the ItemStack one. It's just weird for them because of the instance vs singleton thing. The PREFERRED method is to implement the functions on your own TEs/Entites/Items because you have the code and you can dictate what goes first. The event is just a hacky compromise that we did to appease certain modders.

docs/datastorage/capabilities.md
```diff
@@ -0,0 +1,176 @@
...
+
+Using an Existing Capability
+----------------------------
+
+This most notably applies to the Item Handler capability added by Forge, which replaces the vanilla IInventory and ISidedInventory with an automation-friendly system.
+
+As mentioned earlier, TileEntities, Entities, and ItemStacks implement the capability provider feature, through the `ICapabilityProvider` interface. This interface adds two methods, `hasCapability` and `getCapability`, which can be used to query the capabilities present in the objects.
+
+In order to obtain a capability, you will need to refer it by its unique instance. In the case of the Item Handler, this capability is primarily stored in `CapabilityItemHandler.ITEM_HANDLER_CAPABILITY`, but it is possible to get an instance reference by using the `@CapabilityInject` annotation. This annotation can also be used on a method in order to get notified when a capability is registered, so that certain features can be enabled conditionally. Please refer to the javadocs for this annotation for details on usage.
```
LexManos: I would go more in depth of how to use the annotation on fields. As thats the main intended use. Local static field access is super fast.

docs/datastorage/capabilities.md
```diff
@@ -0,0 +1,176 @@
...
+
+Both the `hasCapability` and `getCapability` methods have a second parameter, of type EnumFacing, which can be used in the TileEntities to request the specific instance for that one face. If passed `null`, a general capability instance will be requested instead. The return type of `getCapability` will correspond to the type declared in the capability passed to the method. For the Item Handler capability, this is indeed `IItemHandler`.
+
```
LexManos: Something along the lines of 'null' is considered the 'self', or 'unknown' value and should be treated as if the request is coming from inside the block, or potentially across dimensions. It should return a interface that doesn't care about side, or just pick a side to treat it as.

docs/datastorage/capabilities.md
```diff
@@ -0,0 +1,176 @@
...
+Exposing a Capability
+---------------------
+
+In order to expose a capability, you will first need an instance of the underlying capability type. There's two ways to obtain such an instance, through the Capability itself, or by explicitly instantiating an implementation of it.
+
+The first method is designed to use a default implementation, if those default values are useful for you. In the case of the Item Handler capability, the default implementation will expose a single slot inventory, which is most probably not what you want.
+
+The second method can be used to customize the existing implementations, or to implement your own. In the case of `IItemHandler`, the default implementation uses the `ItemStackHandler` class, which has an optional argument in the constructor, to specify a number of slots.
+
```
LexManos: It is actually preferred that you do NOT use these 'ItemStackHandler's or default implementations unless the contract of the interface explicitly states that those classes will exist. Sadly there is no way to do that in code, so this is reliant on the API spec for each interface.

docs/datastorage/capabilities.md
```diff
@@ -0,0 +1,176 @@
...
+```Java
+// A dispenser-like block could use something like this
+IItemHandler inventory = new ItemStackHandler(9);
+```
```
LexManos: It is important to note NOT to do this in static fields/methods. AND gate them behind a if (CAPABILITY_HAS_BEEN_REGISTERED) {Typically if (Field_With_CapInject != null)}
As the main point of this is to prevent class not found exceptions when you are in an environment that the capability doesn't exist.

mkdocs.yml
```diff
@@ -12,6 +12,10 @@ pages:
     - Home: 'networking/index.md'
     - Overview: 'networking/overview.md'
     - SimpleImpl: 'networking/simpleimpl.md'
+  - Data Storage:
+    - Capabilities: 'datastorage/capabilities.md'
```
tterrag1098: Not sure if this is the best category. Maybe a new category such as "Forge Systems"?

docs/datastorage/capabilities.md
```diff
@@ -0,0 +1,176 @@
...
+
+Once you have your own instance of the capability interface, you will want to notify users of the capability system that you expose this capability. This is done by overriding the `hasCapability` method, and comparing the instance with the capability you are exposing. If your machine has different slots based on which side is being queried, you can test this with the `facing` parameter. For entities and ItemStacks, this parameter can be ignored. Don't forget to fall back to `super`, otherwise the attached capabilities will stop working.
+
```
LexManos: The facing can be ignored by any implementors but if your system is side aware, it should take this into account.
For example in Players:
IItemHandler COULD be expanded to be side aware and have say, UP == Helmet, DOWN = Boots, ANY_SIDE = chest/sword/legs.
We won't be doing that in forge, but that is just an example of where sided-ness may be import in entities.
Same goes for item stacks, Example:
A mod wants to make furnaces function in a inventory. So on the 2d grid, Sides=output, Top = Input, Bottom = Fuel.

It is also worth it to note that get/hasCapability is designed to be fast, use only If statements when you can, don't rely on maps and the like.  has is also meant to assist in LAZY loading of the capability. If that is the implementer's intention.

docs/datastorage/capabilities.md
```diff
@@ -0,0 +1,176 @@
...
+Attaching Capabilities
+----------------------
+
+As mentioned, attaching capabilities to entities and itemstacks can be done through the AttachCapabilityEvent. This event has 3 sub-events, for extra granularity:
+* `AttachCapabilityEvent.Entity`: Fires only for entities.
+* `AttachCapabilityEvent.TileEntity`: Fires only for tile entities.
+* `AttachCapabilityEvent.Item`: Fires only for item stacks.
+
+In all cases, the event has a method `addCapability`, which can be used to attach capabilities to the target object. Instead of adding capabilities themselves to the list, you add capability providers, which have the chance to return capabilities only from certain sides. While the provider only needs to implement `ICapabilityProvider`, it is good idea to implement `ICapabilitySerializable<T extends NBTBase>`, which on top of returning the capabilities, will allow providing NBT save/load functions.
+
```
LexManos: "If you need to store object specific data it is a good idea to implement"
Because unless you need to store data, implementing that interface does nothing but waste cpu cycles. So don't do it unless you need it.

docs/datastorage/capabilities.md
```diff
@@ -0,0 +1,176 @@
...
+For information on how to implement `ICapabilityProvider`, refer to the [Exposing a Capability](#exposing-a-capability) section.
+
+Creating Your Own Capability
+----------------------------
+
+In general terms, a capability is declared and registered through a single method call.
+
+First of all, we will create our feature as a skeleton class or interface. For the purpose of adding data to an entity (replacing an IEEP), the feature can be just a class, but for an API, you will most probably want to design it around an interface.
```
LexManos: Actually as a practice we want to keep it purely as an interface. Sadly there is no way to force this at the code level so we need to do it at the document level.

docs/datastorage/capabilities.md
```diff
@@ -0,0 +1,176 @@
...
+
+Although the capability instance itself will be initialized internally by Forge, we'll start by creating a class that manages the capability features. In this class, we will create a field to hold the instance of the capability.
+
+```Java
+@CapabilityInject(IExampleCapability.class)
+public static Capability<IExampleCapability> INSTANCE = null;
+```
+
+Then, we will be defining a static `register()` method, which will call `CapabilityManager.INSTANCE.register()`. For the purpose of this documentation we will be describing each part as a separate named class, although anonymous classes are an option.
```
LexManos: It doesnt need to be a register() function i'd prefer you just describe the function call not give copypasta code.

docs/datastorage/capabilities.md
```diff
@@ -0,0 +1,176 @@
...
+{
+  CapabilityManager.INSTANCE.register(IExampleCapability.class, new Storage(), new Factory());
+}
+```
+
+The first parameter to this method, is the type that describes the capability feature. In our example, this type refers to an interface.
+
+The second parameter is an instance of a class that implements `Capability.IStorage<T>`, where T is the same class we specified in the first parameter. This storage class will help manage saving and loading for the default implementation, and it can, optionally, also support other implementations.
+
+For the purpose of our example, we will make the capability interface implement the `INBTSerializable<NBTTagCompound>` interface, which will allow us to delegate the saving and loading to the instance itself, but custom saving and loading are possible, and depending on the case, maybe even preferred.
```
LexManos: It's entirely preferred, the entire point is to not require the interface to do it :/

docs/datastorage/capabilities.md
```diff
@@ -0,0 +1,176 @@
...
+  {
+    if (e.original.hasCapability(ExampleCapability.INSTANCE, null) &&
+        e.entityPlayer.hasCapability(ExampleCapability.INSTANCE, null))
+    {
+      NBTTagCompound tag = e.original.getCapability(ExampleCapability.INSTANCE, null).serializeNBT();
+      e.entityPlayer.getCapability(ExampleCapability.INSTANCE, null).deserializeNBT(tag);
+    }
+  }
+}
+```
```
LexManos: May need to write something a bit cleaner in forge to handle this but ya the whole 'lets create a new entity when respawning/changing worlds is annoying' 
However our docs SHOULD NOT encourage people to do this as it's going to create copy pasta code that will make every capability implement serializeable which is not the intention.
Probably would be best if there was a PR to make this issue act like ItemStack's clone() method, where IT called the serialize function's internally.

docs/datastorage/capabilities.md
```diff
@@ -0,0 +1,176 @@
...
+
+Updating the Contents Over Time
+-------------------------------
+
+By default, the capability data does not tick. In order to change this, the data has to be manually made to update.
+
+This can be done by handling the `TickEvent.PlayerTickEvent` event if the entity is attached to players.
```
LexManos: This is a side note and no, has nothing to do with caps, don't mention it.

docs/datastorage/capabilities.md
gigaherz: Ah, `Item#initCapabilities`? I didn't notice it until now. Fixing.

mkdocs.yml
gigaherz: I'm not fully convinced either, but I think the issue is maybe the name of the category instead. The meaning I wanted was "systems that allow you to attach data to things", to go along with "Networking", but that's the best name I could come up with.

gigaherz:
Thanks for the comments, i just pushed a reworked page.

docs/datastorage/capabilities.md
```diff
@@ -0,0 +1,139 @@
+The Capability System
+==========================
+
+Capabilities allow exposing features in a dynamic and flexible way, without having to resort to directly implementing many interfaces.
+
+In general terms, each capability provides a feature in the form of an interface, alongside with a default implementation which can be requested, and a storage handler for at least this default implementation. The storage handler can support other implementations, but this is up to the capability implementor, so look it up in their documentation before trying to use the default storage with non-default implementations.
+
+Forge adds capability support to TileEntities, Entities, and ItemStacks, which can be exposed either by attaching them through an event or by overriding he capability methods in your own implementations of the objects. This will be explained in more detail in the following sections.
```
bonii-xx: "overriding he capability" -> typo

bonii-xx:
Looking at the capabilities, what I'm missing is a short list with the Forge provided capabilities. You're using the itemhandler and stuff as examples, but the user has no idea what they actually do. Maybe a small header with a one sentence explanation what they do, to give people an idea what's possible with them and clear up confusion.

gigaherz:
I corrected the typo, and I made the fact that there's currently only one capability provided by Forge more explicit.

bonii-xx:
I'd probably word it a tad differently for people who have never used IInventory, that way it's more future proof. Maybe something along the lines of "this capability represents Tiles that have an inventory" + the stuff you have.

gigaherz:
I just didn't think explaining the usage of the IItemHandler was in the scope of that page... But yes I see your point (although inventories are not specific to tiles, they can be applied to entities and itemstacks also). I'll try to explain the capability a bit better after I eat.

gigaherz:
Updated the "forge capabilities" section to roughly explain what IItemHandler provides.

docs/datastorage/extendedentityproperties.md
```diff
@@ -0,0 +1,214 @@
...
+
+    // Read data from propertyData
+  }
+}
+```
+
+Synchronizing Data with Clients
+-------------------------------
+
+By default, the entity data does not synchronize. In order to change this, the mods have to manage their own synchronization code using packets.
```
tterrag1098: does not synchronize -> is not sent to clients

I think that would be clearer

docs/datastorage/extendedentityproperties.md
```diff
@@ -0,0 +1,214 @@
...
+```Java
+private void update()
+{
+  setStoredData(Integer.toString(entity.ticksExisted));
+}
+```
+
+```Java
+@SubscribeEvent
+public void playerTickEvent(LivingEvent.LivingUpdateEvent e)
```
tterrag1098: You talk about the PlayerTickEvent but then use LivingUpdateEvent. It's not necessary to explain what these events are but you should at least be consistent.

gigaherz: Oops! I completely forgot to update the text after changing the code -- I'll probably remove that section altogether, just like I did on the capabilities page.

PR #42, opened 2016-05-17
Updates to my articles

williewillus:
* Change sketchy proxy part in sides.md
* Overhaul states.md because reading it through as someone with no experience with states at all, it all seems very very hand wavey - no or unclear explanations about why things are the way they are. Hopefully this is better

### DIFF
--- a/docs/blockstates/states.md
+++ b/docs/blockstates/states.md
@@ -81,9 +81,7 @@ Sadly, abstractions are lies at their core. We still have the responsibility of 
 
 If you specify any `IProperty`'s, you **must** override `getMetaFromState` and `getStateFromMeta`
 
-Here you will read the values of your properties and return an appropriate integer between 0 and 15, or the other way around.
-
-The reader is left to check examples from vanilla blocks by themselves.
+Here you will read the values of your properties and return an appropriate integer between 0 and 15, or the other way around; the reader is left to check examples from vanilla blocks by themselves.
 
 !!! Warning
     Your getMetaFromState and getStateFromMeta methods MUST be one to one! In other words, the same set of properties and values must map to the same meta value and back. Failing to do this, unfortunately, WON'T cause a crash. It'll just cause everything to behave extremely weirdly.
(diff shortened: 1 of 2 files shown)


docs/blockstates/states.md
```diff
@@ -2,11 +2,15 @@ Block States
...
+This guide is designed for an entry level introduction to Block States. If you know what Extended States are, you'll notice some simplifying assumptions I've made below. They are intentional and are meant to avoid overloading beginners with information they may not immediately need. If you don't know what they are, no need to fear, there will be another document for them eventually.
 
-In Minecraft 1.8, direct manipulation of blocks and metadata values have been abstracted away into what is known as blockstates.
-The premise of the system is to remove the manipulation of raw metadata, which are ugly and carry no meaning.
+Motivation
+----------
 
-For example, consider this
+In Minecraft 1.8 and above, direct manipulation of blocks and metadata values have been abstracted away into what is known as blockstates.
+The premise of the system is to remove the usage and manipulation of raw metadata numbers, which are ugly and carry no meaning.
```
bonii-xx: -ugly
+nondescript

docs/concepts/sides.md
```diff
@@ -34,10 +34,12 @@ How do we resolve this? Luckily, FML provides us with a `@SidedProxy` annotation
 
 !!! note
 
-    It is important to understand that FML picks the proxy to instantiate based on the physical side. A single player world (logical server, logical client within a physical client) will still have a proxy of the type you specify in `clientSide`! If you want behaviours that should run on both sides, make your client proxy extend from the server proxy and call the super class method. Make sure your server proxy does not reference any classes that only exist on the physical server.
+    It is important to understand that FML picks the proxy to instantiate based on the physical side. A single player world (logical server, logical client within a physical client) will still have a proxy of the type you specify in `clientSide`! If you want behaviours that should run on both physical sides, you must arrange for that to happen outside of the proxies
```
tterrag1098: I don't see why this bit is needed at all. It's not related to the integrated vs dedicated note. IMO it should talk about how to run code based on the logical side, using the aforementioned `isRemote`.

docs/concepts/sides.md
```diff
@@ -34,10 +34,12 @@ How do we resolve this? Luckily, FML provides us with a `@SidedProxy` annotation
...
 
 A common use case is to register renderers and models, something which must be called from the main initialization methods `preInit`, `init`, or `postInit`. However, many rendering related classes and registries are not present on the physical server and may crash it. Therefore, we put these actions into the client proxy, ensuring that they will always execute for the physical client.
 
+Since both proxies must be assignable to the same field, it is a common pattern to have an interface `IProxy` as the field type, and then have two implementations, `ClientProxy` and `ServerProxy` for the two corresponding physical sides.
```
tterrag1098: This is *one* common way, but not the only one.

bonii-xx: It does say "a common pattern"?

tterrag1098: I suppose what I meant is that we shouldn't prefer one over the other. I'd prefer to be implementation agnostic as much as possible.

docs/concepts/sides.md
williewillus: I included it because people *have* been confused before as to why their "Server proxy" never runs at all in the physical client, because they think it's instantiated based on logical side (which makes no sense but people have still made that mistake)

tterrag1098: Right, so make that point. I'm just saying that last sentence (the one you edited with this commit) is largely irrelevant, talking about running on "both physical sides."

williewillus:
updated

PR #43, opened 2016-05-17
Article for Sounds

williewillus:


docs/effects/sounds.md
```diff
@@ -0,0 +1,115 @@
...
+
+In either case, you specify the path to your Sound File starting from your "sounds" asset directory. Thus, "mymod:openChestSoundFile" corresponds to the path "assets/mymod/sounds/openChestSoundFile.ogg" and "mymod:music/epicMusic" corresponds to the path "assets/mymod/sounds/music/epicMusic.ogg".
+
+Code Registration
+-----------------
+
+Simply specifying the Sound Events in JSON isn't enough, however. Due to changes in the sound system in 1.9, we must also register Sound Events in code. This very simple to do: 
+
+```Java
+ResourceLocation name = new ResourceLocation("mymod", "openChest");
```
bonii-xx: I'm not convinced that 'name' is a good variable name for this. It's a resource location, let's keep it at that.

bonii-xx:
You reference the wiki a few times but only put the link at the first occurrence. Maybe put a References header at the bottom with links to useful external sites. The wiki of course, maybe also a link to Audacity and https://xiph.org/vorbis/ since people will most likely have to convert their files to ogg.

docs/effects/sounds.md
```diff
@@ -0,0 +1,115 @@
...
+  * Sound Event - Something that triggers a sound effect. Examples include "minecraft:block.anvil.hit" or "botania:spreaderFire"
+  * Sound Category - The category of the sound, for example "player" or "block" or simply "master". The sliders in the sound settings GUI represent these categories
+  * Sound File - The literal file on disk that is played, usually an .ogg file.
+
+Sounds.json
+-----------
+
+This json should be located in your base asset directory (src/main/resource/assets/MODID/sounds.json) and indicates to the vanilla resource system what Sound Events you declare and what Sound Files those events use.
+
+A full specification is available [on the wiki], but we highlight the important parts here with an example:
```
tterrag1098: I agree with boni, make this [wiki] and use the hyperlink throughout.

docs/effects/sounds.md
```diff
@@ -0,0 +1,115 @@
...
+}
+```
+
+Underneath the top-level object, each key we have defined corresponds to a Sound Event we want to tell the game about: "mymod:openChest" and "mymod:epicMusic". Note that they are written in the sounds.json without the modid.
+
+Under each event we specify the category of the Sound Event, then a subtitle localization key to be shown to hard of hearing users when the Sound Event is played. 
+
+Finally, we specify the actual Sound Files to be played. Note that the value is an array - if we specify multiple Sound Files then the game will randomly choose one to play whenever the Sound Event is triggered.
+
+The two examples represent two different ways to specify a Sound File. The wiki has precise details, but generally, make sure to use the second form for long Sound Files such as BGM or music discs, because the "stream" argument tells Minecraft to not load the entire Sound File into memory but instead stream it from disk. Using the second form also allows you to specify the volume, pitch, and random weight of that Sound File - again, see the vanilla wiki for precise details.
```
tterrag1098: Such as here

williewillus:
Updated

PR #50, opened 2016-06-25
The update.json for Charset is no longer available

Meow-J:


williewillus:
I'd like another example from another mod to be added. The way I do the changelogs in botania is just one of many options

tterrag1098:
This is why I dislike external links. If you find a good example and it is FOSS, put it somewhere permanent like gist.

asiekierka:
http://charset.asie.pl/archive/update.json - go grab it and mirror it

PR #53, opened 2016-07-18
Fix paths in locations.md

Choonster:
All file types except `mcmod.info` go in the `assets/<modid>` folder rather than the root `src/main/resources/` folder.

### DIFF
--- a/docs/conventions/locations.md
+++ b/docs/conventions/locations.md
@@ -11,18 +11,18 @@ The `mcmod.info` file is in the root directory.
 
 ### Blockstates
 
-Blockstate definition files are in the JSON format and are in the `./blockstates/` folder.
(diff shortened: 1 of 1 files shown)


tterrag1098:
Why not just one change to line 6?

Choonster:
Because `mcmod.info` doesn't go in `assets/<modid>` like the other files do.

If I added `assets/<modid>` to line 6, should I add an exception for `mcmod.info` on line 6 or specify its full path on line 10?

cpw:
I should fix that.

LexManos:
Instead of having multiple modids in 1 json?

tterrag1098:
Well, barring that change to forge, this patch seems fine.

PR #57, opened 2016-08-08
Add note about `ItemBlocks` in blocks/blocks.md

howtonotwin:
Add note about registering `ItemBlock`s as well as registering `Block`s, and note that `GameRegistry.registerWithItem` should not be used.


williewillus:
I'd say take out the part about registerWithItem, if it's marked Deprecated in the source it's enough of a warning not to use it

diesieben07:
You might want to explain what it means to not have an item form for your block. Many people associate the "thing in the inventory" as "the block".

williewillus:
looks good to me 👍

PR #58, opened 2016-08-09
Add Items category

howtonotwin:
Add category and base page in that category for items analogous to the "Blocks" category.

### DIFF
--- a/docs/items/items.md
+++ b/docs/items/items.md
@@ -1,0 +1,34 @@
+Items
+=====
+
+Along with blocks, items are a key component of most mods. While blocks make up the world around you, items are what let you change it.
+
+Creating an Item
+----------------
+
+### Basic Items
+
+Basic items that need no special functionality (think sticks or sugar) don't need custom classes. You can simply instantiate `Item` and call its various setters to set some simple properties.
+
+|         Method         |                  Description                  |
+|:----------------------:|:----------------------------------------------|
+|    `setCreativeTab`    | Sets which creative tab this item is under. Must be called if this item is meant to be shown on the creative menu. Vanilla tabs can be found in the class `CreativeTabs`. |
+|     `setMaxDamage`     | Sets the maximum damage value for this item. If it's over `0`, 2 item properties "damaged" and "damage" are added. |
+|    `setMaxStackSize`   | Sets the maximum stack size.                  |
+|      `setNoRepair`     | Makes this item impossible to repair, even if it is damageable. |
+|  `setUnlocalizedName`  | Sets this item's unlocalized name, with "item." prepended. |
+|    `setHarvestLevel`   | Adds or removes a pair of harvest class (`"shovel"`, `"axe"`) and harvest level. This method is not chainable. |
+
+The above methods are chainable, unless otherwise stated, meaning they `return this` to facilitate calling them in series.
+
+### Advanced Items
+
+Setting the properties of an item as above only works for simple items. If you want more complicated items, you should subclass Item and override its methods.
+
+Registering an Item
+-------------------
+
+Now that you have created an item, it is time to register it. As with most things, they are registered by first setting the registry name with `setRegistryName`, and then registering them to the item registry in `RegistryEvent.Register<Item>`.
+
+!!! note
+    When you set the registry name with a simple string like `"myItem"`, the active modid is prefixed onto it. So if examplemod was registering this item, the actual name will be `"examplemod:myItem"`.
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -13,6 +13,8 @@ pages:
     - Intro to Blockstates: 'blockstates/states.md' 
     - Intro to Blockstate JSONs: 'blockstates/introduction.md'
     - Forge Blockstate JSON: 'blockstates/forgeBlockstates.md'
+  - Items:
+    - Home: 'items/items.md'
   - Events:
     - Basic Usage: 'events/intro.md'
   - Networking:


docs/items/items.md
```diff
@@ -0,0 +1,47 @@
+Items
+=====
+
+Along with blocks, items make up the backbone of any good mod. While blocks make up the world around you, items are what let you change it.
```
tterrag1098: >backbone of any good mod

Subjective. Maybe "are a key component of many mods"

docs/items/items.md
```diff
@@ -0,0 +1,47 @@
...
+
+### The Item's model
+
+Unlike blocks, which are automatically bound to their blockstate files, items must have their models explicitly registered. This holds true even for `ItemBlock`s. This registering can only happen on the clientside, so make sure you only call this code through your client proxy.
+
+Item model registering happens through `ModelLoader.setCustomModelResourceLocation`. This method registers a mapping from (item, meta) to a `ModelResourceLocation`. A `ModelResourceLocation` is the combination of a `ResourceLocation` (like `examplemod:foo`) with a variant string (like `variant=iron`). For a blockstate JSON `examplemod:block`, we can construct an RL for it, and then an MRL refers to a certain variant within that JSON.
+
+However, items don't have blockstate JSONs, therefore vanilla Minecraft forces the variant to be `"inventory"` and requires that "variants" be done through [property overrides](overrides). Forge goes ahead and undoes this, so you can now use variant strings on items.
```
tterrag1098: This is a very short sentence which glosses over a lot of things. How does Forge "undo" it? Either go into more detail or don't mention it here, it is confusing as it stands.

docs/items/items.md
```diff
@@ -0,0 +1,47 @@
...
+
+This is most useful on `ItemBlock`s, where you can simply register some meta values of an item to a variant on the block's own JSON without having to make several separate item JSONs. For ordinary items, this is useless and you are forced to use property overrides anyway.
```
diesieben07: Useless? I don't think it's useless at all for normal Items. If you just need a simple item model it is much simpler to make a blockstate json which multiple variants that all just retexture the normal item model. Then you just need one json file for many items instead of individual ones. Yes, this is personal taste, but I would not call it useless.
Also, you might want to mention that models will be loaded from models/block when referenced through a blockstate json even if it is being used for an item.

docs/items/items.md
```diff
@@ -0,0 +1,47 @@
...
+----------------
+
+### Basic Items
+
+Basic items that need no special functionality (think sticks or sugar) don't need custom classes. You can simply instantiate `Item` and call its various setters to set some simple properties.
+
+|         Method         |                  Description                  |
+|:----------------------:|:----------------------------------------------|
+|    `setCreativeTab`    | Sets which creative tab this item is under. Must be called if this item is meant to be shown on the creative menu. Vanilla tabs can be found in the class `CreativeTabs`. |
+|     `setMaxDamage`     | Sets the maximum damage value for this item. If it's over `0`, 2 item overrides "damaged" and "damage" are added. |
```
diesieben07: You might want to link to your article about what overrides are.

docs/items/overrides.md
```diff
@@ -0,0 +1,46 @@
+Item Property Overrides
+=======================
+
+Items do not have something analogous to blockstate JSONs, where you can define models in `models/block/` and then stitch them into one big model in a blockstate JSON under `blockstates/` depending on variants. Instead, you must assign an item a single model defined under `models/item/`. Then, the item will expose certain properties from the code. These properties are defined as a name string and a function that maps `ItemStack`s to arbitrary float values. The model JSON defines the default model for the item, and has an extra `"overrides"` tag, which is a list of predicates and a model to use if the predicate is true. The format can be seen on the [wiki][].
```
diesieben07: First of all, yes they do, like you explained. Blockstate jsons can be used for items.
Then, "stitch them into one big model" is very inaccurate. The blockstate json is not itself a model, it simply points to models and describes how they should be used (transformations, textures, etc.).
I would not describe property overrides as the alternative to blockstates, they are simply a different concept.

docs/items/overrides.md
```diff
@@ -0,0 +1,46 @@
...
+        },
+        {
+          "predicate": { // power >= 0 && usage >= 0
+            "power": .25,
+            "usage": 0
+          },
+          "model": "examplemod:item/exampleUnusedPowered",
+        }
+      ]
+    }
```
diesieben07: I think you could cut this example down quite a bit by simply showing off one override instead of 3. Also you might want to put more emphasis on the `>=` part.

docs/items/items.md
howtonotwin: I'm not quite sure what to rephrase this as. Would it be better as "If the variant is `inventory`, the game loads the model file under models/item/, and if the variant is anything else, it will load a blockstate JSON from blockstates/"

docs/items/overrides.md
howtonotwin: Markdown syntax means `[wiki][]` is short for `[wiki][wiki]`

diesieben07: Even if that's true I am thoroughly confused as to why that works.

howtonotwin: It's just how it is. The definition of Markdown's syntax says that `[a][b]` creates a link with the text `a` that links to the URL specified by the reference `b`, and that `[a][]` is shorthand for `[a][a]`.

diesieben07: Yes, but "wiki" is not a valid url...

howtonotwin: There's a reference at the end of the file.

diesieben07: I... well now... that is way to friggin complicated for a tired man like me.
:sleepy:

howtonotwin: `[wiki][]` means `[wiki][wiki]` (note 4 square brackets). `[wiki]: URL` defines the reference `wiki` to point at `URL`. Therefore, `[wiki][wiki]` points at URL, therefore `[wiki][]` points at URL.

diesieben07: Yes, yes I got it already :P

docs/items/overrides.md
```diff
@@ -1,7 +1,8 @@
...
+When you have an item for which you want to use different models depending on its state, you have two options: use a blockstate JSON, or use item properties. When you use a blockstate JSON for an item, you can directly bind certain metadatas to certain variants, and with an `ItemMeshDefinition` you can use arbitrary code to map `ItemStack`s to variants. However, the ability to use blockstates for items is a Forge addition, and vanilla uses item property overrides. An item property assigns a certain `float` value to every ItemStack it is registered for, and vanilla item model definitions can use these values to define "overrides", where an item defaults to a certain model, but if an override matches, it overrides the model and uses another. The format of item models, including overrides, can be found on the [wiki][].
+
```
diesieben07: Well, you *could* also just use multiple models. But this is better.

docs/items/overrides.md
```diff
@@ -12,21 +12,25 @@ To add a property to an `Item`, simply call `addPropertyOverride` on it. The `Re
...
+  "texture": {
+    "__comment": "Default",
+    "layer0": "examplemod:items/examplePartial"
+  },
+  "overrides": [
     {
-      "parent": "item/generated",
-      "texture": {
-        "layer0": "examplemod:items/examplePartial" // Default
+      "__comment": "Power >= 0.75",
```
tterrag1098: // comments are supported, this should be unnecessary

howtonotwin: On GH it highlights it in red, is it different on RTD?

diesieben07: Are they? Definitely not in the official JSON specification. And Gson only supports them in lenient mode, which is not enabled at least in forge's blockstate loader.

bonii-xx: Minecraft 1.9 changed the json loader from lenient to strict. They're always strict now and don't suport //

tterrag1098: I hate everything about that change. Ah well

williewillus: Blame the JSON spec devs :P
Other parts of strict mode are good to have

howtonotwin:
I've been thinking about taking `overrides`.md and the "Registering Item Models` section out, putting them into a new category "Models", then moving "Blockstates" as a subcategory of that one. There might be some more stuff added too. Thoughts?

williewillus:
What I would do is:
* Under the "Blockstates" section
    * Move "Intro to Blockstates" to section "Blocks". Because blockstates are not just for models and all modders that touch blocks *must* understand how they work.
    * Move the other two articles (Intro to Blockstate jsons and Forge Blockstate Json) to a new "Model" section

howtonotwin:
Good idea.

tterrag1098:
Sorry this sat so long, looks good.

PR #60, opened 2016-08-10
Added instructions to the import for IntelliJ

Puggingtons:
A lot of times I have tried to create my workspace to develop mods with the instructions on this site. Only after I made this modification all the required libraries were correctly imported in IntelliJ.


bonii-xx:
That sounds more like your setup needs a special build script. For a plain workspace you don't need any modifications.

williewillus:
You haven't needed that line for the past year or so at least

PR #61, opened 2016-08-12
Add Models category

howtonotwin:
Break up the "Blockstates" section and put the "Intro to Blockstates" under category "Blocks", while putting the other two pages into "Models".
Document the simplest "outer shell" of the system, which is basically how most people will use models.
Also document the internals of the system, which explains how the classes and interfaces involved work.


tterrag1098:
This will break hardlinks. Would need to add redirects. @bonii-xx is this worth it to you?

docs/models/blockstates/introduction.md
```diff
@@ -20,13 +20,8 @@ As an example, let's take a look at the vanilla `oak_log.json`:
 
 As you can see there is no normal state, only different variants depending on the value of "axis". Depending on which axis the log is aligned it will use either a model of the upright log, a model of the sideways log (rotated by 90° or not) or, should the block not have any axis set, it'll display the bark model which has the bark on all 6 sides.
 
-The log only has one property: axis. A blockstate always has to be defined for all of its properties. This can quickly lead to a combinatorial explosion of variants. Let's look at one variant of the vanilla fence:
+The log only has one property: axis. A blockstate always has to be defined for all of its properties. In 1.8, you had to specify the entire variant string, for each and every variant string. This would lead to a very large number of variants, as each and every combination of properties must be defined. The [Forge blockstate format][Forge blockstate] allows you to keep this under control. 1.9 also introduces the "multipart" blockstate format, which has certain advantages and disadvantages compared to Forge's. You can find a description of 1.9's format on the [wiki].
```
bonii-xx: The 1.9 sentence should start with "Minecraft 1.9". It also needs a bit of a rewording along the lines of explaining that both blockstate types allow for similar things (multiparts) but each of them has specific cases that the other one doesn't cover. Also that neither is better/worse, and that forges blockstates are supported from 1.8 on through 1.10+ and don't end in 1.9.

docs/models/blockstates/introduction.md
```diff
@@ -20,13 +20,8 @@ As an example, let's take a look at the vanilla `oak_log.json`:
...
 
-```json
```
bonii-xx: Keep the example to show the difference

docs/models/overrides.md
```diff
@@ -0,0 +1,36 @@
...
...ent models depending on its state, you have a few options: use a blockstate JSON, several item models, or use item properties. When you use a blockstate JSON or several item models for an item, you can directly bind certain metadatas to certain variants/models, and with an `ItemMeshDefinition` you can use arbitrary code to map `ItemStack`s to variants/models. However, what if you need more flexibility? For example, a bow needs to know how long it has been pulled, which is most easily accessed as a property of the holding entity. Minecraft 1.8 introduced item property overrides. An item property assigns a certain `float` value to every ItemStack it is registered for, and vanilla item model definitions can use these values to define "overrides", where an item defaults to a certain model, but if an override matches, it overrides the model and uses another. The format of item models, including overrides, can be found on the [wiki]. In the example with bows, an item property is especially nice as properties can be continuous values. Instead of hardcoding only 4 slots for bow models, item properties allow you to define as many models as you want on the spectrum of the "pulling" property.
```
bonii-xx: >use a blockstate JSON, several item models, or use item properties
Or a custom baked model

bonii-xx:
Hardlinks from outside you mean? I don't think there are many.

Otherwise, there definitively is more content on the model system now. ice to have item overrides in there too. I'd say if it's feasible and no content is lost. I skimmed over the text, but didn't get to check it in detail. Looked fine overall though, besides the few comments.

docs/models/overrides.md
howtonotwin: Including custom baked models here doesn't really make sense. Those are interchangeable with all other models, and therefore they fall under the category of "several item models".

docs/models/files.md
```diff
@@ -0,0 +1,27 @@
+Model Files
+===========
+
+A "model" is simply a shape. It can be a simple cube, it can be several cubes, it can be a truncated icosidodecahedron, or anything in between. Most models you'll see will be in [JSON] format. Models are loaded into `IModel`s by an `ICustomModelLoader` at runtime. As long as there is a corresponding model loader, all models in all formats are interchangeable.
```
tterrag1098: This is not true, different model loaders can have wildly different syntaxes and purposes.

howtonotwin: I meant that once a model has been loaded, you can use it anywhere without worrying about its format. A blockstate won't care if its models are OBJ, B3D, or some mix, and stuff like that.

howtonotwin:
The important part is in! :smile:

docs/models/blockstates/introduction.md
```diff
@@ -20,7 +20,25 @@ As an example, let's take a look at the vanilla `oak_log.json`:
...
+
+Here's an excerpt from the same file in 1.9, which uses the multipart format:
+
+```json
+{ "when": { "east": "true" },
+  "apply": { "model": "oak_fence_side", "y": 90, "uvlock": true }
+}
+```
+
+This is one clause of 5. You can read this as "when east=true, use the model oak_fence_side rotated 90 degrees". This allows the final model to be built up from 5 smaller parts, 4 of which (the connections) are conditional and the 5th being the unconditional central post. This uses only two models, one for the post, and one for the side connection.
```
tterrag1098: clause -> case

docs/models/blockstates/introduction.md
```diff
@@ -1,9 +1,7 @@
...
-A simple block with no properties only has a "*normal*" blockstate which is its default.
-A more complex block that can be displayed in different ways has so called *variants*.
+Blockstate JSONs are Minecraft's way to map "variant strings" in models. A variant string can be absolutely anything, from "inventory" to "power=5" to "I am your father." When the game searches for a model corresponding to a block in the world, it takes the [blockstate] for that position, and then it uses that to find the corresponding blockstate JSON and the variant string within it. The default is to use the block's registry name as the location of the blockstate JSON (block `examplemod:testblock` goes to JSON `examplemod:testblock` goes to filepath `assets/examplemod/blockstates/testblock.json`), and the variant string is pieced together from the blockstate's properties. More information can be found [here][statemapper]. The combination of a resource location (like the path to a blockstate JSON, e.g. `examplemod:blockstate`) and a variant string is represented with a `ModelResourceLocation`.
```
williewillus: this is not fully accurate.

1. IBlockState -> MRL is not what Blockstate JSONs do, That is the job of the statemapper. you're mixing the two together here.
2. I'm not sure variant strings can be *anything*

docs/models/using.md
```diff
@@ -15,6 +15,7 @@ Blocks are not directly linked to models, instead, block<em>states</em> are mapp
 5. Get the name of each value with `IProperty<T>#getName(T)`.
```
williewillus: add a note here that the property names are sorted alphabetically.

howtonotwin: I did, on L20 :P

williewillus: oh, didn't see that. probably would be better to put it in line here. on item 5. or 6.

docs/models/blockstates/introduction.md
howtonotwin: I just tested setting the variant string to `Luke, I AM your father!`, and `"/;:',[]{}()*&^%$#@!<>?|\..`, and `�TBg���p�`. I pulled the last one straight out of `/dev/urandom`. All 3 worked just fine on items, so if there is a limit to that side, I'm not hitting it. On the other hand, whatever is registering block models lowercases the string first, so there's that.

I set the item to a variant in a blockstate with setCustomMRL, and I set the block's with a setCustomStateMapper.

howtonotwin:
Thanks to SquareWheel on IRC for nitpicking everything!

docs/models/overrides.md
williewillus: Minecraft 1.9 introduced overrides not 1.8

docs/models/overrides.md
```diff
@@ -0,0 +1,36 @@
...
+  "overrides": [
+    {
+      "__comment": "power >= .75",
+      "predicate": { 
+        "power": 0.75
+      },
+      "model": "examplemod:item/examplePowered"
+    }
+  ]
+}
```
williewillus: maybe show a stub of what the IItemPropertyGetter would look like for this

docs/models/files.md
```diff
@@ -0,0 +1,40 @@
...
+
+Block and item models differ in a few ways, the major one being [item property overrides][overrides].
+
+When referring to textures in models, keep a few things in mind. First, you reference textures as RLs relative to `textures/`. (E.g. `examplemod:blocks/test` → `assets/examplemod/textures/blocks/test.png`.) Second, when specifying [UV coordinates][UV], (0,0) is taken to mean the **top-left** corner. UVs are *always* from 0 to 16. If the texture is larger or smaller, it is scaled to fit. A texture must also be square, and its side length must be a power of two (e.g. 1x1, 2x2, 4x4, 8x8, but not 4x8). If an mcmeta file is present, the image can be rectangular in order to fit the extra image data, but the final texture as seen externally is still a square (e.g. the command block texture is animated and the on-disk file is rectangular, but the model uses UVs (0,0) to (16,16) because the texture is still seen as square.)
```
williewillus: a bit unclear at first read 
Maybe replace the last super long sentence with something like

"If the texture is animated, the image can be rectangular, and is interpreted as sequential regions of UV(0,0) to UV(16,16) from top to bottom"

docs/models/introduction.md
```diff
@@ -0,0 +1,8 @@
+Intro to Models
+===============
+
+The model system is Minecraft's way of giving blocks and items their shapes. Through the model system, blocks and items are mapped to their models, which define how they look. One of the main goals of the model system is to allow not only textures but the entire shape of a block/item to be changed by resource packs. Indeed, when you create a mod, you are also creating a mini-resource pack for your own blocks and items.
+
+In order to link code to models and textures on disk, there exists the class `ResourceLocation` (RL). You may recognize them from the registry system, however their original vanilla purpose was to identify files on disk; they just happened to be useful as unique identifiers in Forge. An RL is a very simple object composed of 2 `String`s, a domain and a path. When an RL is represented as a plain string, it looks like `domain:path`. Normally, if an RL is being read from a string that doesn't have a domain, the domain defaults to `minecraft`.
```
williewillus: maybe note that domains are often/must be the mod id

docs/models/using.md
```diff
@@ -0,0 +1,133 @@
+
+Using Models
```
williewillus: very general name...

docs/models/using.md
```diff
@@ -0,0 +1,133 @@
...
+
+Now if we ask `mapper` to find the MRL for the blockstate `examplemod:block1[facing=east]`, it will map to `examplemod:east#normal`. If we ask it to map `examplemod:block2[facing=north,color=red]`, it will map to `examplemod:north#color=red`.
+
+##### `setSuffix`
+
+The suffix is a plain string that gets tacked on to the end of the resource path. So if the suffix is set to `_suff`, the `IStateMapper` will map the blockstate `examplemod:block[facing=east]` to the MRL `examplemod:block_suff#facing=east`.
+
+##### `ignore`
+
+This is also fairly self-explanatory. The `IStateMapper` will simply ignore the given properties when mapping a blockstate. Note that calling `ignore` twice will simply combine the two lists, and that a blockstate without one of the ignored properties will not cause an error in the mapper.
```
williewillus: include a short example like the two above

docs/models/using.md
```diff
@@ -0,0 +1,133 @@
...
+
+### Blockstate JSONs for Items
+
+Note that *items* can use *block*state JSONs. This is possible by simply passing an MRL pointing to a blockstate JSON into `ModelLoader.setCustomeModelResourceLocation` or returning it from an `ItemMeshDefinition`. Doing so allows you to take advantage of things like submodels (and combining variants if you use the Forge format) for item models. In the Forge format, this is especially useful for the default item layer model as you can use the `textures` block inside combining variant definitions to build up the layers of the model. Note that there is a major caveat. First, blockstate JSONs can only refer to models under `models/block`; they cannot see models under `models/item`. This means that you can't use `minecraft:item/generated` to get the default item transforms (proper rotation and scale). In a Forge blockstate JSON, you can use the model `minecraft:builtin/generated` and set the `transform` property to `forge:default-item` to work around this. Here's an example:
```
williewillus: this is a mouthful esp if you're a beginner. try breaking it up a bit or removing/reducing some of the fluff about builtin/generated

docs/models/blockstates/introduction.md
```diff
@@ -0,0 +1,46 @@
...
+
+A blockstate always has to be defined for all possible variant strings. When you have many properties, this results in lots of possible variants, as each and every combination of properties must be defined. In Minecraft 1.8's blockstate format, you have to define each and every string explicitly, which leads to long, complicated files. It also doesn't support the concept of submodels, or multiple models in the same blockstate. In order to allievate this, Forge introduced its [own blockstate format][Forge blockstate], which is available in Minecraft 1.8 and up.
+
+Starting from Minecraft 1.9, Mojang also introduced the "multipart" format. You can find a definition of its format on the [wiki][]. Forge's format and the multipart format are not better than each other; they each cover different use cases and it is your choice which one you want to use. Both formats take care of calculating the set of all possible variants for you behind the scenes, so your files can be simpler.
```
williewillus: don't know if we want to talk about this here but the two systems actually work very differently.

Forge version is just syntax sugar to pregenerate all possible combos.
1.9 multipart bakes every part and then decides *at runtime* which lists of quads to return based on the IBlockState. it's basically a json definition for a smartmodel.

docs/models/using.md
```diff
@@ -0,0 +1,133 @@
...
+```json
+{
+  "multipart": [
+    { "apply": "examplemod:item/core", "__comment": "Transforms come from here." },
+    { "when": { "floater": "cube" },
+      "apply": "examplemod:floating_cube",
+      "__comment": "etc." }
+  ]
+}
+```
```
williewillus: does this actually work?
1.9 multiparts dispatch on runtime based on IBlockState, *not* the MRL variant name, iirc.

williewillus: yeah I don't think this actually works.

code from `MultipartBakedModel` in vanilla. Null IBlockState will return empty model.

```Java
public List<BakedQuad> getQuads(@Nullable IBlockState state, @Nullable EnumFacing side, long rand)
    {
        List<BakedQuad> list = Lists.<BakedQuad>newArrayList();

        if (state != null)
        {
            for (Entry<Predicate<IBlockState>, IBakedModel> entry : this.selectors.entrySet())
            {
                if (((Predicate)entry.getKey()).apply(state))
                {
                    list.addAll(((IBakedModel)entry.getValue()).getQuads(state, side, rand++));
                }
            }
        }

        return list;
    }
```

WesCook:
Not sure when you'll be back in IRC, so I'll drop these notes here.  Hopefully is addressing the correct version of the docs.

http://pastebin.com/LdiBDMS2

howtonotwin:
Would it be a good idea to add an "Examples" section? The way I imagine it there'd be gists on each example page containing the files, which would be properly organized into the required file structure. The examples may be sorted by complexity, with the simplest being an almost self-contained mod adding no more than one item, one block, and the corresponding `ItemBlock`, and the advanced ones based on the (currently WIP) "Advanced Models" section I'm working on. As the examples get more complex they also become less "handed on a silver platter" in the sense that they are no longer entire self-contained mods; only the relevant model related parts are included.

I say this because the model system does seem to be confusing to many, no matter how much they read, as it is just too complicated. I think examples, simple at first and then progressively more complex, with proper comments, might help solidify their understanding.

WesCook:
Yes, absolutely it would help.  The Forge docs make great effort to explain the concepts of the system, but there's little help provided in putting it all together.  For instance, seeing an example of where to create a .json and how to reference it is so much easier than understanding and following all the different rules the docs explain.  It's still good to know those rules, but then you have a better framework for actually understanding them.

I'm actually going to suggest taking that concept even further.  The docs are currently laid out as a reference guide.  We discussed this a bit, but my belief is if the beginning chapters are structured closer to a tutorial format (eg. creating your first item/block), they'll be made much more accessible to those getting started.  It lets you set a slower pace where people can learn the concepts as they go, rather than having them dumped on you all at once.

Is a dictionary useful when you're just learning to speak?  Or is it only useful once you have a framework for understanding it, and can reference exactly what you need to know?

The blocks page for instance - where you actually start coding - immediately starts off by teaching you `setHardness()` and other properties.  It's just not useful knowledge yet.  Stepping the reader through creating a new block, then explaining *what* you can do with that block is a much more intuitive flow.

At least, that was my takeaway when reading through the docs originally.  I realize it's an ambitious restructuring, but I do believe it would be significantly easier to use and understand.

Regarding the examples, I think if you were to have a format as I described above, it would then make more sense to have links to those examples provided as you go through the docs organically.  So after being taught the steps to create a new block, you then have the full reference of a self-contained mod to do it yourself.  The first one might even make sense as a Github sample project, so you can see exactly the directory structure to follow.  As you go along it could get simpler/less hand holdy, as you mentioned.

Otherwise, an index or separate section as you described would certainly be helpful too.

Particularly with models, there seem to be a lot of common problems.  So I'd also suggest a troubleshooting section for going over them.  Things like: JSON in right folder, correct parent set for items/blocks, not lowercasing variant names, not putting them in alphabetical order.  There's quite a few gotchas and it'd be easier to debug if they're in one area, rather than spaced out through the docs.

Let me know if you think that's overly ambitious.  I don't mean to step on any toes -- I realize the original authors probably wrote it in this format for a reason -- but if there's interest I'd be happy to do my part in rewriting those pages.

And if not, well, I think the examples are a great idea too.

LexManos:
No, our documentation will not have copy/paste 'examples' This is a staple in the design of this entire site.
If you want copy pasta go to the millions of other 'tutorials'.

williewillus:
I agree. Copy paste examples encourage copying without first understanding, which is the problem that this site is setting out to solve in the first place.

tterrag1098:
> The blocks page for instance - where you actually start coding - immediately starts off by teaching you setHardness() and other properties. It's just not useful knowledge yet. Stepping the reader through creating a new block, then explaining what you can do with that block is a much more intuitive flow.

If you are just starting coding when you read these docs you have already broken the guidelines set out in the main page. This guide is not targeted at those new to coding. It is targeted at those who have a moderate to good understanding of both the game and the language. 

With those prerequisites in mind, the guide is very useful, as it outlines what you can do with blocks, and how to do those things. You are confusing "helpful" with "easy". If you truly read and understand the guide, it is trivial to put into practice.

howtonotwin:
Have some more! 😄

docs/models/advanced/imodelstate+part.md
```diff
@@ -0,0 +1,14 @@
...
+
+One of the uses of this is animation. An `IModelState` can represent a certain frame of the animation and supply transforms to turn the original model into the current frame. By supplying different `IModelState`s over time, an animation can be performed. For example, the B3D model format supports this kind of animation directly through its nodes; however, the animation system is still WIP.
+
+Another, more common use case is Forge blockstate JSONs. The models within the blockstates can be transformed with `transform` tags, which translate into simple `IModelState`s that are passed into the contained models.
+
+Finally, another use case is [perspective aware models][].
+
+Which `IModelPart`s a certain model will use is dependent on the model itself. If I had a `B3DState` that dealt with B3D `NodeJoint`s and tried to use it on a vanilla JSON model, it wouldn't work as vanilla models have no idea what a `NodeJoint` is and will not even ask about them.
```
williewillus: weird to say "Finally" the sentence before but then keep writing afterward

docs/models/advanced/introduction.md
```diff
@@ -0,0 +1,30 @@
...
+
+In order to understand how this works, let's go through the internals of the model system. Throughout this section, you will probably have to refer back to this to grasp a clear understanding of what is happening. This is true in reverse as well. You will likely not understand everything happening here, but as you move through this section you should be able to grasp more and more of it until everything is clear.
+
+!!! important
+    If this is your first time reading through do not skip _anything_! It is _imperative_ that you read everything **in order** so as to build a comprehensive understanding! In the same vein, if this is your first time reading, do not follow links if they lead to pages further ahead in the section.
+
+1. A set of `ModelResourceLocation`s are marked as models to be loaded through `ModelBakery`.
+    * For items, their models must be manually marked for loading with `ModelBakery.registerItemVariants`. (`ModelLoader.setCustomMRL` does this.)
```
williewillus: replace references to ModelBakery here and elsewhere with ModelLoader. ModelLoader is forge's subclass to ModelBakery, and it's simpler to have it be consistent even if the method isn't actually in ModelLoader

docs/models/advanced/itemoverridelist.md
```diff
@@ -0,0 +1,48 @@
...
+
+Given a list of `ItemOverride`s, the constructor copies that list and stores the copy. The list may be accessed with `getOverrides`, and it is used to implement the vanilla `applyOverride`, which, in turn, is used in the vanilla `handleItemState`.
+
+### <s>`applyOverride`</s>
+
+This is a deprecated vanilla method. It is only called in the vanilla `handleItemState`, and in almost all cases can be safely ignored.
+
+### `handleItemState`
+
+This simply takes an `IBakedModel`, an `ItemStack`, a `World`, and an `EntityLivingBase` to produce another `IBakedModel` to use for rendering. This method should not mutate the world in any way.
```
williewillus: mention that if modders want a custom item baked model this is where the interesting magic happens. handleItemState is called every time a baked model is rendered in an item context.

docs/models/overrides.md
```diff
@@ -0,0 +1,52 @@
+Item Property Overrides
```
williewillus: consider merging this with the article over the actual class?

williewillus:
also, I might have missed it, but make sure you give the actual true definition of what a ModelResourceLocation is. It's imo one of the most important parts of understanding the model system.

docs/models/advanced/extended-blockstates.md
```diff
@@ -0,0 +1,19 @@
...
+
+Ordinary block state containers (`BlockStateContainer`) take a set of listed properties (`IProperty<V>`) that define all possible values for themselves and use those to create all possible combinatons of values. These combinations are then turned into `IBlockState`s and stored in the finite set of all possible states. The properties are called "listed" as they appear on the F3 debug screen, all the way to the right.
+
+Extended block state containers take a set of listed properties, and also a set of *unlisted* properties. Unlisted properties do not define all of their possible values, but instead provide a predicate for checking if a value is valid. The listed properties are again used to create the set of states, this time `IExtendedBlockState`s, but the unlisted properties remain unset. The unlisted properties are only set when `IExtendedBlockState::withProperty` is called to do so. This allows unlisted properties to have any type, even other `IBlockState`s.
```
howtonotwin: > An IUnlistedProperty still only has *1* value in the IExtendedBlockState.

I never implied the opposite, and I'm confused what part of what I wrote made you think that.

> `isValid` does not have to do with not having fixed values, it's just a safety check used in `withProperty`.

`withProperty` is rejecting values that fail the `isValid` check with this message:

```java
throw new IllegalArgumentException("Cannot set unlisted property " + property + " to " + value + " on block " + Block.REGISTRY.getNameForObject(getBlock()) + ", it is not an allowed value");
```

which implies, to me, that it's a predicate that checks whether a value is allowed for the property. I'm trying to contrast this against the fixed set of values `IProperty` uses.

docs/models/advanced/imodel.md
```diff
@@ -0,0 +1,159 @@
...
+This is a collection of the `ResourceLocation`s of all the textures this model depends on. These textures are guaranteed to be loaded before this model is baked. For example, a vanilla JSON model depends on all the textures defined within.
+
+### `bake`
+
+This is the main method of `IModel`. It takes an `IModelState`, a `VertexFormat`, and a function `ResourceLocation` → `TextureAtlasSprite`, to return an `IBakedModel`. `IBakedModel` is lower-level than `IModel`, and it is what interacts with blocks and items. The function RL → TAS is used to get textures from RLs (i.e. the RLs of textures are passed to this function and the returned TAS contains the texture).
+
+`IAnimatedModel`
+----------------
+
+`IAnimatedModel` is part of the currently WIP animation API. Right now it does nothing.
```
howtonotwin: I saw no references to it when I searched, but oh well. Will a "TODO: explain" do, or should I just replace the entire section with a "TODO: `IAnimatedModel`".

docs/models/advanced/imodel.md
```diff
@@ -0,0 +1,159 @@
...
+    "dying": {
+      "true": {
+        "__comment": "Custom data is inherited. Therefore, here `meaningOfLife` is inherited but `showQuestion` is overriden. The model itself remains inherited.",
+        "custom": {
+          "showQuestion": true
+        }
+      },
+      "false": {}
+    },
+    "book": {
```
howtonotwin: I wanted to show it worked in submodels too, and didn't think of a way to cram it into `"false"`, but if you think its unnecessary I'll get rid of it.

howtonotwin:
It's in ["Introduction to Blockstate JSONs"](https://github.com/howtonotwin/MCForgeDocumentation/blob/models/docs/models/blockstates/introduction.md); should I move it?

docs/models/advanced/extended-blockstates.md
```diff
@@ -0,0 +1,19 @@
...
+
+The methods on `IExtendedBlockState` are self-explanatory, perhaps with the exception of `getClean`. All `getClean` does is return the base `IBlockState` with none of the unlisted properties set.
+
+Using Extended Blockstates
+--------------------------
+
+A block using `ExtendedBlockState` simply constructs and returns one from its `createBlockState`. The constructor takes the set of listed and unlisted properties. Unlisted properties themselves may be created by implementing the interface in a class and instantiating it, or by wrapping a listed property in one through `Properties.PropertyAdapter`/`Properties.toUnlisted`. Unlisted properties should have default values set in the block's default blockstate, which must be done by casting the `Block::getDefaultState` to `IExtendedBlockState`, setting the properties, and then setting it as the default with `setDefaultState`.
+
+In order for extended states to be actually useful, the block must also override `getExtendedState`, which is similar to `getActualState` in that it enriches a blockstate by filling in the values for unlisted properties. This returned `IBlockState` is what is passed into `IBakedModel::getQuads`.
```
howtonotwin: That *was* a rather large oversight on my part; thanks for catching that.

docs/models/overrides.md
howtonotwin: I want to keep the separate, because the adv model section is meant for those who want to actually go in and do fancy things with the system, while the base models section is just for those who want to slap some JSON together and just "have it work."

docs/models/advanced/ibakedmodel.md
```diff
@@ -0,0 +1,42 @@
+`IBakedModel`
```
howtonotwin: Done

docs/models/advanced/extended-blockstates.md
williewillus: yes but just because there's a validator doesn't mean anything in the context of fixed vs unfixed values. I could have a validator on the fixed value set too

howtonotwin: Oh I see. My wording was a bit unclear. I didn't mean to imply that "having a validator" <=> "not being a fixed value set." I merely wanted to provide some contrast. I'll get to that tomorrow.

docs/models/advanced/imodel.md
```diff
@@ -0,0 +1,159 @@
...
+
+This is a collection of the `ResourceLocation`s of all the models this model depends on. These models are guaranteed to be loaded before this one is baked. For example, a model deserialized from a blockstate JSON will depend on the models defined within. Only models that are directly mapped to a block/item are loaded normally; to ensure loading of other models, they must be declared as dependencies of another. Cyclic dependencies will cause a `LoaderException` to be thrown.
+
+### `getTextures`
+
+This is a collection of the `ResourceLocation`s of all the textures this model depends on. These textures are guaranteed to be loaded before this model is baked. For example, a vanilla JSON model depends on all the textures defined within.
+
+### `bake`
+
+This is the main method of `IModel`. It takes an `IModelState`, a `VertexFormat`, and a function `ResourceLocation` → `TextureAtlasSprite`, to return an `IBakedModel`. `IBakedModel` is lower-level than `IModel`, and it is what interacts with blocks and items. The function RL → TAS is used to get textures from RLs (i.e. the RLs of textures are passed to this function and the returned TAS contains the texture).
```
howtonotwin: I'll be writing something on the deeper parts models Soon™, so I'll leave the `VertexFormat` bare for right now.

tterrag1098:
After the merge conflicts are resolved, is this PR ready for a proper review?

docs/models/blockstates/forgeBlockstates.md
```diff
@@ -0,0 +1,140 @@
...
+    },
+    "model": "cube_all",
+    "uvlock": true
+  },
+  "variants": {
+    "normal": [{}]
+  }
+}
+```
+
```
Alpvax: Could you please include the information from Jay Avery's very informative post [here](http://www.minecraftforge.net/forum/topic/57153-1102-issue-with-fluid-json-files/?do=findComment&comment=266664), regarding fully defined and partially defined variants.

docs/models/blockstates/forgeBlockstates.md
```diff
@@ -0,0 +1,140 @@
...
+    "model": "cube_all",
+    "uvlock": true
+  }
+```
+
+The defaults section contains the default values for all variants. They can be overwritten by the variants. The defaults section is **optional**! You do not need to define defaults, the block can be omitted altogether.
+
+```json
+  "variants": {
+    "normal": [{}]
```
Alpvax: Could you please include the information from Jay Avery's very informative post [here](http://www.minecraftforge.net/forum/topic/57153-1102-issue-with-fluid-json-files/?do=findComment&comment=266664), regarding fully defined and partially defined variants.

howtonotwin: Done

howtonotwin:
So I *still* don't have anything on the *really* low level stuff (like `VertexBuffer`), and I know nothing about the Animation API, but I think this works for everything above that level of abstraction. If there are no further changes to be made, this is good for merge, and someone else (or I, if I feel like it) can add that in later.

docs/models/advanced/extended-blockstates.md
```diff
@@ -0,0 +1,61 @@
...
+
+Most of the methods on `IExtendedBlockState` are self-explanatory, maybe with the exception of `getClean`. `getClean` returns the base `IBlockState` with none of the unlisted properties set.
+
+Why Extended Blockstates
+------------------------
+
+Why use extended blockstates at all? Why not simply pass an `IBlockAccess` and `BlockPos` into the rendering system directly and have the [`IBakedModel`][IBakedModel] itself deal with it? Indeed, extended blockstates allow this to happen anyway! Why have the system at all when we could just do that? The reason is twofold.
+
+First, rendering is threaded, meaning that the world continues to chug along on the main thread while models are being rendered. However, in Minecraft, `World`s are simply not made for concurrent access. This makes the above solution downright *dangerous*, as it would entail multiple threads concurrently accessing the world. Therefore, extended states are used to collect all the data required for the model safely into the `IExtendedBlockState`. These objects can now be passed on to different threads.
```
tterrag1098: `getExtendedState` is called on the same thread, directly before getQuads, so this is a pretty weak justification. There is still the same threading concerns inside extended state methods.

docs/models/advanced/ibakedmodel.md
```diff
@@ -0,0 +1,48 @@
...
+
+### <s>`getItemCameraTransforms`</s>
+
+Deprecated in favor of implementing `IPerspectiveAwareModel`. If a model implements that interface this method will not be called, in favor of an `instanceof` check and a call to `IPerspectiveAwareModel::handlePerspective`. Therefore, most `IBakedModel`s that also implement `IPerspectiveAwareModel` should `return ItemCamerTransforms.DEFAULT` here. See [Perspective][].
+
+### `getOverrides`
+
+Returns the [`ItemOverrideList`][ItemOverrideList] to use for this model. This is only used if this model is being rendered as an item.
+
+### `getQuads`
```
tterrag1098: I would say this section should go first

docs/models/advanced/ibakedmodel.md
```diff
@@ -0,0 +1,48 @@
...
+
+This is the main method of `IBakedModel`. It returns `BakedQuad`s, which contain the low-level vertex data that will be used to render the model. If the model is being rendered as a block, then the `IBlockState` passed in is non-null. Additionally, when applicable, [`Block::getExtendedState`][extended blockstates] is called to create the passed `IBlockState`, which allows for arbitrary data to be passed from the block to the model.
```
tterrag1098: `getExtendedState` is *always* called.

Additionally, it's worth mentioning that this method is called for every layer the block supports, and for every side + null. So that can be anywhere from 7 to 28 times per render. Thus this method should be heavily optimized.

docs/models/advanced/ibakedmodel.md
```diff
@@ -0,0 +1,48 @@
...
+
+The `EnumFacing` passed in is used for face culling. If the block against the given side of the block being rendered is opaque, then the faces associated with that side are not rendered. If that parameter is `null`, all faces not associated with a side are returned (that will never be culled).
+
+The `long` parameter is a random number.
```
tterrag1098: Not just a random number, but a random number seeded by the world position.

docs/models/advanced/imodel.md
```diff
@@ -0,0 +1,112 @@
...
+
+### `getTextures`
+
+This is a collection of the `ResourceLocation`s of all the textures this model depends on. These textures are guaranteed to be loaded before this model is baked. For example, a vanilla JSON model depends on all the textures defined within.
+
+### `bake`
+
+This is the main method of `IModel`. It takes an [`IModelState`][IModelState], a `VertexFormat`, and a function `ResourceLocation` → `TextureAtlasSprite`, to return an [`IBakedModel`][IBakedModel]. `IBakedModel` is less abstract than `IModel`, and it is what interacts with blocks and items. The function RL → TAS is used to get textures from RLs (i.e. the RLs of textures are passed to this function and the returned TAS contains the texture).
+
+### `process`
```
tterrag1098: This is not a member of `IModel`

docs/models/advanced/imodel.md
```diff
@@ -0,0 +1,112 @@
...
+      },
+      "false": {}
+    }
+  }
+}
+```
+
+As seen above, custom data can be of any type. Additionally, it is inherited from the defaults into the variants. The custom data is passed in as an `ImmutableMap<String, String>`. This is a map where the keys are the property names (in the above example, "meaningOfLife", "showQuestion", and "title"). Astute observers may notice that numeric and boolean data were defined in within the blockstate but this method only receives `String`s. This is because all data is converted into strings before being processed. If a model does not understand what a property means, it should just ignore it.
+
+### `smoothLighting`
```
tterrag1098: This is not a member of `IModel`

docs/models/advanced/imodel.md
```diff
@@ -0,0 +1,112 @@
...
+
+In vanilla, smooth lighting enables ambient occlusion. This flag can be controlled by the `smooth_lighting` property in a Forge blockstate (which can appear wherever a `model` property can and is inherited). The default implementation does nothing.
+
+### `gui3D`
```
tterrag1098: This is not a member of `IModel`

docs/models/advanced/imodel.md
```diff
@@ -0,0 +1,112 @@
...
+
+`gui3D` controls whether a model looks "flat" in certain positions (e.g. with `gui3d` set to `true`, `EntityItem` renders a stack with multiple items as several layers of the model. With `gui3d` set to `false`, the item is always one layer), and also controls lighting inside GUIs. This flag can be controlled by the `gui3d` property in a Forge blockstate. The default implementation does nothing.
+
+### `retexture`
```
tterrag1098: This is not a member of `IModel`

docs/models/advanced/imodel.md
```diff
@@ -0,0 +1,112 @@
...
+      "varZ": null
+    },
+    "model": "examplemod:universe"
+  }
+}
+```
+
+In this example, the `textures` block will be deserialized as-is into an `ImmutableMap` with the exception that `null`s are turned into `""` (i.e. the final result is `"varA" → "examplemod:items/hgttg", "varB" → "examplemod:blocks/earth", "varC" → "#varA", "varZ" → ""`). Then, `retexture` is called to change the textures as needed. How this is done is up to the model. It may be advisable, however, to support resolving texture variables such as "#var" (like vanilla JSON models) instead of taking them literally. The default implementation does nothing.
+
+### `uvlock`
```
tterrag1098: This is not a member of `IModel`

docs/models/advanced/introduction.md
```diff
@@ -0,0 +1,49 @@
...
+
+In order to understand how this works, let's go through the internals of the model system. Throughout this section, you will probably have to refer to this to grasp a clear understanding of what is happening. This is true in reverse as well. You will likely not understand everything happening here, but as you move through this section you should be able to grasp more and more of it until everything is clear.
+
+!!! important
+    If this is your first time reading through do not skip _anything_! It is _imperative_ that you read everything **in order** so as to build a comprehensive understanding! In the same vein, if this is your first time reading, do not follow links if they lead to pages further ahead in the section.
+
+1. A set of `ModelResourceLocation`s are marked as models to be loaded through `ModelLoader`.
+
+    * For items, their models must be manually marked for loading with `ModelLoader.registerItemVariants`. (`ModelLoader.setCustomMRL` does this.)
```
tterrag1098: setCustomMRL -> setCustomModelResourceLocation

Don't use shorthand when it hasn't been defined

docs/models/advanced/introduction.md
```diff
@@ -0,0 +1,49 @@
...
+
+    * For blocks, their statemappers produce a `Map<IBlockState, MRL>`. All blocks are iterated and then the values of this map are marked to be loaded.
```
tterrag1098: Again, what is an MRL? Where has that acronym been defined?

docs/models/advanced/perspective.md
```diff
@@ -0,0 +1,61 @@
+Perspective
```
tterrag1098: This whole section needs reworking as `IPerspectiveAwareModel` no longer exists

docs/models/blockstates/forgeBlockstates.md
```diff
@@ -0,0 +1,203 @@
...
+      "normal": { "model": "some:model" }
+    }
+```
+
+This kind of definition does not interact with the property definitions. Instead, it defines a variant "normal" directly, without merging with the others. It still inherits from a "defaults" block, if present. If the variant is defined as a list, then each element is a variant definition, and the one that will be used is random:
+
+```json
+    "defaults": { "model": "some:model" }
+    "variants": {
+      "__comment": "When used, the model will have a 75% chance of being rotated.
```
tterrag1098: Need a closing quote and comma here

docs/models/blockstates/forgeBlockstates.md
```diff
@@ -0,0 +1,203 @@
...
+    }
+```
+
+A property definition is disambiguated from a straight variant by the type of the first entry. If the first entry of `variants.<something>` is an object, then it is a property definition. If it is anything else, it is a straight variant. In order to avoid mixups, it is recommended to wrap straight variants in a list with one element:
+
+```json
+   "variants": {
+     "simple": [{
+       "custom": {},
+       "model": "some:model"
```
tterrag1098: Need a comma here

docs/models/blockstates/introduction.md
```diff
@@ -0,0 +1,52 @@
...
+    "axis=z":   { "model": "oak_log_side" },
+    "axis=x":   { "model": "oak_log_side", "y": 90 },
+    "axis=none":   { "model": "oak_bark" }
+  }
+}
+```
+
+Here we define 4 variant strings, and for each we use a certain model, either the upright log, the sideways log (rotated or not), and the all bark model (this model is not seen normally in vanilla; you have to use `/setblock` to create it). Since logs use the default `IStateMapper`, these variants will define the look of a log depending on the property `axis`.
+
+A blockstate always has to be defined for all possible variant strings. When you have many properties, this results in lots of possible variants, as every combination of properties must be defined. In Minecraft 1.8's blockstate format, you have to define every string explicitly, which leads to long, complicated files. It also doesn't support the concept of submodels, or multiple models in the same blockstate. In order to allievate this, Forge introduced its [own blockstate format][Forge blockstate], which is available in Minecraft 1.8 and up.
```
tterrag1098: I'm not sure how much of this is even yours, but this page should probably at least mention vanilla multipart blockstate files. Really this page should just mostly be a link to the MC wiki about models.

docs/models/advanced/ibakedmodel.md
howtonotwin: That feels too much like an implementation detail. `IBakedModel`s aren't really supposed to know about `World`s, are they?

docs/models/advanced/perspective.md
howtonotwin: Same story as with `IModel`. The interface lives on as a default noop method in `IBakedModel`. I'll correct the references to the interface, and then could you check this page more thoroughly?

docs/models/blockstates/introduction.md
howtonotwin: The next paragraph is about that and contains a link to the wiki.

howtonotwin:
Can you hold off on merging? I think the colors thing should be moved into `models`.

howtonotwin:
This should be it, for now.

tterrag1098:
I think it's time. I will go through tonight and fix any info that's become outdated since this was proposed.

PR #70, opened 2016-10-11
Add section on setting up with IntelliJ IDEA without using a terminal

gigaherz:


tterrag1098:
I'll assume this is correct as I don't use IDEA. Problems get forwarded to you ;)

elifoster:
@tterrag1098 It's correct, don't worry :smile_cat:

PR #73, opened 2016-10-30
Document registry system

howtonotwin:
Document what registries are, what they do, how to create them, and how to register objects to them. Also document the new registry events as the preferred method of registering objects.

Update other pages to refer to the new page when referring to registration.

### DIFF
--- a/docs/blocks/blocks.md
+++ b/docs/blocks/blocks.md
@@ -12,7 +12,7 @@ For simple blocks, which need no special functionality (think cobblestone, wood 
 
 - `setHardness` - Controls the time it takes to break the block. It is an arbitrary value. For reference, stone has a hardness of 1.5, and dirt 0.5. If the block should be unbreakable, a convenience method `setBlockUnbreakable` is provided.
 - `setResistance` - Controls the explosion resistance of the block. This is separate from hardness, but `setHardness` will also set the resistance to 5 times the hardness value, if the resistance is any lower than this value.
-- `setSoundType` - Controls the sound the block makes when it is punched, broken, or placed. Requires a `SoundType` argument, see the [sounds] page for more details.
+- `setSoundType` - Controls the sound the block makes when it is punched, broken, or placed. Requires a `SoundType` argument, see the [sounds][] page for more details.
 - `setLightLevel` - Controls the light emission of the block. **Note:** This method takes a value from zero to one, not zero to fifteen. To calculate this value, take the light level you wish your block to emit and divide by 16. For instance a block which emits level 5 light should pass `5 / 16f` to this method.
 - `setLightOpacity` - Controls the amount light passing through this block will be dimmed. Unlike `setLightLevel` this value is on a scale from zero to 15. For example, setting this to `3` will lower light by 3 levels every time it passes through this type of block.
 - `setUnlocalizedName` - Mostly self explanatory, sets the unlocalized name of the block. This name will be prepended with "tile." and appended with ".name" for localization purposes. For instance `setUnlocalizedName("foo")` will cause the block's actual localization key to be "tile.foo.name". For more advanced localization control, a custom Item will be needed. We'll get into this more later.
@@ -27,26 +27,19 @@ Of course, the above only allows for extremely basic blocks. If you want to add 
 Registering a Block
 -------------------
 
-So now that you have this block you've created, it's time to actually put it in the game. Thankfully this is extremely simple to do.
-
-As with most things, blocks can be registered using `GameRegistry.register(...)`. For this method your block must have a "registry name" which is just another way of saying "unique name". The preferred method to set your block's registry name is by calling `setRegistryName` on it inside the call to `register`. For instance, `GameRegistry.register(myBlock.setRegistryName("foo"))`.
+Blocks must be [registered][registering] to function.
 
 !!! important
-
-    There is a very large difference between a block in the world and a block in your inventory. A block in the world is governed by an instance of `Block`, it has a blockstate, a 4-bit meta, and maybe a tile entity. Meanwhile, an item in your inventory is controlled by an instance of `Item`, with a 16-bit damage/meta, and maybe an NBT tag. As a bridge between the different worlds of `Block` and `Item`, we have the class `ItemBlock`. `ItemBlock` is a subclass of `Item` that has a field `block` that holds a reference to the `Block` it represents. `ItemBlock` defines some of the behavior of a "block" as an item, like how a right click places the block. It's possible to have a `Block` without an `ItemBlock`. (E.g. `minecraft:water` is a block, but not an item.)
     
-    When you register a block, you *only* register a block. The block does not automatically have an `ItemBlock`. In order to give your block an `ItemBlock`, you should construct an `ItemBlock` from it and register that as well. The simplest way to do this like this: `GameRegistry.register(new ItemBlock(myBlock).setRegistryName(myBlock.getRegistryName()))`.
+    A block in the world and a "block" in an inventory are very different entities. A block in the world is represented by an `IBlockState`, and its behavior defined by an instance of `Block`. Meanwhile, an item in an inventory is an `ItemStack`, controlled by an `Item`. As a bridge between the different worlds of `Block` and `Item`, there exists the class `ItemBlock`. `ItemBlock` is a subclass of `Item` that has a field `block` that holds a reference to the `Block` it represents. `ItemBlock` defines some of the behavior of a "block" as an item, like how a right click places the block. It's possible to have a `Block` without an `ItemBlock`. (E.g. `minecraft:water` exists a block, but not an item. It is therefore impossible to hold it in an inventory as one.)
 
-!!! note
-
-    When using a simple string, the currently active mod's ID will be added. So if I was doing this from "mymod", the real registry name would be "mymod:foo".
-
-It is also possible to pass a `ResourceLocation` directly into `register`, but this is just convenience for calling `setRegistryName` with the `ResourceLocation` beforehand.
+    When a block is registered, *only* a block is registered. The block does not automatically have an `ItemBlock`. To create a basic `ItemBlock` for a block one should use `new ItemBlock(block).setRegistryName(block.getRegistryName())`. The unlocalized name is the same as the block's. Custom subclasses of `ItemBlock` may be used as well. Once an `ItemBlock` has been registered for a block, `Item.getItemFromBlock` can be used to retrieve it.
 
 Further Reading
 ---------------
 
-For information about block properties, such as those used for vanilla blocks like wood types, fences, walls, and many more, see the section on [blockstates].
+For information about block properties, such as those used for vanilla blocks like wood types, fences, walls, and many more, see the section on [blockstates][].
 
 [sounds]: ../effects/sounds.md
+[registering]: ../concepts/registries.md#registering-things
 [blockstates]: ../blockstates/states.md
--- a/docs/concepts/registries.md
+++ b/docs/concepts/registries.md
@@ -1,0 +1,84 @@
+Registries
+==========
+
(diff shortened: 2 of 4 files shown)


docs/concepts/registries.md
```diff
@@ -0,0 +1,82 @@
+Registries
+==========
+
+Registration is the process of taking the objects of a mod (items, blocks, sounds, etc.) and making them known to the game. Registering things is important, as without registration the game will simply not know about these objects in a mod and will exhibit great amounts of unexplainable (and probably crashy) behavior.
```
kenzierocks: "crashy" sounds weird, don't know a good solution.

docs/concepts/registries.md
```diff
@@ -0,0 +1,82 @@
...
...IForgeRegistryEntry`. This allows the RL to act like a "registry name" for the object. The registry name for an object may be accessed with `get`/`setRegistryName`. The setter can only ever be called once, and calling it twice results in an exception. Every type of registrable object has its own registry, and names in two different registries will not collide. (E.g. there's a registry for `Block`s, and a registry for `Item`s, and a `Block` may be registered with the name `mod:example`, and an `Item` may be registered with the same name, without colliding. However, if two blocks were registered with that name, the game would crash.) The default implementation of `IForgeRegistryEntry` also provides two convenience implementations of `setRegistryName`: one where the parameter is a single string, and one where there are two string parameters. The overload that takes a single string checks whether the input contains a `:` (i.e. it checks whether the passed in stringified `ResourceLocation` has a domain), and if it doesn't it uses the current modid as the resource domain. The two argument overload simply constructs the registry name using the `modID` as the domain and `name` as the path.
```
kenzierocks: >A registry is a simple object similar to a map that assigns values to keys.

"A registry is a simple object**,** similar to a map**,** that assigns values to keys."
Also, the registry associates an integer ID as well that is used internally.

>Forge uses registries from `ResourceLocation`s to registry objects, which are of type `IForgeRegistryEntry`.

"registry objects" -> "register objects"

>E.g. there's a registry for `Block`s, and a registry for `Item`s, and a `Block` may be registered with the name `mod:example`, and an `Item` may be registered with the same name, without colliding.

e.g. there's a registry for blocks, and a registry for items. A block and item may both be registered with the name `mod:example`.

>uses the current modid as the resource domain.

either modID to be consistent with the parameter name or mod ID for "proper" english

docs/concepts/registries.md
```diff
@@ -0,0 +1,82 @@
...
+
+There also exists the "registry registry", a registry that registers other registries. This registry registers registries with `ResourceLocation` names; additionally, it registers registries to `Class` keys. This allows one to look up the registry that registers a certain class (e.g. one can look up `Block.class` in the registry registry with `GameRegistry.findRegistry` to get the registry that registers blocks).
```
kenzierocks: >`ResourceLocation` names; additionally, it registers registries to `Class` keys.

"`ResourceLocation` names and `Class` keys."

docs/concepts/registries.md
```diff
@@ -0,0 +1,82 @@
...
+@SubscribeEvent
+public void registerBlocks(RegistryEvent.Register<Block> event) {
+    e.getRegistry().registerAll(block1, block2, ...);
+}
+```
+
+The order in which `RegistryEvent.Register` events fire is arbitrary, with the exception that `Block` will *always* fire first, and `Item` will *always* fire second, right after `Block`.
+
+!!! important
+    These events are fired *before* preinit. This means that `@Mod.EventBusSubscriber` (or `MinecraftForge.EVENT_BUS.register` in the `@Mod` class's constructor for e.g. Scala mods which do not support `static`) should be used to register the event handler before preinit.
```
kenzierocks: Instead of the constructor, `FMLConstructionEvent` is probably safer.

docs/concepts/registries.md
```diff
@@ -0,0 +1,82 @@
...
+
+There is another, older way of registering objects into registries, using `GameRegistry.register`. Anytime something suggests using this method, it should be replaced with an event handler for the appropriate registry event. This method simply finds the registry corresponding to an `IForgeRegistryEntry` with `IFRE::getRegistryType`, and then registers the object to the registry. There is also a convenience overload that takes an IFRE and an RL, which is equivalent to `GameRegistry.register(ifre.setRegistryName(rl))`.
```
kenzierocks: It might not be clear that `IFRE::getRegistryType` is shorthand.

Also, I have no clue what you are referencing with "There is also a convenience overload that takes an IFRE and an RL, which is equivalent to `GameRegistry.register(ifre.setRegistryName(rl))`."

docs/concepts/registries.md
```diff
@@ -0,0 +1,82 @@
...
+
+It is possible to have Forge inject values from registries into `public static final` fields of classes. This is done by annotating classes and fields with `@ObjectHolder`. If a class has this annotation, all the `public static final` fields within are taken to be object holders too, and the value of the annotation is the domain of the holder (i.e. every field uses it as the default domain for the registry name of the object to inject). If a field has this annotation, and the value does not contain a domain, the domain is chosen from the surrounding class's `@ObjectHolder` annotation. If the class is not annotated in this situation, the field is ignored with a warning. If it does contain a domain, then the object to inject into the field is the object with that name. If the class has the annotation and one of the `public static final` fields does not, then the resource path of the object's name is taken to be the field's name. The type of the registry is taken from the type of the field.
+
+!!! note
+    If an object is not found, whether the object itself hasn't been registered or the registry does not exist, a debug message is logged and the field is left unchanged.
```
kenzierocks: >If an object is not found, whether the

If an object **is found and the**

docs/effects/sounds.md
```diff
@@ -1,78 +1,63 @@
...
------------
+`sounds.json`
+-------------
 
-This json should be located in your base asset directory:
-```
-src/main/resources/assets/[mod id]/sounds.json
-```
-This indicates to the vanilla resource system what Sound Events you declare and what Sound Files those events use.
+This json should be located in the `assets` directory of a mod: `src/main/resources/assets/<mod id>/sounds.json`. This file declares sound events and identifies which sound files they use.
```
kenzierocks: Personally, I think path would be better using `<modID>` so there's no space in it. Just an opinion though.

docs/effects/sounds.md
```diff
@@ -1,78 +1,63 @@
...
   }
 }
 ```
 
-Underneath the top-level object, each key we have defined corresponds to a Sound Event we want to tell the game about: `"mymod:openChest"` and `"mymod:epicMusic"`. Note that they are written in the sounds.json without the modid.
+!!! important
+    Like the rest of the resource system, everything_should_be_in_snake_case.
 
-Under each event we specify the category of the Sound Event, then a subtitle localization key to be shown to hard of hearing users when the Sound Event is played. 
+Underneath the top-level object, each key corresponds to a sound event. Note that they are written in the sounds.json without the modid. Each event specifies its category, and a localization key to be shown when subtitles are enabled. Finally, the actual sound files to be played are specified. Note that the value is an array; if multiple sound files are specified then the game will randomly choose one to play whenever the sound event is triggered.
```
kenzierocks: Same thing with `modid` here.

docs/concepts/registries.md
kenzierocks: Never mind, its javadoc states "Should not be used by mods."

docs/concepts/registries.md
howtonotwin: Throughout the docs "modid" is used, so I'll leave it like that. I searched for all usages of all forms of that word with `grep -E '[Mm]od\s?[Ii][Dd]' -r docs/` and found it the most common.

docs/effects/sounds.md
howtonotwin: Leaving alone as I said [above](#discussion_r86695350).

docs/concepts/registries.md
diesieben07: "... as un-registered objects can behave in weird and seemingly inexplicable ways or might even make the game crash when used."

docs/blocks/blocks.md
```diff
@@ -27,26 +27,19 @@ Of course, the above only allows for extremely basic blocks. If you want to add
...
-
-!!! note
-
-    When using a simple string, the currently active mod's ID will be added. So if I was doing this from "mymod", the real registry name would be "mymod:foo".
-
-It is also possible to pass a `ResourceLocation` directly into `register`, but this is just convenience for calling `setRegistryName` with the `ResourceLocation` beforehand.
+    There is a very large difference between a block in the world and a block in an inventory. A block in the world is governed by an instance of `Block`, it has a blockstate, a 4-bit meta, and maybe a tile entity. Meanwhile, an item in an inventory is controlled by an instance of `Item`, with a 16-bit damage/meta, and maybe an NBT tag. As a bridge between the different worlds of `Block` and `Item`, there exists the class `ItemBlock`. `ItemBlock` is a subclass of `Item` that has a field `block` that holds a reference to the `Block` it represents. `ItemBlock` defines some of the behavior of a "block" as an item, like how a right click places the block. It's possible to have a `Block` without an `ItemBlock`. (E.g. `minecraft:water` is a block, but not an item.)
```
diesieben07: Might as well clean this up if you are changing it anyways.

- "very large difference" sounds clunky, if my English does not betray me differences are big, not large, right? In any case, "very" does not fit here imho, "a big difference" is enough.
- "A block in the world is governed by an instance of `Block`" - No, it's not. If anything this should be `IBlockState`, but even that is not true. Moreover "blockstate" and "4-bit meta" are not parallel, but orthogonal concepts. The blockstate exists and is _serialized_ into the metadata.

docs/blocks/blocks.md
```diff
@@ -27,26 +27,19 @@ Of course, the above only allows for extremely basic blocks. If you want to add
...
+    
+    When a block is registered, *only* a block is registered. The block does not automatically have an `ItemBlock`. To create a basic `ItemBlock` for a block one should use `new ItemBlock(block).setRegistryName(block.getRegistryName())`. Custom subclasses of `ItemBlock` may be used as well. Once an `ItemBlock` has been registered for a block, `Item.getItemFromBlock` can be used to retrieve it.
```
diesieben07: Maybe add a sentence or two about the effects caused by not having an `ItemBlock`?

docs/concepts/registries.md
```diff
@@ -0,0 +1,84 @@
...
...ype `IForgeRegistryEntry`. This allows the RL to act like a "registry name" for the object. The registry name for an object may be accessed with `get`/`setRegistryName`. The setter can only ever be called once, and calling it twice results in an exception. Every type of registrable object has its own registry, and names in two different registries will not collide. (E.g. there's a registry for `Block`s, and a registry for `Item`s, and a `Block` and an `Item` may be registered with the same name `mod:example` without colliding. However, if two blocks were registered with that name, the game would crash.) The default implementation of `IForgeRegistryEntry` (`IForgeRegistryEntry.Impl`) also provides two convenience implementations of `setRegistryName`: one where the parameter is a single string, and one where there are two string parameters. The overload that takes a single string checks whether the input contains a `:` (i.e. it checks whether the passed in stringified `ResourceLocation` has a domain), and if it doesn't it uses the current modid as the resource domain. The two argument overload simply constructs the registry name using the `modID` as the domain and `name` as the path.
```
diesieben07: I do not agree with the word "crash" being used here: "However, if two blocks were registered with that name, the game would crash.". I would prefer something like "an exception would be thrown" or something.

docs/blocks/blocks.md
howtonotwin: Something along the lines of "represented as an `IBlockState`, which is controlled by a `Block`" with something similar for `Item` with `ItemStack`?

diesieben07: "represented as an IBlockState, which is a Block instance and it's properties" is what I would say.

tterrag1098:
@diesieben07 do you approve these changes?

diesieben07:
Sure .

PR #75, opened 2016-11-28
Add documentation on configuration files

rohannb:


docs/utilities/config.md
```diff
@@ -0,0 +1,35 @@
...
+Configuration files are accessed through an object of type `net.minecraftforge.common.config.Configuration`. Its constructor takes a `java.io.File` object as input, which will be used to save the configurations. The recommended file to use is `FMLPreInitializationEvent.getSuggestedConfigurationFile()`, which points to the file `.minecraft/config/[modid].cfg`
+
+```java
+public static Configuration config;
+
+@EventHandler
+public void preInit(FMLPreInitializationEvent event) {
+  config = new Configuration(event.getSuggestedConfigurationFile());
+}
+```
```
diesieben07: I don't think this code example is necessary.

docs/utilities/config.md
```diff
@@ -0,0 +1,35 @@
...
+See the javadocs on `Configuration` for more info.
+
+Config GUIs
+-----------
+
+!!! important
+
+    You must have your mcmod.info file set up to use config GUIs. See [Structuring Your Mod: The `mcmod.info` file](../gettingstarted/structuring.md#the-mcmodinfo-file) to learn how.
+
+In-game, within the "Mods" dialog, there is a button labeled "Config" that is disabled by default. If you would like to know how to enable and make use of the "Config" dialog, @Minalien has [a great guide on setting it up](https://github.com/Minalien/BlogArchive/blob/master/ForgeTutorials/Spotlight__Config_GUIs.md).
```
diesieben07: This is kind of a cop-out, to be honest. This is the part about config files that would actually require documentation the most and you are just linking to some tutorial.
In my opinion this makes this article not very useful.

TehNut:
If we're going to document the config system, we should document the new method using the annotations instead.

docs/utilities/config.md
tterrag1098: External source materials should be used sparingly and only where absolutely necessary, such as external code examples. This is just lazy and potentially harmful if the source becomes outdated or offline.

PR #76, opened 2016-11-30
Add Information about Vanilla Blockstates

wincinderith:
I ran into this issue because there was no documentation on it. Hopefully this will save someone else the pain I endured trying to figure this out. :)

### DIFF
--- a/docs/blockstates/introduction.md
+++ b/docs/blockstates/introduction.md
@@ -26,6 +26,8 @@ The log only has one property: axis. A blockstate always has to be defined for a
 "east=false,north=false,south=false,west=false": { "model": "oak_fence_post" }
 ```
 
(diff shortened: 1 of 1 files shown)


bonii-xx:
Property names are resources, all resources need to be lowercase.

tterrag1098:
@bonii-xx I still think this is valid to point out. Personally I don't think of properties as "resources". They are *used* in resources, but aren't resources themselves.

bonii-xx:
They're part of the model resource. I'm not saying it's bad to point it out, But it probably should be pointed out that all resources need to be lowercase.

wincinderith:
@bonii-xx @tterrag1098 How's that?

bonii-xx:
Maybe also add a "this includes the blockstate variant itself, e.g. foobar=baz" the example being a simple enttry from the page.

wincinderith:
@bonii-xx Updated.

PR #79, opened 2016-12-13
Translate the Documentation to zh_TW (Taiwan)

gkbm2013:
Add a zh_TW (a.k.a. Mandarin or Chinese Traditional) translation.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -1,11 +1,14 @@
-MinecraftForge Documentation
+MinecraftForge 說明文件
 ============================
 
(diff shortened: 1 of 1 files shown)


williewillus:
Does RTD support multiple languages? Also you just translated the home page, not any of the actual stuff :p

LexManos:
No, also branding with your site/name is not allowed.

elifoster:
[Actually RTD does support multiple languages](https://docs.readthedocs.io/en/latest/localization.html).

tterrag1098:
> This feature only applies to Sphinx documentation.

PR #81, opened 2017-01-02
Add warning about actual states and TE's

williewillus:
This can cause things like Vazkii/Botania#1999 to happen if you don't do this check

Vanilla does this check in the flower pot itself


diesieben07:
Oh my this is so ugly... But what can you do.

williewillus:
I'm just puzzled why they thought having getTileEntity mutate the world if it can't find anything be a good idea

tterrag1098:
A getter mutating the world is a fundamental mistake, yes. It would be worthwhile to investigate a forge PR for this issue, but for now this note is probably a good idea.

diesieben07:
What puzzles me is that your workaround even works reliably. A HashMap should explode terribly with multiple threads, even if only one of them writes.

tterrag1098:
@diesieben07 a ChunkCache is only ever accessed from the thread it is created in, though.

diesieben07:
Ah right. I get it now.

PR #84, opened 2017-01-03
Update registries.md

elifoster:
- "that assigns values to keys" -> "that assigns keys to values" because that is more accurate to what a map is
- Explicitly define the abbreviation for ResourceLocation with (RL), since RL is used later on in the docs.
- `register`(`All`) -> `register` (or `registerAll`), when I first read this I didn't really understand what the All was telling me, and had to go look in the source code to figure it out. This should be more clear.


tterrag1098:
>"that assigns values to keys" -> "that assigns keys to values" because that is more accurate to what a map is
I'm not sure I'd agree

elifoster:
Honestly that bit is kind of pointless either way, because surely we all know what a map is. The only reason I didn't remove it was because later sentences were sort of dependent on it, and I wasn't sure how I'd accurately change them.

tterrag1098:
Honestly, I think the first 2 changes are unnecessary (the second change being superceded by #86). I would still appreciate the third change as a quick patch.

elifoster:
@tterrag1098 done, see #88.

PR #86, opened 2017-01-16
Add "Resources" page to "Concepts"

howtonotwin:
Describe how Minecraft deals with resources. Explain what `ResourceLocation`s are.

Edit other pages to remove duplication of this information. Instead, have them refer to the new page.


docs/concepts/resources.md
```diff
@@ -0,0 +1,15 @@
+Resources
+=========
+
+A resource is extra data used by the game, and is stored in a data file, instead of being in the code. Resources are contained in the `assets` directory on the classpath. In the default MDK, this directory is under the `src/main/resources` directory of the project. They include things such as models, textures, and localization files.
```
elifoster: Expand the `MDK` abbreviation

docs/concepts/resources.md
```diff
@@ -0,0 +1,15 @@
...
+
+When multiple resource packs are enabled, they are merged. Generally, files from resource packs at the top of the stack override those below; however, for certain files, such as localization files, they are actually merged contentwise. Mods actually define resource packs too, and their resources are simply merged with the default resource pack inside the core Minecraft JAR in the same way as any other.
+
+All resources should have snake_case paths; this is enforced. This casing convention is used even outside the resource system.
```
elifoster: This isn't true... I have tons of assets that use camelCase.

docs/concepts/resources.md
```diff
@@ -0,0 +1,15 @@
...
+
+`ResourceLocation`
```
elifoster: Perhaps describe ModelResourceLocations a bit as well.

docs/concepts/resources.md
shadowfacts: In 1.11 asset names must be completely lower case. This basically enforces snake_case

elifoster: Well then this should be changed to:

>All resources should be entirely lower-cased; this is enforced after 1.11. Using snake_case is recommended but not required.

williewillus: yeah, saying "is used even outside the resource system" implies to me people should be coding Java in snake case :P

docs/concepts/resources.md
```diff
@@ -0,0 +1,15 @@
...
+------------------
+
+Minecraft identifies resources using `ResourceLocation`s (RLs). An RL contains two parts: a domain and a path. It generally points to the resource at `assets/<domain>/<ctx>/<path>`, where `ctx` is a context-specific path fragment that depends on how the RL is being used. When an RL is written/read as/from a string, it is seen as `<domain>:<path>`. If the domain and the colon are left out, then when the string is read into an RL the domain will normally default to `"minecraft"`. A mod should put its resources into a domain with the same name as its modid (E.g. a mod with id `examplemod` should place its resources in `assets/examplemod`, and RLs pointing to those files would look like `examplemod:<path>`.). RLs are used outside the resource system, too, as they happen to be a great way to uniquely identify objects.
```
williewillus: I'm not sure if the part about `ctx` is needed nor accurate

williewillus: additionally, expand on their non-asset uses (registries) beyond 1 sentence :P

docs/concepts/resources.md
elifoster: That too.

docs/concepts/resources.md
howtonotwin: That doesn't belong here. It's more related to models than "general resource stuff."

docs/concepts/resources.md
howtonotwin: Examples of extra context:

* Sounds: `sounds`
* Blockstates: `blockstates`
* Blockstate model refs: `models/block`
* Models: `models`
* Textures: `textures`

etc.

docs/concepts/resources.md
```diff
@@ -0,0 +1,17 @@
+Resources
+=========
+
+A resource is extra data used by the game, and is stored in a data file, instead of being in the code. Resources are contained in the `assets` directory on the classpath. In the default mod development kit, this directory is under the `src/main/resources` directory of the project. They include things such as models, textures, and localization files.
+
+When multiple resource packs are enabled, they are merged. Generally, files from resource packs at the top of the stack override those below; however, for certain files, such as localization files, they are actually merged contentwise. Mods actually define resource packs too, and their resources are simply merged with the default resource pack inside the core Minecraft JAR in the same way as any other.
+
+All resources should have snake_case paths; this is enforced in 1.11 and above. This casing convention is used even outside the resource system, generally whenever a `String` identifier is needed (e.g. modids). (This does *not* mean Java identifiers should be snake cased.)
```
elifoster: Should domains be snake_case too? Clarify

>(This does *not* mean Java identifiers should be snake cased.)

Emphasis on "not" makes it sound like you shouldn't do that, when in fact there's no reason not to– it's just a style thing.

TehNut: I agree. Emphasis should probably be put on "Java identifiers" if any at all.

docs/concepts/resources.md
```diff
@@ -0,0 +1,17 @@
+Resources
+=========
+
+A resource is extra data used by the game, and is stored in a data file, instead of being in the code. Resources are contained in the `assets` directory on the classpath. In the default mod development kit, this directory is under the `src/main/resources` directory of the project. They include things such as models, textures, and localization files.
+
+When multiple resource packs are enabled, they are merged. Generally, files from resource packs at the top of the stack override those below; however, for certain files, such as localization files, they are actually merged contentwise. Mods actually define resource packs too, and their resources are simply merged with the default resource pack inside the core Minecraft JAR in the same way as any other.
+
+All resources should have snake_case paths and domains; furthermore, lowercasing is enforced in 1.11 and above. This casing convention is used even outside the resource system, generally whenever a `String`/`ResourceLocation` identifier is needed (e.g. modids). (This does not mean Java identifiers should be snake cased.)
```
tterrag1098: This sentence is clunky. Perhaps "All mod resources should have lowercase paths and filenames, and this standard is enforced in MC 1.11 and above."

docs/concepts/resources.md
```diff
@@ -0,0 +1,17 @@
...
+
+`ResourceLocation`
+------------------
+
+Minecraft identifies resources using `ResourceLocation`s (RLs). An RL contains two parts: a domain and a path. It generally points to the resource at `assets/<domain>/<ctx>/<path>`, where `ctx` is a context-specific path fragment that depends on how the RL is being used. When an RL is written/read as/from a string, it is seen as `<domain>:<path>`. If the domain and the colon are left out, then when the string is read into an RL the domain will normally default to `"minecraft"`. A mod should put its resources into a domain with the same name as its modid (E.g. a mod with id `examplemod` should place its resources in `assets/examplemod`, and RLs pointing to those files would look like `examplemod:<path>`.). RLs are used outside the resource system, too, as they happen to be a great way to uniquely identify objects (e.g. [registries][]).
```
tterrag1098: "will normally default to" remove the word 'normally' here, as it will *always* default to "minecraft"

tterrag1098: "A mod should put its resources into a domain with the same name as its modid" While this is a decent standard, it is not required or enforced, and in some cases, it may be desirable to have an asset domain that is different from the modid (or multiple asset domains).

docs/concepts/resources.md
tterrag1098: The second sentence seems almost unnecessary. I'm not even sure what it's trying to say, really. Perhaps this bit of info could be merged somewhere else?

docs/concepts/resources.md
```diff
@@ -0,0 +1,17 @@
...
+
+Wherever a `ResourceLocation` is used, it should almost always be in snake case.
```
tterrag1098: This seems already covered, but if it is left in, the word "almost" can be removed. It is a standard now that RLs should be all lowercase.

docs/concepts/resources.md
elifoster: Yeah I don't understand what the second sentence is telling me either.

docs/concepts/resources.md
elifoster: Yes this is already covered. If this sentence stays, it should clarify that this is only a standard starting in 1.11. But honestly this should just be removed because it's redundant and less informative.

docs/concepts/resources.md
howtonotwin: In at least one instance it doesn't. `IForgeRegistryEntry::setRegistryName(String)` defaults the domain to the modid if absent.

docs/concepts/resources.md
```diff
@@ -0,0 +1,15 @@
...
+
+All resources should have snake case paths and filenames (lowercase using "_" for word boundaries), which is enforced in 1.11 and above.
+
+`ResourceLocation`
+------------------
+
+Minecraft identifies resources using `ResourceLocation`s (RLs). An RL contains two parts: a domain and a path. It generally points to the resource at `assets/<domain>/<ctx>/<path>`, where `ctx` is a context-specific path fragment that depends on how the RL is being used. When an RL is written/read as/from a string, it is seen as `<domain>:<path>`. If the domain and the colon are left out, then when the string is read into an RL the domain will almost always default to `"minecraft"`. A mod should put its resources into a domain with the same name as its modid (E.g. a mod with id `examplemod` should place its resources in `assets/examplemod`, and RLs pointing to those files would look like `examplemod:<path>`.). This is not a requirement, and in some cases it can be desirable to use a different (or even more than one) domain. RLs are used outside the resource system, too, as they happen to be a great way to uniquely identify objects (e.g. [registries][]).
```
tterrag1098: The abbreviation to "RL" is unnecessary here

docs/concepts/resources.md
```diff
@@ -0,0 +1,15 @@
+Resources
+=========
+
+A resource is extra data used by the game, and is stored in a data file, instead of being in the code. Resources are contained in the `assets` directory on the classpath. In the default mod development kit, this directory is under the `src/main/resources` directory of the project. It includes things such as models, textures, and localization files.
+
+When multiple resource packs are enabled, they are merged. Generally, files from resource packs at the top of the stack override those below; however, for certain files, such as localization files, data is actually merged contentwise. Mods actually define resource packs too, in their `resources` directories, but they are seen as subsets of the "Default" resource pack. Mod resource packs cannot be disabled, but they can be overriden by other resource packs.
+
+All resources should have snake case paths and filenames (lowercase using "_" for word boundaries), which is enforced in 1.11 and above.
```
tterrag1098: "lowercase using" -> "lowercase, using"

docs/concepts/resources.md
howtonotwin: In every context except for actual code, the abbreviation is used. I don't see why this is any different, and it may be even more important *to* use it because this page introduces the abbreviation. Besides, all abbreviations are unnecessary; we use them to represent something much longer while conveying the same intent, which is especially nice when the term is used this many times in a single paragraph.

tterrag1098: Where else is that abbreviation used?

howtonotwin: I didn't mean just these docs; I meant *out there*, in whatever text the reader happens to come across, or even on irc, where "rl" is often thrown about without prior clarification. In addition, my models pr uses it.

tterrag1098: See #92. I am also in favor of dropping unnecessary abbreviations. Saving a few characters isn't worth potential confusion.

elifoster: How is it confusing if it is explicitly stated that that is the abbreviation for it?

williewillus: personally, RL is just extremely lazy irc shorthand (which actually came from "MRL" ModelResourceLocation first). I don't think it needs to be in the doc really

howtonotwin: Would it be possible to hold a vote? Otherwise, I think I'll expand the abbreviations and note that many other texts will use RL.

LexManos: Official documentation shouldn't use abbreviations. Honestly it doesn't even need to be mentioned that its a shorthand. IRC/other chats use a lot of chort hand that doesn't need to be documented.

tterrag1098:
All good. Just need to fix that merge conflict.

tterrag1098:
Ok I guess that works.

PR #89, opened 2017-02-05
Added information about ItemStacks to capabilities

Denyol:
The Capabilities doc now has more information about adding custom inventories to a ItemStack, this was added because I was having trouble understanding how to add inventories to ItemStacks and makes the specific documentation more understandable.

### DIFF
--- a/docs/datastorage/capabilities.md
+++ b/docs/datastorage/capabilities.md
@@ -69,7 +69,7 @@ public <T> T getCapability(Capability<T> capability, EnumFacing facing) {
 
 It is strongly suggested that direct checks in code are used to test for capabilities instead of attempting to rely on maps other other data structures, since capability tests can be done by many objects every tick, and they need to be as fast as possible in order to avoid slowing down the game.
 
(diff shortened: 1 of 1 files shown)


docs/datastorage/capabilities.md
```diff
@@ -69,6 +69,8 @@ public <T> T getCapability(Capability<T> capability, EnumFacing facing) {
 
 It is strongly suggested that direct checks in code are used to test for capabilities instead of attempting to rely on maps other other data structures, since capability tests can be done by many objects every tick, and they need to be as fast as possible in order to avoid slowing down the game.
 
+In the case of exposing a capability to an custom ItemStack of your mod, it is required that the `Item` class overrides `initCapabilities` to return a custom class implementing `ICapabilityProvider`. The custom `ICapabilityProvider`needs to override the above methods for exposing capabilities.
```
elifoster: `an custom` -> `a custom`

Missing space between `ICapabilityProvider` and `needs`

elifoster (changes requested):

docs/datastorage/capabilities.md
Denyol: Added, thank you for the input.

tterrag1098:
I think this would be better inside an admonition, as it is more of a warning/footnote and just having it as a simple paragraph makes it easy to miss.

PR #90, opened 2017-02-27
Add classpath fix note for IDEA 2016 to terminal steps, too

TeNNoX:
This note is only in the steps list for "Terminal-free IntelliJ IDEA configuration":

> 8. Finally, assuming you use IDEA 2016 or newer, you will have to fix the classpath module. Go to `Edit configurations` and in both `Minecraft Client` and `Minecraft Server`, change `Use classpath of module` to point to the task with a name like `<project>_main`.

It should be added to  the first list "From Zero to Modding", too.

In addition to that, I fixed the "Terminal-free configuration" list's indexing.

### DIFF
--- a/docs/gettingstarted/index.md
+++ b/docs/gettingstarted/index.md
@@ -19,6 +19,7 @@ From Zero to Modding
 6. Load your project into your IDE.
     * For Eclipse, create a workspace anywhere (though the easiest location is one level above your project folder). Then simply import your project folder as a project, everything will be done automatically.
     * For IntelliJ, you only need to create run configs. You can run `gradlew genIntellijRuns` to do this.
+    * **Note:** If you are using IDEA 2016 or newer, you will have to fix the classpath module. Go to `Edit configurations` and in both `Minecraft Client` and `Minecraft Server`, change `Use classpath of module` to point to the task with a name like `<project>_main`.
 
 !!! note
 
--- a/docs/gettingstarted/index.md
+++ b/docs/gettingstarted/index.md
@@ -39,13 +39,15 @@ From Zero to Modding
 Terminal-free IntelliJ IDEA configuration
 ------------------------------------------
 
-These instructions assume that you have created the project folder as described in the steps 1 to 3 of the section above. Because of that, the numbering starts at 4.
+These instructions assume that you have created the project folder as described in the **steps 1 to 3 of the section above**. Because of that, the numbering starts at 4.
 
-4. Launch IDEA and choose to open/import the `build.gradle` file, using the default gradle wrapper choice. While you wait for this process to finish, you can open the gradle panel, which will get filled with the gradle tasks once importing is completed.
-5. Run the `setupDecompWorkspace` task (inside the `forgegradle` task group). It will take a few minutes, and use quite a bit of RAM. If it fails, you can add `-Xmx3G` to the `Gradle VM options` in IDEA's gradle settings window, or edit your global gradle properties.
-6. Once the setup task is done, you will want to run the `genIntellijRuns` task, which will configure the project's run/debug targets. 
-7. After it's done, you should click the blue refresh icon **on the gradle panel** (there's another refresh icon on the main toolbar, but that's not it). This will re-synchronize the IDEA project with the Gradle data, making sure that all the dependencies and settings are up to date.
-8. Finally, assuming you use IDEA 2016 or newer, you will have to fix the classpath module. Go to `Edit configurations` and in both `Minecraft Client` and `Minecraft Server`, change `Use classpath of module` to point to the task with a name like `<project>_main`.
+<ol start="4">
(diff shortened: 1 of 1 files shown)


shadowfacts:
The better solution to this issue is to, when the project is being imported into IDEA, make sure that it's set to **not** create a separate module for each source set.

TeNNoX:
This PR is more of a quickfix for the people out there running into this problem. (like me, earlier ;)
An actual fix would definitely be even better.

elifoster:
@shadowfacts I disagree. It's incredibly useful for when you are creating an API for your mod (and you don't want to wrongly force the API to be dependent on the mod, which is a problem I see quite frequently in the modding scene) and when you are using automated testing.

docs/gettingstarted/index.md
```diff
@@ -38,13 +39,15 @@ From Zero to Modding
...
-5. Run the `setupDecompWorkspace` task (inside the `forgegradle` task group). It will take a few minutes, and use quite a bit of RAM. If it fails, you can add `-Xmx3G` to the `Gradle VM options` in IDEA's gradle settings window, or edit your global gradle properties.
-6. Once the setup task is done, you will want to run the `genIntellijRuns` task, which will configure the project's run/debug targets. 
-7. After it's done, you should click the blue refresh icon **on the gradle panel** (there's another refresh icon on the main toolbar, but that's not it). This will re-synchronize the IDEA project with the Gradle data, making sure that all the dependencies and settings are up to date.
-8. Finally, assuming you use IDEA 2016 or newer, you will have to fix the classpath module. Go to `Edit configurations` and in both `Minecraft Client` and `Minecraft Server`, change `Use classpath of module` to point to the task with a name like `<project>_main`.
+<ol start="4">
```
elifoster: Why is HTML being used here instead of markdown? Seems weird.

tterrag1098: It is weird. Should be fixed.

shadowfacts:
@elifoster per cpw's video (https://twitter.com/voxcpw/status/836027952374611968), that's the recommended way of doing things.

elifoster:
@shadowfacts Did cpw provide a reason for that? I've run into WAY too many mod APIs that are dependent on the mod itself. [Here](https://github.com/Esteemed-Innovation/Esteemed-Innovation/tree/1.10/src) is how it looks for Esteemed Innovation, and it is working flawlessly. The API compiles independently of the mod (good for distributing the API for potential addon developers) by force— referencing a mod class will just not work— and the mod does not compile without the API.

docs/gettingstarted/index.md
shadowfacts: It's because otherwise the list starts at one, even though it says "the numbering starts at 4."

mezz:
I think the reason cpw avoided separate modules per source set is because it makes the run configs work easier and because of this bug in gradle 3.4:
https://github.com/gradle/gradle/issues/1276#issuecomment-282495000

matthewprenger:
Getting MinecraftForge/ForgeGradle#351 pulled would eliminate the need for this PR. 

As for the separate or not source sets debate, I agree with elifoster. Disabling separate source sets is an option, but you lose functionality. IMO a better option is to just hold gradle back to 3.3 until IntelliJ 2017 officially drops.

cpw:
@mezz @matthewprenger @elifoster I avoided it because it's a simpler setup. Not many modders will be building complex APIs (I hope) that require separate modules - those that do, probably already understand how to do it. If you're starting out, a single module "looks" like a sensible setup.

The whole 3.4 "fuxxoring" IntelliJ is just a shitshow. Someone from gradle? dev popped up on my twitter saying they didn't intend to break IntelliJ, but it happened anyway. Yay for no communication!

Also, the EAP is barely useable at the minute - I got at least one error in my video from the other day. I cannot edit the gradle config - it just errors constantly when trying to do so. Same for the run configs.

I have recommended staying on 3.3 until 3.5+ gradle fix or concrete 2017 IntelliJ release occurs.

elifoster:
@cpw Unfortunately those that are building complex APIs, even if they are otherwise solid developers, often still make the mistake of having the API dependent on the mod. I see developers wanting to do integration/addons complaining about it all the time on Twitter. It was present in a mod I maintain, but did not originally write, for over 2 years.

mezz:
@elifoster I think this is about the default choice though, where the simpler approach is better. There should definitely be some extra information in the docs for people who want to create a separate API jar so that both approaches are supported. Once IntelliJ becomes stable with gradle again I will probably switch a couple of my projects to work with proper separate modules, I didn't know it was possible before.

docs/gettingstarted/index.md
mezz: I think you can escape the auto-numbering in markdown by writing it like `4\.`

Jerzerak:
Worth noting that IntelliJ IDEA 2017.1 has been [released](https://blog.jetbrains.com/idea/2017/03/intellij-idea-2017-1-java-9-kotlin-1-1-spring-gradle-javascript-go-and-more/) with support for composite builds and automatic project configuration for Gradle

tterrag1098:
So then this PR is unnecessary?

Jerzerak:
@tterrag1098 Correct

PR #92, opened 2017-03-17
Minor formatting changes

MidnightLightning:
Remove abbreviation for clarity


elifoster:
I agree with the back ticks, but I think RL is used enough within the community that it should be used (or at least mentioned) here. I think putting "(RL)" after the first instance of "ResourceLocation" (on each page where RL is used) would be the best way to deal with it. This is how we do it on the FTB Gamepedia, like "Redstone Flux (RF)".

MidnightLightning:
I think abbreviations are fine in general wikis and chats about the game, but the reason I suggested the change here is this is a technical manual, designed for developers. Examples and phrases here are likely to be copied and pasted or searched for directly, which would be a lot easier if they weren't abbreviated. Yes, it only takes a little bit of learning to remember some of the common abbreviations for things, but if you have the space and time to spell it out, why not?

The rest of this guide doesn't really abbreviate, which is why this instance stood out to me as I was reading through it as an experienced programmer in general, but new to Forge development.

mezz:
For what it's worth, I've never seen ResourceLocation abbreviated and wouldn't know what RL is.

williewillus:
^ the only abbreviation I've seen is MRL

Issue #85, opened 2017-01-03
Registries: why?


elifoster:
There are two things to address here.
1. It is recommended to use the RegistryEvents to register your stuff, but you can just as easily and correctly do it using the GameRegistry in preInit (which is the way we've done it forever, too). It is not explained *why* it is recommended to use RegistryEvents over the other way. This needs to be explained.
2. The value of making a registry with the RegistryBuilder is not explained. I can just as easily make a static class in my API that functions as a registry. Why should we use the RegistryBuilder over a private static map or list gated behind some helper functions? This needs to be explained in the docs.

cosmicdan:
I suppose "why" belongs in the paragraph about using the old method. I can think of a few reasons why the new system is preferred:

1) Faster startup
2) Easier cross-mod compatibility and interaction
3) Unfathomable future possibilities thanks to a more managed API

In short, the reason for these events is so the things are registered at what Forge considers as the best time to register. That potential for optimization should be reason enough.

LexManos:
1) The main difference is reinitalization. We are slowly {fighting modders every step of the way} trying to move Minecraft into a more data/instanced system. If people use the registry events it means we can nuke the registry, and re-run the register event and end up with a valid registry. We can't do that if modders are adding/removing crap all over the place. Why is this important? Connecting to servers, or loading multiple single player worlds. We hack around things using delegates and magic reflection replacements. But it would be nice to get away from that.

2) RegistryBuilder is just a normal builder pattern. Thats a fairly common java object style. As for why use a Forge registry over a Map<K,V>? If you have soft dependencies on things having the events fire the creation is ideal. However, if you don't care about forcing a hard dep , and you don't care about syncing the ids over the network/disc {or if you dont use integer IDs anywhere at all} then you can just use normal Maps.
But Forge registries have a lot of benefits:
k/v <-> int id
S->C sync/compressed id usage.
Override tracking/management 
Few other things, but there are benefits, and costs.

cosmicdan:
Great explanation Lex, makes much more sense than what I said.

So the original issue was the documentation lacking an explanation on why the new system is preferred over the old system. Is it even necessary to justify such a change within the _documentation_?

elifoster:
@cosmicdan As long as both are present I believe so, personally.

cosmicdan:
> As long as both are present I believe so, personally.

[...]

PR #93, opened 2017-03-17
Basic information on TESRs

elifoster:
This is by no means exhaustive, but it provides a brief introduction to TESRs and explains a few important things (like them using single-instance design).

Adds a new section, "Tile Entities", which should hopefully be expanded.


docs/tileentities/tesr.md
```diff
@@ -0,0 +1,20 @@
...
+Creating a TESR
+---------------
+
+To create a TESR, create a class that inherits from `TileEntitySpecialRenderer`. It takes a generic argument, of which the object must inherit from `TileEntity`. The generic argument is used in the TESR's method, `renderTileEntityAt`.
+
+Only one TESR exists for a given tile entity. Therefore, values that are specific to a single instance in the world should be stored in the tile entity being passed to the renderer rather than in the TESR itself. For example, an integer that increments every frame, if stored in the TESR, will increment every frame for every tile entity of this type in the world.
+
+### `renderTileEntityAt`
+
+This method is called every frame in order to render the tile entity. It passes the instance of the tile entity being renderered, its coordinates as doubles at the origin of the block, partialTicks<!-- I'm not actually sure what this is used for. -->, and the destroy stage of the block if being destroyed.
```
shadowfacts: `partialTicks` is the amount of time, measured in fractions of a tick, that have passed since the last full tick. It's used for interpolating animated/moving things between ticks, so they don't look really jittery.

diesieben07:
I think you should at least mention FastTESR.

elifoster:
I've never used or looked at any implementations of FastTESR.

docs/tileentities/tesr.md
```diff
@@ -0,0 +1,26 @@
...
+
+Only one TESR exists for a given tile entity. Therefore, values that are specific to a single instance in the world should be stored in the tile entity being passed to the renderer rather than in the TESR itself. For example, an integer that increments every frame, if stored in the TESR, will increment every frame for every tile entity of this type in the world.
+
+### `renderTileEntityAt`
+
+This method is called every frame in order to render the tile entity. 
+
+#### Parameters
+* `tileentity`: This is the instanceof the tile entity being rendered. It uses the generic argument, so, if the class is being inherited correctly, it will not need to be cast to your tile entity class.
+* `x`, `y`, `z`: The coordinates of the tile entity as doubles. They are at the origin of the block.
```
williewillus: I don't think this is true, TESR's aren't passed world coords.

In `TileEntityRendererDispatcher.renderTileEntity(TileEntity, float, int)`: 

`this.renderTileEntityAt(tileentityIn, (double)blockpos.getX() - staticPlayerX, (double)blockpos.getY() - staticPlayerY, (double)blockpos.getZ() - staticPlayerZ, partialTicks, destroyStage);`

elifoster: You're right. You can get the world coords from the te which is probably what I confused myself with when reading my TESRs.

How should I word that then? The position of the tile entity relative to the player's view?

diesieben07: "The coordinates at which the TESR should render it's model" or something like that.

tterrag1098:
FastTESR should be listed as the preferred way to do a TESR nowadays. *Only* if you specifically need operations which cannot be done with purely the VertexBuffer, should you use a normal TESR.

elifoster:
Can either of you provide any information on it? I've tried finding some and have been unsuccessful. The most informative thing I've found is the [class itself](https://github.com/MinecraftForge/MinecraftForge/blob/c545b8ecd4e96107ccf134c8fd89906357d586ad/src/main/java/net/minecraftforge/client/model/animation/FastTESR.java) but it doesn't have any documentation so it isn't very useful.

I will list it as the preferred way to do TESRs if someone gives me the actual use case for it vs. the "normal" one and *why* it is the preferred way to do a TESR.  Saying something is preferred without justification doesn't provide enough information/reason to use it over the thing people are already used to, as I described on a different topic at #85, and I'd prefer my PRs avoid doing that 😃

diesieben07:
I'll write up something this afternoon.

tterrag1098:
FastTESR's *only* advantage is that it s batched with all other FastTESR's. The stipulation of this is that the rendering code must ONLY interact with the VertexBuffer and nothing else (no GL operations). It's really that simple.

elifoster:
Thank you. Diesieben's PR to my fork explained it as well. That is useful to know and is, as far as I know, not possible to find without actually digging into the rendering code.

elifoster:
Any suggestions or anything needed to get this merged?

PR #100, opened 2017-05-10
[WIP] Animation API Docs

mincrmatt12:
This is _NOT_ finished yet, but I thought I would create the PR now so I can get some feedback on this.

### DIFF
--- a/docs/animation/armature.md
+++ b/docs/animation/armature.md
@@ -1,0 +1,147 @@
+Armature Files
+==============
+
+Armature Files define joints and clips for animating a model.
+
+File Structure
+----------------
+
+An example armature file, taken from the forge debug mod (requires #3875)
+```json
+{
+  "joints": {
+    "stick": {"2": [1.0]},
+    "cube": {"3": [1.0]}
+  },
+  "clips": {
+    "default": {
+      "loop": true,
+      "joint_clips": {
+        "stick": [
+          {
+            "variable": "offset_x",
+            "type": "uniform",
+            "interpolation": "linear",
+            "samples": [0, 0.6875, 0]
+          }
+        ],
+        "cube": [
+          {
+            "variable": "offset_x",
+            "type": "uniform",
+            "interpolation": "linear",
+            "samples": [0, 0.6875, 0]
+          },
+          {
+            "variable": "axis_z",
+            "type": "uniform",
+            "interpolation": "nearest",
+            "samples": [ 1 ]
+          },
+          {
+            "variable": "origin_x",
+            "type": "uniform",
+            "interpolation": "nearest",
+            "samples": [ 0.15625 ]
+          },
+          {
+            "variable": "origin_y",
+            "type": "uniform",
+            "interpolation": "nearest",
+            "samples": [ 0.40625 ]
+          },
+          {
+            "variable": "angle",
+            "type": "uniform",
+            "interpolation": "linear",
+            "samples": [0, 120, 240, 0, 120, 240]
+          }
+        ]
+      },
+      "events": {}
+    }
+  }
+}
+
+```
+
+The file is organized in two sections, joints and clips.
+
+Joints
+--------
+Each joint defines a connection to the model file
+
+The format is like this:
+```json
+ "joints": {
+  "joint_name": {"index in model file": [1.0]}
+ }
+```
+
+- `joint_name` is the name of the joint
+- `index in model file` is a 0-indexed number denoting which model element this joint controls
+
+Not all elements need to have a joint, only the ones you are animating.
+
+
+Clips
+-------
+
+Clips are essentially instructions on how to use a value to animate some collection of joints.
+They also include events to fire at certain points.
+
+They are formatted like this:
+```json
+"clips": {
+  "clip_name": {
+    "loop": true/false,
+    "joint_clips": {
+      "joint_name": [
+       {
+        "variable": "some_variable",
+        "type": "uniform",
+        "interpolation": "interpolation_type",
+        "samples": ["animation", "keypoints"]
+       }
+      ]
+    },
+    "events": {
+      "event_time 0-1": "event_name"    
+    }
+  }
+}
+```
+
+- loop: whether or not the parameter value can continue increasing and the animation loops, or if it stops at the ending state
+
+### Joint Clips
+Each `joint_clip` is a set of variables to change for a joint. The `type` attribute is currently ignored, but must be `"uniform"`.
+
+`interpolation` can be one of the following:
+  - nearest - if value < 0.5 use the first sample, else the second sample. Useful for static variables if only given one value
+  - linear - linearly interpolate between samples
+
+`variable` can be one of the following:
+
+- offset_x, offset_y, offset_z - translation
+- scale - uniform scaling
+- scale_x, scale_y, scale_x - scaling on certain axes
+- axis_x, axis_y, axis_z - rotation axes
+- angle - rotation angle
+
+and if #3875 is merged, there will be:
+
+- origin_x, origin_y, origin_z - rotation origin
+
+
+### Events
+
+Each clip can fire events, formatted like this:
+```json
+"events": {
+  "event_time": "event_text"
+}
+```
+For more information about events and what `event_text` means, see the page on ASMs.
+
+`event_time` is a value denoting when to fire the event. When the parameter controlling this clip reaches a point equal to or greater than the `event_time`, the event is fired
--- a/docs/animation/asm.md
+++ b/docs/animation/asm.md
@@ -1,0 +1,70 @@
+ASM Files
+===========
+
+ASM Files are the meat of the animation API. These define how the animation is carried out and how to use the clips defined in the armature file.
+
+Concepts
+----------
+
+The ASM contains _parameters_, _clips_, _states_, and _transitions_. The simplest ones to explain are states and transitions, so I'll start with those
+
+### States
+
+The Animation _State_ Machine can be in many different _states_. You define which states there are in the states section.
+Each _clip_ implements a state, telling the ASM how to animate it.
+
+### Transitions
+
+Right now, the only uses for transitions are to define which states can go to which other states, no way to make a transition-only
+clip exists. In the code, the only uses for the `transitions` member of AnimationStateMachine is to verify that you said a transition
+can occur between the current state and the desired next state.
+
+### Parameters
+
+!!! note
+    
+    Parameters are called `TimeValues` in the code
+
+Each parameter can either be defined in the ASM or at load-time. Load-time parameters are usually of the type `VariableValue`, allowing you to set them from within your code.
+Other kinds are `SimpleExprValue`, allowing you to do calculations relative to the current game time, `CompositionValue`, allowing you to chain parameters together and `ParameterValue`, allowing you to reference other parameters.
+A parameter is an input to a clip, telling it how far into the animation to animate.
+
+
+### Clips
+
+!!! note
+    
+    Clips can either be ASM-clips, ones that are defined in the ASM, or armature-clips, ones that are defined in the armature file.
+    For the rest of this page, "clips" will refer to ASM-clips unless otherwise stated
+
+A clip controls what a state will do. They can be one of many types, the easiest to use being the `TimeClip`. These animate a given clip by a given parameter.
+Other kinds include the `IdentityClip`, which does nothing, the `ClipReference`, which should "only be used for debugging" and just does the same as another clip,
+the `SlerpClip`, which blends between two clips, useful for transitions, and the `TriggerClip`, used to both animate a clip and fire an event when its parameter goes positive.
+
+Code API
+----------
+
+TODO
+
+
+File Format
+-------------
+
+The ASMs have a slightly strange and confusing file format, but here's how it works.
+
+First, an example:
+```json
+{
+  "parameters": {
+    "anim_cycle": ["/", "#cycle_length"]
+  },
+  "clips": {
+    "default": ["apply", "forgedebugmodelanimation:block/rotatest@default", "#anim_cycle" ]
+  },
+  "states": [
+    "default"
+  ],
+  "transitions": {},
+  "start_state": "default"
+}
+```
--- a/docs/animation/intro.md
+++ b/docs/animation/intro.md
@@ -2,5 +2,29 @@ Intro to the Animation API
 ===============================
 
 The Forge Animation API lets you animate json (and B3D) models.
+Before you start reading this, you should know how vanilla JSON models are created, and should have 
+read the documentation on blockstates
 
-Terms
+
+First, some terms:
+
+| Term | Description |
+| --- | --------- |
+| Model File | Your JSON file to be animated |
+| Armature File | The JSON file which defines the joints and clips of a model |
+| Clip | A collection of variables to apply to certain joints. Sort of an "animation group" |
+| Joint | A name for a json cube |
+| ASM File | Defines how to use the armature file to animate the model |
+| Parameter | Some value that is used to control what part of a clip we are on |
+| State | A state for the animation to be in, controls what clips are active |
+| Transitions | Controls which states go into what other states |
+| Event | An event caused by animation, readable by the thing being animated |
+ 
+ 
+Location Conventions
+-----------------------
+ 
+ ASM Files are normally stored in the `asms/block` for blocks or `asms/item` for items and so on.
+ 
+ Armature files must be stored in the `armatures` folder, and should be in the `blocks` subfolder for blocks and so on.
+
--- a/docs/animation/intro.md
+++ b/docs/animation/intro.md
@@ -1,0 +1,6 @@
+Intro to the Animation API
+===============================
+
+The Forge Animation API lets you animate json (and B3D) models.
+
+Terms
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 site/
+.idea/
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -16,6 +16,10 @@ pages:
     - Intro to Blockstates: 'blockstates/states.md' 
     - Intro to Blockstate JSONs: 'blockstates/introduction.md'
     - Forge Blockstate JSON: 'blockstates/forgeBlockstates.md'
+  - Animation API:
+    - Intro to the Animation API: 'animation/intro.md'
+    - Armatures: 'animation/armature.md'
+    - Animation State Machines: 'animation/asm.md'
   - Tile Entities:
     - Special Renderer: 'tileentities/tesr.md'
   - Items:


docs/animation/intro.md
```diff
@@ -0,0 +1,30 @@
...
+read the documentation on blockstates
+
+
+First, some terms:
+
+| Term | Description |
+| --- | --------- |
+| Model File | Your JSON file to be animated |
+| Armature File | The JSON file which defines the joints and clips of a model |
+| Clip | A collection of variables to apply to certain joints. Sort of an "animation group" |
```
williewillus: not accurate. clip = set of transformations that can be applied to a joint

docs/animation/intro.md
```diff
@@ -0,0 +1,30 @@
...
+| Joint | A name for a json cube |
```
williewillus: not accurate: joint = any part of the model you want to consider as one cohesive unit for animation purposes. doesn't have to be the json system, as for example b3d models can have joints defined too

docs/animation/intro.md
```diff
@@ -0,0 +1,30 @@
...
+| ASM File | Defines how to use the armature file to animate the model |
+| Parameter | Some value that is used to control what part of a clip we are on |
+| State | A state for the animation to be in, controls what clips are active |
+| Transitions | Controls which states go into what other states |
+| Event | An event caused by animation, readable by the thing being animated |
```
williewillus: How about "A way to receive notifications in-code when the animation reaches a certain point"

docs/animation/intro.md
```diff
@@ -0,0 +1,30 @@
...
+
+First, some terms:
+
+| Term | Description |
+| --- | --------- |
+| Model File | Your JSON file to be animated |
+| Armature File | The JSON file which defines the joints and clips of a model |
+| Clip | A collection of variables to apply to certain joints. Sort of an "animation group" |
+| Joint | A name for a json cube |
+| ASM File | Defines how to use the armature file to animate the model |
```
williewillus: this seems really vague

docs/animation/armature.md
```diff
@@ -0,0 +1,147 @@
...
+ "joints": {
+  "joint_name": {"index in model file": [1.0]}
+ }
+```
+
+- `joint_name` is the name of the joint
+- `index in model file` is a 0-indexed number denoting which model element this joint controls
+
+Not all elements need to have a joint, only the ones you are animating.
+
```
williewillus: maybe add a note that the `[1.0]` is required(?) but unused

docs/animation/armature.md
```diff
@@ -0,0 +1,147 @@
...
+- offset_x, offset_y, offset_z - translation
+- scale - uniform scaling
+- scale_x, scale_y, scale_x - scaling on certain axes
+- axis_x, axis_y, axis_z - rotation axes
+- angle - rotation angle
+
+and if #3875 is merged, there will be:
+
+- origin_x, origin_y, origin_z - rotation origin
+
```
williewillus: missing explanation for `samples`, and where the input values for joint clips come from

docs/animation/asm.md
```diff
@@ -0,0 +1,70 @@
...
+
+### Clips
+
+!!! note
+    
+    Clips can either be ASM-clips, ones that are defined in the ASM, or armature-clips, ones that are defined in the armature file.
+    For the rest of this page, "clips" will refer to ASM-clips unless otherwise stated
+
+A clip controls what a state will do. They can be one of many types, the easiest to use being the `TimeClip`. These animate a given clip by a given parameter.
+Other kinds include the `IdentityClip`, which does nothing, the `ClipReference`, which should "only be used for debugging" and just does the same as another clip,
```
williewillus: `"only be used for debugging"` what?

williewillus:
For all of your json examples, I would recommend making something like an expanded grammar,
 like [this](https://gist.github.com/williewillus/9ebb0d04329526e31564) but more explanations.

also recommend using javascript on the syntax highlighter instead of json so it won't complain about as many things (like // comments which aren't valid json but are valid javascript)

williewillus:
Good start at documenting the monster, though :P

docs/animation/armature.md
mincrmatt12: It has something to do with scale, got to check the source

docs/animation/asm.md
mincrmatt12: it says that in the source code, idk

docs/animation/armature.md
williewillus: fry said it was unused, something about blending the joint clips/TRSR's together

docs/animation/asm.md
williewillus: when writing documentation, "idk" is not a good thing to be saying ;p

I think this is the code-side deserialization of a `#` reference in the ASM json, that's what the comment means. So it's not that you should only use it for debugging, it's just that you'll find it rare to construct one in code instead of specifying it in json

docs/animation/armature.md
```diff
@@ -0,0 +1,147 @@
+Armature Files
+==============
+
+Armature Files define joints and clips for animating a model.
+
+File Structure
+----------------
+
+An example armature file, taken from the forge debug mod (requires #3875)
```
elifoster: What is this number referring to?

docs/animation/armature.md
```diff
@@ -0,0 +1,147 @@
...
+
+### Events
+
+Each clip can fire events, formatted like this:
+```json
+"events": {
+  "event_time": "event_text"
+}
+```
+For more information about events and what `event_text` means, see the page on ASMs.
```
elifoster: Link to the actual ASM page here.

docs/animation/asm.md
```diff
@@ -0,0 +1,70 @@
+ASM Files
```
elifoster: Change this to "Animation State Machine Files".

docs/animation/asm.md
```diff
@@ -0,0 +1,70 @@
...
+===========
+
+ASM Files are the meat of the animation API. These define how the animation is carried out and how to use the clips defined in the armature file.
```
elifoster: Change this to "Animation State Machine (ASM) Files are the meat..."

docs/animation/intro.md
```diff
@@ -0,0 +1,30 @@
+Intro to the Animation API
+===============================
+
+The Forge Animation API lets you animate json (and B3D) models.
```
elifoster: `json` -> `JSON`

docs/animation/intro.md
elifoster: Link to Animation State Machine page here perhaps

docs/animation/armature.md
mincrmatt12: forge pr

docs/animation/intro.md
mincrmatt12: i plan to cover b3d-specific stuff at the end of the docs, because very few mods i've seen actually use b3d models.

docs/animation/armature.md
mincrmatt12: looking in the code it takes the final transformation and multiplies it by that

mincrmatt12: looking at the b3d specification, it appears to be put there because b3d has it

mincrmatt12:
@williewillus IMO for a beginner who just wants to animate a model a grammar file seems a bit over the top, but i'll include them anyways

docs/animation/intro.md
pWn3d1337: I think .obj is important as there are conversion tools for techne/tabula models and those were common in 1.7.

mincrmatt12: AFAIK the API doesn't actually support .obj models

pWn3d1337: Are you sure on that? That's a would be a problem, because forge supports loading of json, obj and b3d. So it should work on all 3. If not, we need to open another issue :-/.

mincrmatt12: It doesn't, because obj has no builtin way of specifying the joints. You could do it by implementing a lot of the animation interfaces for OBJ stuff, but then you would need forge specific extensions. Or some way to define meshes with the semi-standard "g" tag

pWn3d1337: That's really bad. I have to look for a b3d converter.

https://en.wikipedia.org/wiki/Wavefront_.obj_file the o and g tags could work.

wiiv:
Hello, I've been playing around with this and I'd like to know if the "origin" variables are already implemented.

tterrag1098:
What's left to be done here? I would really love for this to be merged soon.

docs/animation/armature.md
```diff
@@ -0,0 +1,183 @@
...
+  }
+}
+
+```
+
+The file is organized in two sections, joints and clips.
+
+Joints
+--------
+Each joint defines a connection to the model file. A joint can animate any number of elements in the model file, but they are all animated with the same input value and transforms
```
williewillus: "a connection to the model file" feels a bit ambiguous. How about

`A joint defines an object that animates together as a single unit in the animation together. For example, in the vanilla json model format, multiple elements can belong to the same joint.`

docs/animation/armature.md
```diff
@@ -0,0 +1,183 @@
...
+
+<joint_definition> ::= {
+    <string>: [ <float> ] // index_model, joint_weight
+}
+
+```
+
+- `joint_name` is the name of the joint
+- `index_model` is a 0-indexed number denoting a model element this joint controls. Must be a string (see example)
+- `joint_weight` is a weight (0-1) of how much this element will animate. The transformation matrix caused by a clip animating this element will be multiplied by this number.
```
williewillus: according to fry a while ago the weights are unused, is that still the case?

also by "transformation matrix will be multiplied by this number" what does that mean, the transform is lerped, or the whole matrix is just scalarly multiplied, or what?

docs/animation/armature.md
```diff
@@ -0,0 +1,183 @@
...
+    "variable": <variable>,
+    "type": "uniform",
+    "interpolation": <interpolation>,
+    "samples": [ float, ... ]
+}
+
+
+```
+
+- loop: whether or not the parameter value can continue increasing and the animation loops, or if it stops at the ending state
```
williewillus: not sure what "parameter value can continue increasing and the animation loops" means, the grammar isn't quite clear

docs/animation/armature.md
```diff
@@ -0,0 +1,183 @@
...
+
+- offset_x, offset_y, offset_z - translation
+- scale - uniform scaling
+- scale_x, scale_y, scale_x - scaling on certain axes
+- axis_x, axis_y, axis_z - rotation axes
+- angle - rotation angle
+
+and if PR #3875 is merged, there will be:
+
+- origin_x, origin_y, origin_z - rotation origin
```
williewillus: the pr got merged (i think)

docs/animation/armature.md
```diff
@@ -0,0 +1,183 @@
...
+
+### Events
+
+Each clip can fire events, formatted like this:
+```javascript
+<event> :: {
+    <event_time>: "event_text"
+}
+```
+For more information about events and what `event_text` means, see the page on ASMs.
```
williewillus: use the full name here and link said page

docs/animation/armature.md
```diff
@@ -0,0 +1,183 @@
...
+
+
+```
+
+- loop: whether or not the parameter value can continue increasing and the animation loops, or if it stops at the ending state
+
+### Joint Clips
+Each `joint_clip` is a set of variables to change for a joint. The `type` attribute is currently ignored, but must be `"uniform"`.
+
+`samples` defines what value the animation will take on, and its interpretation depends on the value of `interpolation`.
```
williewillus: i don't think this really captures what samples really means, but unfortunately I can't think of something better.
Maybe `samples defines the boundary and granularity of values used for the transform`? meh

docs/animation/asm.md
```diff
@@ -0,0 +1,312 @@
...
+
+An example:
+```java
+@Nullable
+private final IAnimationStateMachine asm;
+private final VariableValue cycle = new VariableValue(4);
+
+public Spin() {
+    if (FMLCommonHandler.instance().getSide() == Side.CLIENT) {
+        asm = ModelLoaderRegistry.loadASM(new ResourceLocation(MODID, "asms/block/rotatest.json"), ImmutableMap.of("cycle_length", cycle));
```
williewillus: is this actually side safe? i'm not sure how loading of methods works but we don't want to show dediserver-crashing code in the official docs

docs/animation/asm.md
```diff
@@ -0,0 +1,312 @@
...
+- `r`: output = floor(input / arg) * arg;
+- `R`: output = ceil(input / arg) * arg;
+- `f`: output = input - floor(input / arg) * arg;
+- `F`: output = ceil(input / arg) * arg - input;
+
+##### Example explanations:
+- input + 4
+- (input / 5) + 1
+- input + 2 + value of parameter other
+- input + value of parameter other + value of parameter cycle given input 3
```
williewillus: I'm reading this md file unrendered but maybe move these to be right under their respective examples?

docs/animation/asm.md
```diff
@@ -0,0 +1,312 @@
...
+
+- blend the closed clip to the open clip, giving both clips the unaltered time as input and blend progress `#progress`.
+- blend the result of the move clip when given the input parameter `mover` to the end clip with the unaltered time as the input with blend progress `#progress`.
+
+### States
+
+The states section of the file is simply a list of all possible states.
+For example 
+```json
+"states": {
```
williewillus: this should be a json array right?

mincrmatt12:
sorry for the delay, (life happens), i'm going to be working on this more in the coming days, and hopefully finishing it

docs/animation/asm.md
mincrmatt12: actually you are right, this should be proxy'd (oops)

docs/animation/armature.md
mincrmatt12: it did indeed

docs/animation/armature.md
mincrmatt12: its more like keyframes in traditional animation

docs/animation/armature.md
mincrmatt12: i remember seeing it being used it ModelBlockAnimation, lines 545-570.

docs/animation/asm.md
mincrmatt12: i think it might be better this was, so you can see some examples, get the explanation of the entire parameter type, and then you can check to see the explanations of the examples

docs/animation/armature.md
mincrmatt12: sound good to me

ichttt:
I don't think the abbreviate Animation State Machine to ASM should be used (or at least the page should be  renamed), as the term is already used by forge with a different meaning, and most people assoisiate assemly bytecode with the term, and forge has things called ASMDataTable, ASMEventHandler, ASMTransformer and so on. But I would like some other people's feedback on that

mincrmatt12:
You have a point, but saying Animation State Machine every time is very verbose. Perhaps putting a warning /note admonition at the top would work

Anime-Quatroctus:
implementing.md "Your block must also render in the ENTITYBLOCK_ANIMATED render layer." This is not true, if the BlockState has Properties.StaticProperty. And the TESR is an AnimationTESR it will get the model from the static=false variant and use that instead. Everything else looks amazing, what else is there left to do?

mincrmatt12:
All that needs to be done is finish the documentation on implementing the api for items and entities. Also thanks for telling me about the block stuff, most of what I know is from reading the source and experimentation.

docs/animation/implementing.md
```diff
@@ -0,0 +1,42 @@
+Implementing the API
+======================
+
+Depending on what you want to animate with the API, code-side implementation is a bit different.
+Documentation on the ASM api itself (for controlling the animation) is found on the [ASM][asm] page because it is independent of what
+you are animation.
```
MajorTuvok: Shouldn't it be animating?

docs/animation/implementing.md
```diff
@@ -0,0 +1,42 @@
...
+{
+    @Override
+    public void handleEvents(Chest chest, float time, Iterable<Event> pastEvents)
+    {
+        chest.handleEvents(time, pastEvents);
+    }
+}); 
+```
+
+In this example, we've overrided the `handleEvents()` callback when we registered the TESR because the implementation is simple, but you could easily subclass
```
MajorTuvok: And maybe 'overridden' here?

MajorTuvok:
(I'm no native speaker (and especially unsure about the correct 3.form of (to) override, but I'm pretty sure it is not the standard - ed?). Please correct me if I'm wrong...)

docs/animation/implementing.md
Anime-Quatroctus: Please make a thread on http://www.minecraftforge.net/forum/forum/70-modder-support/

MajorTuvok: ???
I just wanted to help with spelling mistakes... 
I wasn't even asking questions...

Anime-Quatroctus: Im sorry I misinterpreted what you said and posted. Yes that should be "animating" instead of animation.

MajorTuvok: No Problem...
I modified it in hopes of making it clearer to others, to ehat I'm referring to.

docs/animation/implementing.md
mincrmatt12: Yes, you are correct. Fixed.

docs/animation/implementing.md
mincrmatt12: Fixed.

docs/animation/implementing.md
```diff
@@ -0,0 +1,103 @@
...
+you are animating.
+
+Blocks
+--------
+
+Animations for blocks are done with the AnimationTESR, which is a FastTESR. Because of this, having a TileEntity for your block
+is necessary. Your TileEntity must provide the `ANIMATION_CAPABILITY`, which is recieved by calling its `.cast` method with your
+ASM. Your block must also render in the `ENTITYBLOCK_ANIMATED` render layer if you do not provide a `StaticProperty` in the block's blockstate.
+
+The `StaticProperty` is a property you can add to your block's blockstate by adding `Properties.StaticProperty` to the list of your block's properties inside
```
PaleoCrafter: Although it's kind of obvious from the name, you should mention that all parts of the model that *can* be static should probably also be rendered in the static state, that's the whole point.

docs/animation/implementing.md
```diff
@@ -0,0 +1,103 @@
...
+    public void handleEvents(Chest chest, float time, Iterable<Event> pastEvents)
+    {
+        chest.handleEvents(time, pastEvents);
+    }
+}); 
+```
+
+In this example, we've overridden the `handleEvents()` callback when we registered the TESR because the implementation is simple, but you could easily subclass
+AnimationTESR to achieve the same effect. `handleEvents()` callback for blocks takes two arguments: the tile entity being rendered, and an iterable of the events.
+The call to ``chest.handleEvents`` calls a method located in the fictional `Chest` TileEntity, as the ASM is not accessible inside of the handleEvents method. is not accessible inside of the handleEvents method.
```
PaleoCrafter: Accidentally repeated yourself, it seems. `handleEvents` in backticks please, and I think you've got one to many on each side of `chest.handleEvents`.

docs/animation/asm.md
```diff
@@ -0,0 +1,307 @@
...
+The ASM contains _parameters_, _clips_, _states_, and _transitions_. The simplest ones to explain are states and transitions, so I'll start with those
+
+### States
+
+The Animation _State_ Machine can be in many different _states_. You define which states there are in the states section.
+Each _clip_ implements a state, telling the ASM how to animate it.
+
+### Transitions
+
+Right now, the only uses for transitions are to define which states can go to which other states, no way to make a transition-only
```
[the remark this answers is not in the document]
mincrmatt12: I realize that, but from the perspective of someone who just wants to animate something this definition makes it clear that you are not able to say, define an animation that plays _during_ a transition between two states

docs/animation/asm.md
```diff
@@ -0,0 +1,307 @@
...
+
+Different types of parameters have different formats for `<parameter_definition>`, and the simple ones are:
+
+- `IdentityValue`: the string `#identity`,
+- `ParameterValue`: the parameter to reference, prefixed with `#`, e.g. `#my_awesome_parameter`
+- `ConstValue`: a number to use as the constant to return
+
+Other kinds are a bit more complicated
+
+#### `SimpleExprValue`
```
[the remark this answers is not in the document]
mincrmatt12: In my opinion, it makes more sense to leave the names as they are in the code since being able to look up the implementation may be useful for people.

docs/animation/asm.md
mincrmatt12: The idea I had with the examples was to first show examples, then explain the parameter, and then show the examples again with their meaning now that the reader has read the explanation.

docs/animation/asm.md
```diff
@@ -0,0 +1,307 @@
...
+```json
+[ "compose", "#cycle", 3]
+[ "compose", "#test", "#other"]
+[ "compose", [ "+", 3], "#other"]
+[ "compose", [ "compose", "#other2", "#other3"], "#other"]
+```
+##### Explanation
+
+`CompositionValue` takes two parameter definitions as inputs, and does `value1(value2(input))`. In other words, it chains
+the two inputs, calling the second one with the given input, and the first one with the output of the second one.
```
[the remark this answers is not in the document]
mincrmatt12: If the reader knows about that, they wouldn't need the more detailed explanation.

docs/animation/asm.md
```diff
@@ -0,0 +1,307 @@
...
+}
+```
+
+As with parameters, different kinds of clips have different formats for `<clip_definition>`, but the simple ones are:
+
+- `IdentityClip`: the string `#identity`
+- `ClipReference`: the clip name prefixed with `#`, e.g. `#my_amazing_clip`
+- `ModelClip`: a model resource location + `@` + the name of the armature-clip, e.g. `mymod:block/test@default` or `mymod:block/test#facing=east@moving`
+
+#### `TimeClip`
```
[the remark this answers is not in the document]
mincrmatt12: And again I would rather keep the names for easy searching, as it applies more here as creating custom clips may be required for adding animations to custom models.

docs/animation/implementing.md
mincrmatt12: Curse the incompatibilities between reStructuredText and Markdown!

docs/animation/intro.md
```diff
@@ -0,0 +1,28 @@
...
+!!! note
+    
+    Although you can use B3D models with the Animation API, most of this documentation will assume you
+    are using JSON files. (TODO) See the page on using B3D models for more information
+
+The Animation API is made up of two main components: armature files and animation state machine (ASM) files.
+Armature files define joints and clips for JSON files. Joints are names for cubes in the model file with weights, while
+clips are a set of transformations to apply to joints. ASM files define the various states an animation can be in, as well as what
+transitions exist between those states. They also define the parameters for an animation, generally used as inputs to clips, but can also
+trigger events. Events are essentially a way to recieve in-code notifications when the animation reaches a certain point, but can also trigger
```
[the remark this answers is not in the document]
mincrmatt12: There's an option in the armature file to make a clip automatically loop, so value outside of the 0-1 range have their integer part removed to create a looping animation, so the input to a clip is not always 0-1

mincrmatt12:
Thanks for the input, I've tried to address the changes that you're requesting.

docs/animation/asm.md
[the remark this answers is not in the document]
PaleoCrafter: Ah, I understand where you're coming from now. Maybe say that a transition is not an animation unto itself or a continuous change between two states but instead an instantaneous switch to a new state. Also mention that these smooth transitions have to be modelled by an extra state, as you point out yourself that that might be a source of confusion the elimination of which is the entire point of documenting this.

docs/animation/asm.md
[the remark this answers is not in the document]
PaleoCrafter: It's okay to mention the implementation name (in parentheses maybe), but the primary focus should be on the high level concepts and this for example represents mathematical expressions for useful calculations, while `SimpleExprValue` doesn't carry much information.

Besides, most people reading these docs won't ever see any of the parameter implementation classes except for `VariableValue`. The point of documenting this is to not have to look at the implementation.

docs/animation/asm.md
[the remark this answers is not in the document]
PaleoCrafter: I'm not saying drop the explanation, just to mention the comparison for people who *are* familiar with the concept. It's also helpful if people maybe don't understand your explanation and want to look it up themselves.

docs/animation/asm.md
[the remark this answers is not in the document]
mincrmatt12: Ok, I see what you mean.

mincrmatt12:
Alright, now I think I fixed the issues with class names

mincrmatt12:
Are there any more changes that need to be made before this is merged?

PR #103, opened 2017-05-17
More block interaction docs

### LINKED ISSUE #94: Needed block interaction methods
<del>I think it would be useful to have lists of "hooks" (not sure what else you'd call them) in relevant pages. For example, the Block page says that it would be too long to document every single event like function. I agree with that. However, I think having a complete list of them with a brief description would not be that difficult and could be incredibly useful. This would also apply for items and anything else with a similarly huge amount of functions like this. </del>

Here is a list of Block Interaction functions that are not documented yet (1.10.2-12.18.3.2254):
* [ ] onBlockDestroyedByPlayer
* [ ] neighborChanged
* [ ] onBlockAdded
* [ ] breakBlock: Done at #103
* [ ] onBlockDestroyedByExplosion
* [x] onBlockActivated
* [ ] onEntityWalk: Done at #103
* [ ] onBlockClicked
* [ ] onEntityCollidedWithBlock
* [ ] onBlockPlacedBy
* [ ] eventReceived
* [ ] onLanded
* [ ] onBlockHarvested
* [ ] removedByPlayer
* [ ] onBlockExploded
* [ ] onPlantGrow

elifoster:
***This PR is not ready! I am submitting it to get feedback on what I currently have and assistance as I go!***

This is to resolve #94 and get rid of those nasty "*Coming Soon!*"s from the block interaction page.

### DIFF
--- a/docs/blocks/interaction.md
+++ b/docs/blocks/interaction.md
@@ -64,7 +64,25 @@ Another common use for activation is, well, activation. This can be something li
 
 Player Break/Destroy
 --------------------
-*Coming Soon*
+### `breakBlock`
+
+This method is called when the block is broken. It is called on the server only.
+
+It is called before the tile entity, if it has one, is updated (and therefore removed from the world), but after the block has been replaced by another in the chunk.
+
+Its full signature is as follows:
+```java
+public void breakBlock(World worldIn, BlockPos pos, IBlockState state)
+```
+
+A common usage for this method is to handle the dropping of a block's inventory items.
+
+#### Return Value
+
+It has no return value.
+
+#### Parameters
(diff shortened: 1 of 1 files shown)


G-Rath:
I think it would be best to have parameters before return, since that's how it's laid out in JavaDoc, and for `onBlockActivated` currently.

elifoster:
@G-Rath Definitely agree. That's how most doc systems, from what I've seen, lay it out.

tterrag1098:
onEntityWalk is not really what I was thinking about when I wrote in that section header, though I think it does fit. But I think the more often used methods like collision/selection bounds should be higher priority.

Akxe:
Add sideness information about events. Is it server event, client, or both?

PR #108, opened 2017-06-09
Migrate docs to a theme based on the 'new' Forge brand

PaleoCrafter:
This has been in the works ever since the new forums were introduced, I just never got around to finishing it. The theme can be previewed under my [compiled fork](http://paleocrafter-mcforge.readthedocs.io/en/new-brand-theme/) (the content is outdated, but that doesn't really matter for the theme). I'm open for feedback on the general look and feel and potential issues.

The new theme has the following benefits over the default RTD one:
- Better handling of too wide text/code
- Collapsible navigation items
- Much better customisability due to controlling the entire theme

As a technical note, probably for @tterrag1098:
With this (re)design, I also introduce an additional dependency for mkdocs. In order for it to work, the python-requirements.txt file needs to be specified under Advanced Settings on RTD.

**Note:**
The stylesheet introduced in this PR is not to be edited manually, it is the result of combining multiple SASS modules from [MinecraftForge/Web](https://github.com/MinecraftForge/Web/blob/master/sass/documentation.scss). In the future, I hope to remove the stylesheet from this repo and automate the deployment from the Web stuff, then all style changes can go through a central place.


forge_theme/base.html
```diff
@@ -0,0 +1,117 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="utf-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <title>{% if page_title %}{{ page_title }} - {% endif %}{{ site_name }}</title>
+
+    <link rel="apple-touch-icon" sizes="180x180" href="http://forge.mineformers.de/images/favicons/apple-touch-icon.png">
```
kashike: Would it not be better to link to minecraftforge.net?

PaleoCrafter: Whoops... forgot to change this from my old design. Technically it'd be best if we had a CDN to host all this stuff on, but that might only happen some time later down the road.

PaleoCrafter:
@loordgek: Code-wise the sticky navigation is a simple change to a `fixed` position after the scrolling reaches the navigation, so it might be a little snappy depending on the increments of your scroll wheel. Additionally, it goes sticky not right after the navigation is hit but a little earlier, otherwise it'd be even jumpier since the navigation isn't flush with the top of the viewport.
I've changed the script slightly such that it uses font-size based measurements (i.e. `rem` units), so there isn't any discrepancy between the style and the code and it should be *slightly* less snappy.

diesieben07:
What about `position: sticky` with a polyfill?

PaleoCrafter:
Welp, wasn't aware of `position: sticky`, never read about it or came across it. Thanks, @diesieben07, it appears to work fine, can't really say if it's any less snappy, though.

loordgek:
yaay no longer snappy
works fine now

loordgek:
the green is a bit to green, bit hard to read "Important"
http://paleocrafter-mcforge.readthedocs.io/en/new-brand-theme/styleguide/

make it darker 
http://mcforge.readthedocs.io/en/latest/styleguide/

PaleoCrafter:
Thanks for noticing that. I had changed the colours previously, the changes got lost somehow, though.

MinecraftPlaye:
I would recommend a larger font size for better readability.
Also why do some sites on the sidebar have the plus icon on them and others don't? 
Actually all of them are opening some kind of menu with more sites to show.
I would also change the theme for the marker which shows you the selected site.
For example change this marker to white and all other sites have the background color applied.

PaleoCrafter:
First of all, regarding the plus icons: Only "categories" that have sub*pages* will be an actual collapsible element (i.e. the plus icon). The smaller links below individual *sites* are sections, i.e. links to anchors within the page. For stuff like the "Networking" category which have an index page, it was requested that the top-most element links to that category, so combining it with a collapsible is a little difficult.
The marker for the selected site (grey background, bold top-level text), however, is fairly obvious in my opinion. If more people demand a change, I will do so, though.
Regarding the font size: I kept it at 14px since that matches the forums, but if enough people want it, I'll gladly change the size to 16px (which I personally prefer).

PR #114, opened 2017-07-02
Documented available RegistryEvents

Drakmyth:
Given that it is now "recommended and HIGHLY encouraged" to use RegistryEvents over registering during loading events, I thought it might be helpful to list which objects can be registered using RegistryEvents. I also added a link to the Loading Stages page to encourage modders not to continue doing their registrations there. In response to #85, I also included a bit of information around why RegistryEvents should be preferred.

Not having an in-depth knowledge of the inner-workings or direction motivations of Forge itself, if anything I've added is incorrect I'm happy to make changes.

### DIFF
--- a/docs/concepts/registries.md
+++ b/docs/concepts/registries.md
@@ -23,6 +23,8 @@ public void registerBlocks(RegistryEvent.Register<Block> event) {
 
 The order in which `RegistryEvent.Register` events fire is arbitrary, with the exception that `Block` will *always* fire first, and `Item` will *always* fire second, right after `Block`. After the `Register<Block>` event has fired, all [`ObjectHolder`][ObjectHolder] annotations are refreshed, and after `Register<Item>` has fired they are refreshed again. They are refreshed for a third time after *all* of the other `Register` events have fired.
 
+`RegistryEvent`s are currently supported for the following types: `Block`, `Item`, `Potion`, `Biome`, `SoundEvent`, `PotionType`, `Enchantment`, `IRecipe`, `VillagerProfession`, `Entity`
+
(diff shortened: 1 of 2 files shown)


docs/concepts/registries.md
```diff
@@ -23,6 +23,8 @@ public void registerBlocks(RegistryEvent.Register<Block> event) {
 
 The order in which `RegistryEvent.Register` events fire is arbitrary, with the exception that `Block` will *always* fire first, and `Item` will *always* fire second, right after `Block`. After the `Register<Block>` event has fired, all [`ObjectHolder`][ObjectHolder] annotations are refreshed, and after `Register<Item>` has fired they are refreshed again. They are refreshed for a third time after *all* of the other `Register` events have fired.
 
+`RegistryEvent`s are currently supported for the following types: `Block`, `Item`, `Potion`, `Biome`, `SoundEvent`, `PotionType`, `Enchantment`, `IRecipe`, `VillagerProfession`, `Entity`
```
Drakmyth: I got this list by looking at `net.minecraftforge.registries.GameData`. I'm not 100% certain the `Entity` is supported, but it seemed like that was the intent so I included it here. I don't have time to test it right now, but I intend to at some point over the next few days. If that is not accurate, it can certainly be removed.

docs/conventions/loadstages.md
```diff
@@ -6,8 +6,15 @@ There are some other events that are important too, depending on what your mod d
 Each of these stages occurs at a different point in the loading stage and thus what can be safely done in each stage varies.
 
 !!! note
+
     Loading stage events can only be used in your `@Mod` class, in methods marked with the `@EventHandler` annotation.
 
+!!! important
+
+    Many objects that were previously registered in loading stage event handlers (Blocks, Items, Recipes, etc.) should now be registered via [RegistryEvents][registering].
+    This is to pave the way to being able to reload mods dyanamically at runtime, which can't be done using loading stages (as they are fired once upon application startup).
```
sjx233: dynamically

docs/concepts/registries.md
tterrag1098: Using the raw `<Entity>` event does not seem to work, I believe you still need to go through the old methods.

LexManos: There is no Entity registry. There is a EntityEntry registry.

PR #118, opened 2017-07-19
SimpleImpl: Add warning about insecure handling of packets on the server

JamiesWhiteShirt:
Too many mods are vulnerable to arbitrary chunk generation due to insecure packet handling. I feel defensive packet handling deserves to be mentioned in the documentation.

I kind of want to write "Never trust the client", but I don't want it too be pushy and it feels a bit out of scope. Thoughts?

### DIFF
--- a/docs/networking/simpleimpl.md
+++ b/docs/networking/simpleimpl.md
@@ -83,7 +83,9 @@ It is recommended (but not required) that for organization's sake, this class is
 
     Be defensive when handling packets on the server. A client could attempt to exploit the packet handling by sending unexpected data.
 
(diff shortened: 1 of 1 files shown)


PaleoCrafter:
Maybe you could elaborate a bit more on why exactly arbitrary chunk generation is bad. The reader of the docs might well not be precisely aware of the inner works of Minecraft, so some explanation of how loading arbitrary chunks can cause issues would be great.

JamiesWhiteShirt:
I'll look into expanding the note on the problem of arbitrary chunk generation.

JamiesWhiteShirt:
Updated with a better explanation of arbitrary chunk generation with big scare words.

tterrag1098:
Looks good to me, thanks.

PR #122, opened 2017-08-05
Correct example MessageHandler to use addScheduledTask(...)

joazlazer:
- Uses an anonymous implementation of `Runnable()` and adds it via `addScheduledTask()` to prevent threading errors
- Updates the warning block to indicate the changes
- Changes `Items.diamond` to its correct name, `Items.DIAMOND`

**Justification**: What is the point of doing something incorrectly just to correct it a few lines down? Docs should show only the correct implementations as examples.

**_Side question_**: Should the documentation use an anonymous `Runnable` implementing class (as is used with this PR) or with a _block lambda expression_? (**lines 65-70** would instead read as follows)

```java
serverPlayer.getServerWorld().addScheduledTask(() -> {
    serverPlayer.inventory.addItemStackToInventory(new ItemStack(Items.DIAMOND, amount));
});
```

### DIFF
--- a/docs/networking/simpleimpl.md
+++ b/docs/networking/simpleimpl.md
@@ -61,7 +61,13 @@ public class MyMessageHandler implements IMessageHandler<MyMessage, IMessage> {
     EntityPlayerMP serverPlayer = ctx.getServerHandler().playerEntity;
     // The value that was sent
     int amount = message.toSend;
-    serverPlayer.inventory.addItemStackToInventory(new ItemStack(Items.diamond, amount));
+    // Excecute the action on the main server thread by adding it as a scheduled task
+    serverPlayer.getServerWorld().addScheduledTask(new Runnable() {
+      @Override
+        public void run() {
+            serverPlayer.inventory.addItemStackToInventory(new ItemStack(Items.diamond, amount));
+        }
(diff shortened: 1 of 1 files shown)


PaleoCrafter:
I'd say go with a lambda. We're "officially" allowed to use Java 8 features in mods as of 1.12, so we might as well take advantage of that.

docs/networking/simpleimpl.md
```diff
@@ -61,7 +61,13 @@ public class MyMessageHandler implements IMessageHandler<MyMessage, IMessage> {
     EntityPlayerMP serverPlayer = ctx.getServerHandler().playerEntity;
     // The value that was sent
     int amount = message.toSend;
-    serverPlayer.inventory.addItemStackToInventory(new ItemStack(Items.diamond, amount));
+    // Excecute the action on the main server thread by adding it as a scheduled task
```
sjx233: typo

joazlazer:
I updated it to use the lambda expression as suggested by Paleo.

tterrag1098:
Thank you for the update, the docs for this page are quite old and were written when 1.8+ was much newer. This makes much more sense now.

PR #124, opened 2017-08-09
Update locations.md language formatting

MysticalHerobine:
Update the past format 'en_US' to 'en_us'

Justification: Help out beginner modders, such as myself with up to date information.

### DIFF
--- a/docs/conventions/locations.md
+++ b/docs/conventions/locations.md
@@ -15,7 +15,7 @@ Blockstate definition files are in the JSON format and are in the `./assets/<mod
 
 ### Localizations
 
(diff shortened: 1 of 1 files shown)


MinecraftPlaye:
You might want to add this little information:
`Version 3 resource packs have to use lowercase filenames for everything.`
This change was introduced in Minecraft 1.11 and should clarify why the language codes in the msn docs aren't writtin in lowercase while Minecraft enforces this.

tterrag1098:
It would also be worth documenting that this is only true for mods that explicitly specify their RP version, and how one can go about doing that.

PR #126, opened 2017-08-21
#94: interaction documentation for onLanded

G-Rath:
#94: Added documentation about the `onLanded` block interaction method.


Akxe:
Add sideness information about this event, is it server event, client, or both?

PR #131, opened 2017-09-28
Allow theme to be switched between light and dark

PaleoCrafter:
As you might well know, the Forge forums are available in a dark and light theme. In an effort to allow a more consistent experience among all our sites, I've written a simple script to switch a colour theme on the fly. It will also be applicable to the files site, but we have some issues due to the aggressive caching there, so we won't be getting a switch there yet.
Note that these changes include adjustments to both admonitions and code highlighting in order to make them work for either theme, but I tried to keep the light variant as close to the original as possible.

As always, the changes can be viewed here: [Theme Switch Example](http://paleocrafter-mcforge.readthedocs.io/en/theme-switch/)
The switch can be found at the very bottom of the page.


tterrag1098:
On the main page here: http://paleocrafter-mcforge.readthedocs.io/en/theme-switch/

The footer appears below the viewport, requiring the user to scroll even when there is no more content. Why is this?

tterrag1098:
This blue text seems out of place:

PaleoCrafter:
The footer being pushed below the fold is an unfortunate side effect of the sidebar's styling and nothing I can really do anything about.
The links are blue there since they don't get any special styling when in the sidebar, so that's just the default dark link look. If you want me to, I can change them to shades of grey for the sidebar.

tterrag1098:
I'm no designer, so I have no advice to give on color schemes. Just that this one looks odd to me :P

elifoster:
The responsive design is kind of messed up. If you make your browser short and wide, the sidebar covers up part of the content. If you make your browser window taller, it does not go back into place, either. I'm not sure if this is specific to your fork or if it is an issue on the main site already.



I would also change the color of the previous/next buttons (`.btn`) from #DE9E59 (#E6B784 when `:active`, `:hover`, or `:focus`) to #35855E (#3E9B6E when `:active`, `:hover`, or `:focus`)



In the above, previous has the `:hover` pseudo-element enabled.

PaleoCrafter:
I've encountered the weird sidebar issue as well from time to time, but couldn't figure out any way to reproduce it reliably. It appears to be related to caching, so I've added an explicit request for a "newer" version of the stylesheets in an attempt to prevent it.
I'm personally not too keen on changing the button colours as the current colour is one of the generic Forge "brand" colors, taken straight from the logo. If the general opinion is that they should be changed, however, I'll gladly do so.

Issue #136, opened 2017-10-07
Conventions.Localization mentions en_US when it should be en_us


Thretcha:
In the the localization section of the convention page en_US is mentioned as an example language file.
It has to be en_us to actually work.

Also please be so nice and add at least one example like the following
item.<unlocalizedname>.name = desired localized name here

Bluexin:
https://github.com/MinecraftForge/Documentation/pull/148 should fix the first part

TehNut:
The second part is unnecessary in my opinion. It is a basic vanilla resource pack feature.

PR #137, opened 2017-10-07
Change en_US to en_us, as Forge doesn't accept en_US

half-cambodian-hacker-man:
Quick fix for #136.

Perhaps examples should be written also?

### DIFF
--- a/docs/conventions/locations.md
+++ b/docs/conventions/locations.md
@@ -15,7 +15,7 @@ Blockstate definition files are in the JSON format and are in the `./assets/<mod
 
 ### Localizations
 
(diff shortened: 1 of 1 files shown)


tterrag1098:
This is actually only true with a pack.mcmeta file specifying version 3.

kashike:
Version 3 [is the version specified in the MDK](https://github.com/MinecraftForge/MinecraftForge/blob/1.12.x/mdk/src/main/resources/pack.mcmeta#L4). The documentation should reflect this.

tterrag1098:
It should also reflect that most people probably don't use the example mod, and wouldn't magically know to add a pack.mcmeta file as it's never really been required before.

PR #142, opened 2017-11-24
[Don't pull] Add article for Data Fixers

williewillus:
Don't pull yet, this may change depending on a PR from @Barteks2x improving the mod fixer registration system

### DIFF
--- a/docs/datastorage/datafixers.md
+++ b/docs/datastorage/datafixers.md
@@ -31,7 +31,7 @@ Fixes are registered to and only run for one `IFixType`.
 
 The interface contains two methods:
 	* `int getVersion()`: This method returns the current save format version of your mod. When the system sees data that has an older version than the result of this method, it will run the fixer.
-	* `NBTTagCompound fixTagCompound(NBTTagCompound compound)`: Accepts the full NBT tag of the requested `IFixType` (i.e. the entire ItemStack, Chunk, or TileEntity), returns the fixed version. For ItemStacks, the top-level tag is passed, containing the Item ID, stack size, metadata, "tag" NBT tag, and capability NBT tag.
+	* `NBTTagCompound fixTagCompound(NBTTagCompound compound)`: Accepts the full NBT tag of the requested `IFixType` (i.e. the entire ItemStack, Chunk, or TileEntity). For ItemStacks, the top-level tag is passed, containing the Item ID, stack size, metadata, "tag" NBT tag, and capability NBT tag. The fixed tag should be returned, at the same hierarchy level as it was passed (i.e. return the top-level ItemStack tag).
 
 ### IDataWalker
 This interface is registered to and only run for one `IFixType`, and essentially represents the ability to discover other `IFixType`'s within the tag.
--- a/docs/datastorage/datafixers.md
+++ b/docs/datastorage/datafixers.md
@@ -31,7 +31,7 @@ Fixes are registered to and only run for one `IFixType`.
 
 The interface contains two methods:
 	* `int getVersion()`: This method returns the current save format version of your mod. When the system sees data that has an older version than the result of this method, it will run the fixer.
-	* `NBTTagCompound fixTagCompound(NBTTagCompound compound)`: Accepts the full NBT tag of the requested `IFixType` (i.e. the entire ItemStack, Chunk, or TileEntity), returns the fixed version
+	* `NBTTagCompound fixTagCompound(NBTTagCompound compound)`: Accepts the full NBT tag of the requested `IFixType` (i.e. the entire ItemStack, Chunk, or TileEntity), returns the fixed version. For ItemStacks, the top-level tag is passed, containing the Item ID, stack size, metadata, "tag" NBT tag, and capability NBT tag.
 
 ### IDataWalker
 This interface is registered to and only run for one `IFixType`, and essentially represents the ability to discover other `IFixType`'s within the tag.
--- a/docs/datastorage/datafixers.md
+++ b/docs/datastorage/datafixers.md
@@ -104,4 +104,4 @@ IDataWalker dollyWalker = new IDataWalker() {
 FMLCommonHandler.instance().getDataFixer().registerWalker(FixTypes.ITEM_INSTANCE, dollyWalker);
 ```
 
-[flattening]: http://minecraft.gamepedia.com/the_flattening
+[flattening]: https://minecraft.gamepedia.com/1.13/Flattening
--- a/docs/datastorage/datafixers.md
+++ b/docs/datastorage/datafixers.md
@@ -1,0 +1,107 @@
+The DataFixer System
+====================
+
+The DataFixer system is provided by vanilla, and allows for fine-grained migrations and adjustments to the world format. Forge expands this system to allow mods to register and use their own fixers.
+
+What would I use the DataFixer for?
+-----------------------------------
+Here are some things vanilla uses datafixers for:
+
+* Renaming the old `HealF` health tag to `Health` (1.9)
+* Replacing all integer ID's in ItemStacks with registry names (~ 1.9)
+* Renaming the ID's of all entities and tile entities from UpperCamel names like `Blaze` to proper registry names like `minecraft:blaze` (1.11)
+* Splitting entities that were variants of each other to separate ID's, e.g. `Elder` tag on `minecraft:guardian` -> `minecraft:elder_guardian` (1.11)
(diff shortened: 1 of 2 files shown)


docs/datastorage/datafixers.md
```diff
@@ -0,0 +1,107 @@
...
+### DataFixer
+This is the main entry point for invoking the fixer system, and it also manages registration of fixes and walkers.
+
+### IFixableData
+This interface represents a single "fix" you would like to make, such as one of the examples listed above.
+Fixes are registered to and only run for one `IFixType`.
+
+The interface contains two methods:
+	* `int getVersion()`: This method returns the current save format version of your mod. When the system sees data that has an older version than the result of this method, it will run the fixer.
+	* `NBTTagCompound fixTagCompound(NBTTagCompound compound)`: Accepts the full NBT tag of the requested `IFixType` (i.e. the entire ItemStack, Chunk, or TileEntity), returns the fixed version
```
raoulvdberge: "Entire ItemStack" -> Be more specific -> Do you want the nbt tag as returned by `ItemStack#serializeNBT` or the nbt tag contained in `tag` in the root tag?

williewillus:
Clarified what itemstack tag is passed

docs/datastorage/datafixers.md
```diff
@@ -0,0 +1,107 @@
...
+
+### IDataWalker
+This interface is registered to and only run for one `IFixType`, and essentially represents the ability to discover other `IFixType`'s within the tag.
+For example, a `TileEntityChest` is of fix type `BLOCK_ENTITY`, but has 27 stacks of contents which are of fix type `ITEM_INSTANCE`. Therefore, an `IDataWalker` is needed to find those stacks and fix them as well.
+
+It has one method:
+	* `NBTTagCompound process(IDataFixer fixer, NBTTagCompound compound, int version)`: Accepts the full NBT tag of the requested `IFixType`, looks for nested `IFixType`'s, and fixes them recursively using `fixer.process()`. 
+For the example above, the walker would look through `compound` for each ItemStack, and then call `fixer.process(FixTypes.ITEM_INSTANCE, stackNBT)` on it, which will recursively invoke all ITEM_INSTANCE fixers and walkers.
```
LemADEC: version parameter is missing in the call. Is it the "data version" or the "current mod version of the save format"?

docs/datastorage/datafixers.md
```diff
@@ -0,0 +1,107 @@
...
+
+### DataFixer
+This is the main entry point for invoking the fixer system, and it also manages registration of fixes and walkers.
+
+### IFixableData
+This interface represents a single "fix" you would like to make, such as one of the examples listed above.
+Fixes are registered to and only run for one `IFixType`.
+
+The interface contains two methods:
+	* `int getVersion()`: This method returns the current save format version of your mod. When the system sees data that has an older version than the result of this method, it will run the fixer.
```
LemADEC: Should we assume the NBT data includes a version tag that is handled by Forge/Vanilla? or do we have to save it somehow with our custom tags?

docs/datastorage/datafixers.md
```diff
@@ -0,0 +1,107 @@
...
+-----------------------------------
+Here are some things vanilla uses datafixers for:
+
+* Renaming the old `HealF` health tag to `Health` (1.9)
+* Replacing all integer ID's in ItemStacks with registry names (~ 1.9)
+* Renaming the ID's of all entities and tile entities from UpperCamel names like `Blaze` to proper registry names like `minecraft:blaze` (1.11)
+* Splitting entities that were variants of each other to separate ID's, e.g. `Elder` tag on `minecraft:guardian` -> `minecraft:elder_guardian` (1.11)
+* Flattening all meta-variants of an Item into separate ID's, e.g. `minecraft:wool` meta 0 -> `minecraft:white_wool` (1.13)
+
+As you can see, this can be used in a very powerful manner to implement any NBT save format fixes you would like to make, without resorting to hacks such as checking for the old tag every tick and replacing it with a new tag.
```
LemADEC: Does this also cover the remapping done through the FMLMissingMappingsEvent?

docs/datastorage/datafixers.md
```diff
@@ -0,0 +1,107 @@
...
+Mods must take care to call the Fixer system whenever they hold onto NBT data of vanilla or other mods.
+For example, consider a dolly item that allows one to store any Tile Entity into itself, under the tag "StoredTE".
+If the owner of that Tile Entity registers a `FixTypes.BLOCK_ENTITY` fixer to run on that Tile Entity, the DataFixer system is not aware that your dolly item has a Tile Entity tag stored inside of itself and thus the dolly is left with an invalid tag inside of it.
+This issue would be resolved in the following manner:
+
+```Java
+IDataWalker dollyWalker = new IDataWalker() {
+	@Override
+	public NBTTagCompound process(IDataFixer fixer, NBTTagCompound compound, int versionIn) {
+		// hasKey checks omitted here, get the stored TE
```
LemADEC: Should we assume the process is called systematically on all items and it's up to mod author to verify that this ItemStack is its own?
Where's the version coming from if it's applied systematically?

docs/datastorage/datafixers.md
williewillus: no, FMLMissingMappingsEvent is a completely orthogonal mechanism that rewrites entries in the FML ID maps, instead of fixing the world data. So the old name would point to the new item in the registry, making the old name still usable.

docs/datastorage/datafixers.md
williewillus: Saving of versions is completely handled by forge/vanilla. Each mod+vanilla has its own data version that is saved in the chunk, with the default at -1.

docs/datastorage/datafixers.md
williewillus: walkers will always run all the time. Fixers are only run if the target object is older than the fixer's declared version, so just by running in the fixer you know that you are outdated.

docs/datastorage/datafixers.md
williewillus: yes, you have to check the id tags in the stack to make sure it's your item

not sure what your second question means, but refer to my earlier comment

docs/datastorage/datafixers.md
LemADEC: that's good to know. That doesn't really answer my question through.

docs/datastorage/datafixers.md
LemADEC: sounds good, could we clarify that in the page?

docs/datastorage/datafixers.md
LemADEC: ok, that makes more sense now. I guess that means I need to re-implement the core system if I save subworlds or schematics?

docs/datastorage/datafixers.md
LemADEC: So the version is the one attached to your mod when the original data was saved, as found in the chunk NBT? in which case, register method will actually attach the walker instance to the calling mod instance, so later on the version parameter is the one matching that specific mod.
Which means one mod cannot fix/patch another mod that way?

docs/datastorage/datafixers.md
williewillus: I mean, you can *probably* use a walker to get to it somehow, if your "subworlds" and "schematics" are nested in the NBT of any of the existing fix types

docs/datastorage/datafixers.md
williewillus: ah, sorry, I misread. it's just a typo, the `version` parameter should just be passed on to the fixer

docs/datastorage/datafixers.md
williewillus: I'm not sure what you're asking but yes in general mods aren't aware of the version of other mods or vanilla.

Barteks2x:
So I'm finally back at the point where I can work on DataFixers, and I completely forgot what it was that I wanted to do...

williewillus:
@Barteks2x lol, I think on IRC you and Lex were talking about something about the order mod fixers/walkers run relative to vanilla? 

Or how the system could get confusing if you move some structure in a fixer and need to walk it at the same time. I don't remember the exact details :P

Barteks2x:
I think my issue was when you have a mod fixer/walker that depends on the vanilla NBT structure that could have changed. Honestly not sure how big of an issue that's going to be. Vanilla fixers just deal with vanilla version. Mod fixers need to deal with mod version *and* vanilla version. And if you jump for example 2 vanilla versions in one update, the mod fixer will probably get the NBT from that old vanilla version, when in fact it may expect just one version before. And there are some issues with the comments in the code being misleading.

williewillus:
idk how datafixers work in 1.13 now so nevermind about this

PR #150, opened 2018-02-07
Add section about persisting Chunk and TileEntity capabilities

JamiesWhiteShirt:


docs/datastorage/capabilities.md
```diff
@@ -129,8 +129,26 @@ private static class Factory implements Callable<IExampleCapability> {
...
+
+`ItemStackHandler`, commonly used for inventories in TileEntities, has an overridable method `void onContentsChanged(int slot)` designed to be used to mark the TileEntity as dirty.
+
+```Java
+public class MyTileEntity extends TileEntity {
+
+  private final IItemHandler inventory = new ItemStackHandler(...) {
+    @Override
+    protected void onContentsChanged(int slot) {
+      markDirty();
```
tterrag1098: To be technically correct, this should call super. Other than that looks fine.

PR #153, opened 2018-03-16
Merge my extended state doc into existing one

williewillus:
This merges the second half of #152 into the existing extended state article (replacing the usage example that was already there). I found the first half pretty good so I'm keeping it over my first half.

### DIFF
--- a/docs/models/advanced/extended-blockstates.md
+++ b/docs/models/advanced/extended-blockstates.md
@@ -30,13 +30,7 @@ public BlockStateContainer createBlockState() {
 }
 ```
 
-Similarly to listed properties, you will also need to set a default value for your unlisted property.
-
-```Java
-setDefaultState(((IExtendedBlockState)getDefaultState())
-	.withProperty(UNLISTED_PROP, val)
-	.withProperty(LISTED_PROP, val)
-```
+Note that you do not need to set default values for your unlisted properties. 
 
 
 Filling Extended States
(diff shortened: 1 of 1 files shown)


docs/models/advanced/extended-blockstates.md
```diff
@@ -16,43 +16,78 @@ Why use extended blockstates at all? Why not simply pass an `IBlockAccess` and `
...
+	.withProperty(LISTED_PROP, val)
+```
+
+
+Filling Extended States
+----------------------------
+
+Before an `IBlockState` is passed to an `IBakedModel`, it will always have `Block.getExtendedState` called on it first. In this method, you will give all your unlisted properties values. Assuming you registered at least one unlisted property in the previous section, the `IBlockState` parameter can be safely casted to `IExtendedBlockState`, which has a `withProperty` method for unlisted properties analogous to its listed property cousin. Here, you can query whatever you want from the World, the Tile Entity, etc. (with appropriate safety checks, of course) and insert it into the extended blockstate.
+
+!!! Warning
```
tterrag1098: I think this should be lowercase

docs/models/advanced/extended-blockstates.md
```diff
@@ -16,43 +16,78 @@ Why use extended blockstates at all? Why not simply pass an `IBlockAccess` and `
...
+
+Example:
+```Java
+@Override
+public BlockStateContainer createBlockState() {
+	return new BlockStateContainer.Builder(this).add(LISTED_PROP).add(UNLISTED_PROP).build();
+}
+```
+
+Similarly to listed properties, you will also need to set a default value for your unlisted property.
```
bs2609: This shouldn't be required?

williewillus:
addressed (finally)

PR #158, opened 2018-04-17
Add Dependency Management page

JamiesWhiteShirt:
Based on what was introduced in this PR https://github.com/MinecraftForge/MinecraftForge/pull/4841. This is mostly about the dependency extraction part, but also adds a necessary introduction to repositories.

Requesting review from @LexManos on this one. A specific thing I could use feedback on is the name "The mod repository". That is what I have called the repository where contained dependencies are archived. Is there perhaps a better name? I also deliberately chose to not document the deprecated behavior, namely placing contained dependencies in the jar root and not specifying the Maven-Artifact manifest attribute.

### DIFF
--- a/docs/gettingstarted/dependencymanagement.md
+++ b/docs/gettingstarted/dependencymanagement.md
@@ -1,0 +1,23 @@
+Dependency Management
+=====================
+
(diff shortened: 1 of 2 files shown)


tterrag1098:
Assigning lex as I know next to nothing about this system.

LexManos:
Thats the gist of it I guess. You don't really go into explaing the classpath or why its a good idea. But sure.

PR #165, opened 2018-05-04
Add internationalization and localization page

JamiesWhiteShirt:
I see lots of confusion about localization, and often see it used incorrectly. Documenting it should hopefully help.

### DIFF
--- a/docs/concepts/internationalization.md
+++ b/docs/concepts/internationalization.md
@@ -35,7 +35,7 @@ Unlike registry names, unlocalized names are not namespaced. It is therefore con
 !!! note
     The only purpose of unlocalized names is internationalization. Do not use them for logic. Use registry names instead.
 
-    A common pattern is using `getUnlocalizedName().subString(5)` to assign registry names. This is fragile and uses the unlocalized name for logic, which is considered bad practice. It could instead be reversed so that the registry name is set first, then the unlocalized name is set according to the registry name `MODID + "." + getRegistryName().getResourcePath()`.
+    A common pattern is using `getUnlocalizedName().substring(5)` to assign registry names. This is fragile and uses the unlocalized name for logic, which is considered bad practice. It could instead be reversed so that the registry name is set first, then the unlocalized name is set according to the registry name `MODID + "." + getRegistryName().getResourcePath()`.
 
 Localization methods
 --------------------
--- a/docs/concepts/internationalization.md
+++ b/docs/concepts/internationalization.md
@@ -1,0 +1,57 @@
+Internationalization and localization
+=====================================
+
+Internationalization, i18n for short, is a way of designing code so that it requires no changes to be adapted for various languages. Localization is the process of adapting displayed text to the user's language.
+
+I18n is implemented using _translation keys_. A translation key is a string that identifies a piece of displayable text in no specific language. For example, `tile.dirt.name` is the language key referring to the name of the Dirt block. This way, displayable text may be referenced with no concern for a specific language. The code requires no changes to be adapted in a new language.
+
+Localization will happen in the game's locale. In a Minecraft client the locale is specified by the language settings. On a dedicated server, the only supported locale is en_US.
+
+Language files
+--------------
+
+Language files are located by `assets/domain/lang/locale.lang` (e.g. the US English translation for `examplemod` would be `assets/examplemod/lang/en_us.lang`). Resource pack format 3 requires the locale name to be lowercased. The file format is simply lines of key-value pairs. Lines starting with `#` are treated as comments.
+
+```properties
+# items
+item.examplemod.example_item.name=Example Item Name
+
+# blocks
+tile.examplemod.example_block.name=Example Block Name
+
+# commands
+commands.examplemod.examplecommand.usage=/example <value>
+```
+
+A language file may contain escape sequences if the file starts with the comment `#PARSE_ESCAPES`.
+
+Localization methods
+--------------------
+
+!!! warning
+    A common issue is having the server localize for clients. The server can only localize in its own locale, which does not necessarily match the locale of connected clients.
+    
+    To respect the language settings of clients, the server should have clients localize text in their own locale using TextComponentTranslation or other methods preserving the language neutral translation keys.
+
+### `net.minecraft.client.resources.I18n` (client only)
+
+**This I18n class can only be found on a Minecraft client!** It is intended to be used by code that only runs on the client. Attempts to use this on a server will throw exceptions and crash.
+
+- `format(String, Object...)` localizes in the client's locale with formatting. The first parameter is a language key, and the rest are used for `String.format(String, Object...)`.
+
+### `net.minecraft.util.text.translation.I18n` (deprecated)
+
+**This class is deprecated and should often be avoided.** It is intended to be used by code that runs on the logical server. Because localization should rarely happen on the server, other alternatives should be considered.
+
+- `translateToLocal(String)` localizes in the game's locale without formatting. The parameter is a language key.
+- `translateToLocalFormatted(String, Object...)` localizes in the game's locale with formatting. The first parameter is a language key, and the rest are used for `String.format(String, Object...)`.
+
+### `TextComponentTranslation`
+
+`TextComponentTranslation` is an `ITextComponent` that is localized and formatted lazily. It is very useful when sending messages to players because it will be automatically localized in their own locale.
+
+The first parameter of the `TextComponentTranslation(String, Object...)` constructor is a language key, and the rest are used for formatting. The only supported format specifiers are `%s` or `%1$s`. Formatting arguments may be other `ITextComponent`s that will be inserted into the resulting formatted text with all their attributes preserved.
+
+### `TextComponentHelper`
+
+- `createComponentTranslation(ICommandSender, String, Object...)` creates a localized and formatted `ITextComponent` depending on a command sender. The localization and formatting is done eagerly if the command sender is a vanilla client. If not, the localization and formatting is done lazily with a `TextComponentTranslation`. This is only useful if the server should allow vanilla clients to connect.
--- a/docs/concepts/internationalization.md
+++ b/docs/concepts/internationalization.md
@@ -10,7 +10,7 @@ Localization will happen in the game's locale. In a Minecraft client the locale 
 Language files
 --------------
 
-Language files are located by `assets/domain/lang/[locale].lang` (e.g. the US English translation for `examplemod` would be `assets/examplemod/lang/en_us.lang`). Resource pack format 3 requires the locale name to be lowercased. The file format is simply lines of key-value pairs separated by `=`. Lines starting with `#` are ignored. Lines without a separator are ignored. The file must be encoded in UTF-8.
+Language files are located by `assets/[namespace]/lang/[locale].lang` (e.g. the US English translation for `examplemod` would be `assets/examplemod/lang/en_us.lang`). Resource pack format 3 requires the locale name to be lowercased. The file format is simply lines of key-value pairs separated by `=`. Lines starting with `#` are ignored. Lines without a separator are ignored. The file must be encoded in UTF-8.
 
 ```properties
 # items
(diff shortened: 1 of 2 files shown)


docs/concepts/internationalization.md
```diff
@@ -0,0 +1,57 @@
...
+**This class is deprecated and should often be avoided.** It is intended to be used by code that runs on the logical server. Because localization should rarely happen on the server, other alternatives should be considered.
+
+- `translateToLocal(String)` localizes in the game's locale without formatting. The parameter is a language key.
+- `translateToLocalFormatted(String, Object...)` localizes in the game's locale with formatting. The first parameter is a language key, and the rest are used for `String.format(String, Object...)`.
+
+### `TextComponentTranslation`
+
+`TextComponentTranslation` is an `ITextComponent` that is localized and formatted lazily. It is very useful when sending messages to players because it will be automatically localized in their own locale.
+
+The first parameter of the `TextComponentTranslation(String, Object...)` constructor is a language key, and the rest are used for formatting. The only supported format specifiers are `%s` or `%1$s`. Formatting arguments may be other `ITextComponent`s that will be inserted into the resulting formatted text with all their attributes preserved.
```
williewillus: maybe specify that the positional arg can be any number? e.g. `%2$s`, `%3$s`, `%<n>$s`
In case people weren't familiar with that notation for String.format (I wasn't until just now)

LexManos:
Anjy chance of adding a note about NOT using unlocalized names as registry keys. Something along the lines of "You're a freaking moron if you use unlocalize.substring seriously stop fucking doing this" but more PC?

JamiesWhiteShirt:
I very much dislike the practice of `getUnlocalizedName().substring(5)` myself. Unfortunately the ones who use it are not the ones who read the docs. Anyway, that reminded me of the lack of namespacing in unlocalized names so I managed to squeeze it in without looking too out of place.

docs/concepts/internationalization.md
```diff
@@ -0,0 +1,69 @@
...
+item.examplemod.example_item.name=Example Item Name
+
+# blocks
+tile.examplemod.example_block.name=Example Block Name
+
+# commands
+commands.examplemod.examplecommand.usage=/example <value>
+```
+
+A language file may contain escape sequences if the file starts with the comment `#PARSE_ESCAPES`.
```
PaleoCrafter: The comment turns on full blown properties syntax, not just escape sequences. The MC lang format isn't a subset of properties syntax (namely that files containing `:` in keys have different semantics in properties format, as the colon is a separator there), so I think elaborating a bit more on what this comment entails would be great.

docs/concepts/internationalization.md
```diff
@@ -0,0 +1,72 @@
...
+Internationalization, i18n for short, is a way of designing code so that it requires no changes to be adapted for various languages. Localization is the process of adapting displayed text to the user's language.
+
+I18n is implemented using _translation keys_. A translation key is a string that identifies a piece of displayable text in no specific language. For example, `tile.dirt.name` is the language key referring to the name of the Dirt block. This way, displayable text may be referenced with no concern for a specific language. The code requires no changes to be adapted in a new language.
+
+Localization will happen in the game's locale. In a Minecraft client the locale is specified by the language settings. On a dedicated server, the only supported locale is en_US.
+
+Language files
+--------------
+
+Language files are located by `assets/domain/lang/locale.lang` (e.g. the US English translation for `examplemod` would be `assets/examplemod/lang/en_us.lang`). Resource pack format 3 requires the locale name to be lowercased. The file format is simply lines of key-value pairs encoded in UTF-8. Lines starting with `#` are treated as comments.
```
3TUSK: Two questions:
1. All supported locales are defined in `pack.mcmeta` of Minecraft default resourcepack. [Here](https://minecraft.gamepedia.com/Language#Available_languages) is a (probably incomplete) list of valid locale codes. Should it be mentioned here?
2. The `*.lang` format only uses `=` to separate keys and values, and by default, ignores any invalid lines (i.e. lines that do not contains `=`), in addition to ignoring `#` comments. Should it be mentioned here?

docs/concepts/internationalization.md
```diff
@@ -0,0 +1,72 @@
...
+
+I18n is implemented using _translation keys_. A translation key is a string that identifies a piece of displayable text in no specific language. For example, `tile.dirt.name` is the language key referring to the name of the Dirt block. This way, displayable text may be referenced with no concern for a specific language. The code requires no changes to be adapted in a new language.
+
+Localization will happen in the game's locale. In a Minecraft client the locale is specified by the language settings. On a dedicated server, the only supported locale is en_US. A list of available locales can be found on the [Minecraft Wiki](https://minecraft.gamepedia.com/Language#Available_languages).
+
+Language files
+--------------
+
+Language files are located by `assets/domain/lang/locale.lang` (e.g. the US English translation for `examplemod` would be `assets/examplemod/lang/en_us.lang`). Resource pack format 3 requires the locale name to be lowercased. The file format is simply lines of key-value pairs separated by `=`. Lines starting with `#` are ignored. Lines without a separator are ignored. The file must be encoded in UTF-8.
```
tterrag1098: Maybe `[locale].lang` to avoid confusion?

docs/concepts/internationalization.md
```diff
@@ -0,0 +1,72 @@
...
+
+### `net.minecraft.client.resources.I18n` (client only)
+
+**This I18n class can only be found on a Minecraft client!** It is intended to be used by code that only runs on the client. Attempts to use this on a server will throw exceptions and crash.
+
+- `format(String, Object...)` localizes in the client's locale with formatting. The first parameter is a language key, and the rest are used for `String.format(String, Object...)`.
+
+### `net.minecraft.util.text.translation.I18n` (deprecated)
+
+**This class is deprecated and should often be avoided.** It is intended to be used by code that runs on the logical server. Because localization should rarely happen on the server, other alternatives should be considered.
```
tterrag1098: It should *always* be avoided.

docs/concepts/internationalization.md
```diff
@@ -0,0 +1,72 @@
...
+
+### `TextComponentTranslation`
+
+`TextComponentTranslation` is an `ITextComponent` that is localized and formatted lazily. It is very useful when sending messages to players because it will be automatically localized in their own locale.
+
+The first parameter of the `TextComponentTranslation(String, Object...)` constructor is a language key, and the rest are used for formatting. The only supported format specifiers are `%s` and `%1$s`, `%2$s`, `%3$s` etc. Formatting arguments may be other `ITextComponent`s that will be inserted into the resulting formatted text with all their attributes preserved.
+
+### `TextComponentHelper`
+
+- `createComponentTranslation(ICommandSender, String, Object...)` creates a localized and formatted `ITextComponent` depending on a command sender. The localization and formatting is done eagerly if the command sender is a vanilla client. If not, the localization and formatting is done lazily with a `TextComponentTranslation`. This is only useful if the server should allow vanilla clients to connect.
```
tterrag1098: Don't you mean if the *receiver* of the message is a vanilla client? `sender` is the name of the parameter but this corresponds to ICommandSender.

JamiesWhiteShirt:
Makes sense.

JamiesWhiteShirt:
@tterrag1098 psst

docs/concepts/internationalization.md
```diff
@@ -0,0 +1,72 @@
...
+
+Including the comment `#PARSE_ESCAPES` anywhere in the file will enable the slightly different and more complex Java properties file format. This format is required to support multiline values.
+
+!!! note
+    Enabling `#PARSE_ESCAPES` enables changes various aspects of the parsing. Among other things, the `:` character will be treated as a key-value separator. It must either be excluded from translation keys or escaped using `\:`.
+
+Unlocalized names
+-----------------
+
+Blocks, Items and other Minecraft constructs have unlocalized names. Unlocalized names are simply translation keys that are used when displaying their names.
```
thiakil: Translation keys are kinda explained in the first section, 2nd paragraph. Maybe something like "In Minecraft (MCP) the translation keys are called `unlocalised names`"? or even merging the 2 paragraphs

docs/concepts/internationalization.md
```diff
@@ -0,0 +1,72 @@
...
+
+Unlike registry names, unlocalized names are not namespaced. It is therefore conventional to have unlocalized names prefixed with your modid (e.g. `examplemod.example_item`) to avoid conflicts. Otherwise, in the event of a conflict, the localization of one item will override the other.
```
thiakil: conventional -> highly recommended

Many mods don't do it when they should

docs/concepts/internationalization.md
```diff
@@ -0,0 +1,72 @@
...
+
+    A common pattern is using `getUnlocalizedName().substring(5)` to assign registry names. This is fragile and uses the unlocalized name for logic, which is considered bad practice. It could instead be reversed so that the registry name is set first, then the unlocalized name is set according to the registry name `MODID + "." + getRegistryName().getResourcePath()`.
+
+Localization methods
+--------------------
+
+!!! warning
+    A common issue is having the server localize for clients. The server can only localize in its own locale, which does not necessarily match the locale of connected clients.
+    
+    To respect the language settings of clients, the server should have clients localize text in their own locale using TextComponentTranslation or other methods preserving the language neutral translation keys.
```
thiakil: `TextComponentTranslation` should be in code tags

docs/concepts/internationalization.md
JamiesWhiteShirt: I have described an "unlocalized name" as a translation key for the names of `Blocks, Items and other Minecraft constructs` only, and not as a term to be used interchangably with "translation key". In 1.13 the term "unlocalized name" will be gone.

Some might disagree on this, so I need some people's opinions.

docs/concepts/internationalization.md
JamiesWhiteShirt: All right, will change that.

docs/concepts/internationalization.md
JamiesWhiteShirt: Yup

docs/concepts/internationalization.md
thiakil: > and not as a term to be used interchangably with "translation key"

True, not all translation keys are unlocalised names (it was early when I wrote that). It's still a little weird to be mentioning them as a different thing just because they're used in one (type of) place. `and other Minecraft constructs` is also super vague.

Perhaps it needs to mention the `.name` addition some things get?

> In 1.13 the term "unlocalized name" will be gone.

IMO that's worth mentioning in the doc given how close they are to release

JamiesWhiteShirt:
All right, another round. The "unlocalized name" -> "translation key" has already been commited to MCP mapping snapshots, so I had to add a note. I could definitely use some feedback on how that is/could be handled on this page.

Or I could just wait some more...

docs/concepts/internationalization.md
```diff
@@ -0,0 +1,75 @@
...
+
+I18n is implemented using _translation keys_. A translation key is a string that identifies a piece of displayable text in no specific language. For example, `tile.dirt.name` is the translation key referring to the name of the Dirt block. This way, displayable text may be referenced with no concern for a specific language. The code requires no changes to be adapted in a new language.
+
+Localization will happen in the game's locale. In a Minecraft client the locale is specified by the language settings. On a dedicated server, the only supported locale is en_US. A list of available locales can be found on the [Minecraft Wiki](https://minecraft.gamepedia.com/Language#Available_languages).
+
+Language files
+--------------
+
+Language files are located by `assets/domain/lang/[locale].lang` (e.g. the US English translation for `examplemod` would be `assets/examplemod/lang/en_us.lang`). Resource pack format 3 requires the locale name to be lowercased. The file format is simply lines of key-value pairs separated by `=`. Lines starting with `#` are ignored. Lines without a separator are ignored. The file must be encoded in UTF-8.
```
thiakil: Perhaps put `domain` in [] like `locale`?

docs/concepts/internationalization.md
```diff
@@ -0,0 +1,75 @@
...
+!!! warning
+    A common issue is having the server localize for clients. The server can only localize in its own locale, which does not necessarily match the locale of connected clients.
+    
+    To respect the language settings of clients, the server should have clients localize text in their own locale using `TextComponentTranslation` or other methods preserving the language neutral translation keys.
+
+### `net.minecraft.client.resources.I18n` (client only)
+
+**This I18n class can only be found on a Minecraft client!** It is intended to be used by code that only runs on the client. Attempts to use this on a server will throw exceptions and crash.
+
+- `format(String, Object...)` localizes in the client's locale with formatting. The first parameter is a language key, and the rest are formatting arguments for `String.format(String, Object...)`.
```
thiakil: This section and below uses "language key" instead of "translation key"

docs/concepts/internationalization.md
bs2609: On that note: `domain` -> `namespace` (https://github.com/ModCoderPack/MCPBot-Issues/issues/526)

JamiesWhiteShirt:
Ready for another round.

docs/concepts/internationalization.md
```diff
@@ -0,0 +1,75 @@
+Internationalization and localization
+=====================================
+
+Internationalization, i18n for short, is a way of designing code so that it requires no changes to be adapted for various languages. Localization is the process of adapting displayed text to the user's language.
+
+I18n is implemented using _translation keys_. A translation key is a string that identifies a piece of displayable text in no specific language. For example, `tile.dirt.name` is the translation key referring to the name of the Dirt block. This way, displayable text may be referenced with no concern for a specific language. The code requires no changes to be adapted in a new language.
+
+Localization will happen in the game's locale. In a Minecraft client the locale is specified by the language settings. On a dedicated server, the only supported locale is en_US. A list of available locales can be found on the [Minecraft Wiki](https://minecraft.gamepedia.com/Language#Available_languages).
```
thiakil: If we're doing this mainly focused on v3+ `en_US` should be lowercased

docs/concepts/internationalization.md
```diff
@@ -0,0 +1,75 @@
...
+
+Language files
+--------------
+
+Language files are located by `assets/[namespace]/lang/[locale].lang` (e.g. the US English translation for `examplemod` would be `assets/examplemod/lang/en_us.lang`). Resource pack format 3 requires the locale name to be lowercased. The file format is simply lines of key-value pairs separated by `=`. Lines starting with `#` are ignored. Lines without a separator are ignored. The file must be encoded in UTF-8.
```
thiakil: Should we say "Resource pack format 3 (specified in [pack.mcmeta](https://minecraft.gamepedia.com/Tutorials/Creating_a_resource_pack#pack.mcmeta))" ?

docs/concepts/internationalization.md
```diff
@@ -0,0 +1,75 @@
...
+# commands
+commands.examplemod.examplecommand.usage=/example <value>
+```
+
+Including the comment `#PARSE_ESCAPES` anywhere in the file will enable the slightly different and more complex Java properties file format. This format is required to support multiline values.
+
+!!! note
+    Enabling `#PARSE_ESCAPES` changes various aspects of the parsing. Among other things, the `:` character will be treated as a key-value separator. It must either be excluded from translation keys or escaped using `\:`.
+
+Usage with Blocks and Items
```
thiakil: Should this section include a brief note about the game looking for a key with `.name` appended by default?

docs/concepts/internationalization.md
bs2609: `en_US` is the actual name of the locale, though - this isn't talking about a lang file name.

docs/concepts/internationalization.md
```diff
@@ -0,0 +1,75 @@
...
+---------------------------
+
+!!! note
+    As of July 14th 2018, the term "unlocalized name" has been replaced by "translation key" in MCP.
+
+Block, Item and a few other Minecraft classes have built-in translation keys used to display their names. They are set by calling the corresponding `setTranslationKey(String)` / `setUnlocalizedName(String)` method, or by overriding the corresponding `getTranslationKey` / `getUnlocalizedName` method.
```
tterrag1098: I would write this only using `getTranslationKey` and change the above note to mention the old naming scheme. This will be more future-proof.

JamiesWhiteShirt:
@tterrag1098 Just letting you know the comments are addressed.

PR #168, opened 2018-06-04
RegistryEvent.Register handlers must be static.

kylev:
I've seen this turn into a bug around forums.

### DIFF
--- a/docs/concepts/registries.md
+++ b/docs/concepts/registries.md
@@ -12,7 +12,7 @@ The recommended way to register things is through the `RegistryEvent`s. These [e
 
 ```java
 @SubscribeEvent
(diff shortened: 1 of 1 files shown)


TehNut:
They do not *have* to be static. They only need to be static if the class is registered to `MinecraftForge.EVENT_BUS` or if `@Mod.EventBusSubscriber` is used. Registering an instance the old way does not require static.

This is already covered on the event page.

LexManos:
> They do not have to be static. They only need to be static ~~if the class is registered to MinecraftForge.EVENT_BUS or~~ if @Mod.EventBusSubscriber is used. Registering an instance the old way does not require static.

PR #175, opened 2018-06-22
Fixes #169.

### LINKED ISSUE #169: Entity registration
In 1.12.2, entity registration is a bit more complicated than registering blocks, items, etc. which are singletons.
I suggest to update the registries concept page to cover entities registration in a new section.

JTK222:
Superfences should add support nested admonitions.
But I am not sure if the Forge Design supports them properly.

### DIFF
--- a/docs/concepts/registries.md
+++ b/docs/concepts/registries.md
@@ -24,7 +24,7 @@ The order in which `RegistryEvent.Register` events fire is alphabetically, with 
 There is another, older way of registering objects into registries, using `GameRegistry.register`. Anytime something suggests using this method, it should be replaced with an event handler for the appropriate registry event. This method simply finds the registry corresponding to an `IForgeRegistryEntry` with `IForgeRegistryEntry::getRegistryType`, and then registers the object to the registry. There is also a convenience overload that takes an `IForgeRegistryEntry` and a `ResourceLocation`, which is equivalent to calling `IForgeRegistryEntry::setRegistryName`, followed by a `GameRegistry.register` call.
 
 !!! information
-  Registering an Entity might be a little bit confusing at first as it doesn't use the `Entity` class, but an `EntityEntry`. These are created by making use of `EntityEntryBuilder`.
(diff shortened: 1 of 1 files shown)


docs/concepts/registries.md
```diff
@@ -23,6 +23,11 @@ The order in which `RegistryEvent.Register` events fire is alphabetically, with
 
 There is another, older way of registering objects into registries, using `GameRegistry.register`. Anytime something suggests using this method, it should be replaced with an event handler for the appropriate registry event. This method simply finds the registry corresponding to an `IForgeRegistryEntry` with `IForgeRegistryEntry::getRegistryType`, and then registers the object to the registry. There is also a convenience overload that takes an `IForgeRegistryEntry` and a `ResourceLocation`, which is equivalent to calling `IForgeRegistryEntry::setRegistryName`, followed by a `GameRegistry.register` call.
 
+!!! information
+  Registering Entities might be a little bit confusing at first as it doesn't use the Entity Class, but an `EntityEntry`. You can create these using the `EntityEntryBuilder`.
```
TehNut: Entities -> an `Entity`

Entity Class should be lowercase. Or at least use `Entity` class.

>You can create these using the `EntityEntryBuilder`.

These are created by making use of `EntityEntryBuilder`.

docs/concepts/registries.md
```diff
@@ -23,6 +23,11 @@ The order in which `RegistryEvent.Register` events fire is alphabetically, with
...
+  !!! note
+    `EntityEntryBuilder#id()` equals the `setRegistryName()` method from other Objects, with the difference that it also takes a mod internal int ID. A simple counter during registration is enough as this ID is only used for networking.
```
TehNut: >`EntityEntryBuilder#id()` equals the `setRegistryName()` method from other Objects

`EntityEntryBuilder#id()` is equivalent to the `setRegistryName()` method from `IForgeRegistryEntry`.

JTK222:
Push @tterrag1098

JTK222:
Removed the nested admonitions as it seems not to work with the forge theme

PR #189, opened 2018-09-25
Add 'static' to the lootLoad event listener method

shelbykauth:
This does not work without the 'static' keyword.  As per http://www.minecraftforge.net/forum/topic/66581-solved-cant-get-loottable-loading-event-to-fire/

I know it's just a single word, but it will save a lot of time and energy for a lot of coders.


williewillus:
But it does work, they're two different ways of registering event handlers

PR #196, opened 2018-10-16
Grammar update

munchicken:
Removed unnecessary word "will" from "Note" section.


tterrag1098:
Please submit #194 and #195 as part of this PR, there is no need to separate them.

Issue #167, opened 2018-05-30
Documentation on Advancements


MinecraftPlaye:
Some documentation about how advancements work within Forge and what different types and possibilities we have with them should be added.

PR #200, opened 2018-11-23
Adding Documentation for Advancements

LukeGi:
As suggested by #167, I am adding some documentation about how to create advancements and what options there are for them.

### DIFF
--- a/docs/advancements/basics.md
+++ b/docs/advancements/basics.md
@@ -1,0 +1,42 @@
+Advancements Basics
+===================
+
+The Home of Advancements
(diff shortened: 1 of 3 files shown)


tterrag1098:
Is this necessary? The minecraft wiki has pages describing this process already. If documentation were to be done anywhere I think it should be on custom criteria and triggers.

tterrag1098:
I never like to duplicate information. The Minecraft wiki is more consistently updated and is very thorough. These docs only target mods.

PR #208, opened 2018-12-03
Fixes #207

### LINKED ISSUE #207: OreDictionary entries are registered too early due to outdated information
The current documentation tells that OreDictionary entries should be registered during preInit, however the registry events fire *after* preInit, which means registering entries to the OreDictionary will not work at that time.

Current version of the Doc at [OreDictionary](https://mcforge.readthedocs.io/en/latest/utilities/oredictionary/ "ReadTheDocs"):
> Add entries to the OreDictionary during the FMLPreInitializationEvent phase, after initializing the blocks and items that you will register.

strikerrocker:


tterrag1098:
I would say this is not totally accurate. One should probably just register things to the ore dictionary directly inside the registry events. Probably worth mentioning.

docs/utilities/oredictionary.md
```diff
@@ -61,7 +61,7 @@ Another use of the OreDictionary in crafting is the [WILDCARD_VALUE](#wildcard_v
 Registering Items to the OreDictionary
 --------------------------------------
 
-Add entries to the OreDictionary during the `FMLPreInitializationEvent` phase, after initializing the blocks and items that you will register.
+Add entries to the OreDictionary during the registry events, after initializing the blocks and items that you will register.
```
MajorTuvok: Shouldn't it be "that you registered" then?

PR #231, opened 2019-08-03
Let's put an end to "Where's setupDecompWorkspace?" in 1.13+

DaemonUmbra:


Niko-sk2x:
This doesn't explain gradle daemon breaking things and how to get around those issues (and it can't be disabled for IDE integration).

PR #240, opened 2019-12-17
Add explaination about side-only mods and network version predicates

JoJoDeveloping:
Added documentation for the displaytest extensionpoint, which needs to be overwritten if your mod is side-only.

Also added explanation about network version predicates with SimpleImpl, and how they are affected/called if the mod is not present on the other side.

### DIFF
--- a/docs/concepts/sides.md
+++ b/docs/concepts/sides.md
@@ -58,3 +58,16 @@ Whenever you want to send information from one logical side to another, you must
 This is actually very commonly inadvertently done through static fields. Since the logical client and logical server share the same JVM in a single player scenario, both threads writing to and reading from static fields will cause all sorts of race conditions and the classical issues associated with threading.
 
 This mistake can also be made explicitly by accessing physical client-only classes such as `Minecraft` from common code that runs or can run on the logical server. This mistake is easy to miss for beginners, who debug in a physical client. The code will work there, but will immediately crash on a physical server.
+
+
+Writing One-Sided Mods
+----------------------
+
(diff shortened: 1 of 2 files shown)


docs/concepts/sides.md
```diff
@@ -58,3 +58,16 @@ Whenever you want to send information from one logical side to another, you must
 This is actually very commonly inadvertently done through static fields. Since the logical client and logical server share the same JVM in a single player scenario, both threads writing to and reading from static fields will cause all sorts of race conditions and the classical issues associated with threading.
 
 This mistake can also be made explicitly by accessing physical client-only classes such as `Minecraft` from common code that runs or can run on the logical server. This mistake is easy to miss for beginners, who debug in a physical client. The code will work there, but will immediately crash on a physical server.
+
+
+Writing Side-Only Mods
```
tterrag1098: Perhaps "one sided" would be a better term here?

docs/concepts/sides.md
```diff
@@ -58,3 +58,16 @@ Whenever you want to send information from one logical side to another, you must
...
+----------------------
+
+In recent versions, Minecraft Forge has removed a "sidedness" attribute from the mods.toml. This means that your mods are expected to work whether they are loaded on the Client or on the Server. This means that for side-only mods, you would typically register your event handlers inside a `callWhenOn`, instead of directly calling the relevant registration methods in the constructor. If your mod is loaded on the wrong version, it should simply do nothing, listen to no events, and so on. A side-only mod by nature should not register blocks, items, ... since they can't be used on a server lacking that mod.
```
tterrag1098: `callWhenOn` is not documented anywhere else.

tterrag1098 (changes requested):

docs/concepts/sides.md
JoJoDeveloping: Yeah, that sounds better. Will use that.

docs/concepts/sides.md
JoJoDeveloping: It has javadoc..
Should I add it and `runWhenOn` to this documentation page while I'm at it?

tterrag1098 (approved):

PR #242, opened 2020-01-08
Updated Events section

Aeltumn:
I'm going to give every section a pass getting it up-to-date with 1.14 whenever I've got some time.

Here are some tiny improvements to the Events section, I've added some information and an example on how to use the properties of the Mod.EventBusSubscriber annotation to your advantage to register client-only events. This will hopefully aid people separating their client/common classes because they use the annotation instead.

### DIFF
--- a/docs/events/intro.md
+++ b/docs/events/intro.md
@@ -90,7 +90,7 @@ Many events have different variations of themselves, these can be different but 
 Mod Event Bus
 -------------
 
-The mod event bus is primarily meant for listening to lifecycle events in which mods should initialize. Many of these events are also ran in parallel so mods can be initialized at the same time. This does mean you can't directly execute code from other mods in these events, use the `InterModEnqueueEvent` and `InterModProcessEvent` events for that.
+The mod event bus is primarily meant for listening to lifecycle events in which mods should initialize. Many of these events are also ran in parallel so mods can be initialized at the same time. This does mean you can't directly execute code from other mods in these events, use the `InterModComms` system for that.
 
 These are the four most commonly used events that are called during mod initialization on the mod event bus:
 * FMLCommonSetupEvent
--- a/docs/events/intro.md
+++ b/docs/events/intro.md
@@ -90,9 +90,9 @@ Many events have different variations of themselves, these can be different but 
 Mod Event Bus
 -------------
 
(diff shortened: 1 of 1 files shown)


docs/events/intro.md
```diff
@@ -43,8 +43,21 @@ which must be registered like this: `MinecraftForge.EVENT_BUS.register(MyStaticF
...
+	@SubscribeEvent
+	public static void drawLast(RenderWorldLastEvent event) {
+		System.out.println("Drawing!");
+	}
+}
+```
+
 !!! note
-    This does not register an instance of the class; it registers the class itself (i.e. the event handling methods must be static).
+    This does not register an instance of the class; it registers the class itself (i.e. the event handling methods must be static.
```
tterrag1098: Don't remove that paren

tterrag1098 (changes requested):
This could use a bit more documentation on the mod bus, and which types of events use it.

docs/events/intro.md
```diff
@@ -84,3 +86,17 @@ Sub Events
 ----------
 
 Many events have different variations of themselves, these can be different but all based around one common factor (e.g. `PlayerEvent`) or can be an event that has multiple phases (e.g. `PotionBrewEvent`). Take note that if you listen to the parent event class, you will receive calls to your method for *all* subclasses.
+
+Mod Event Bus
+-------------
+
+The mod event bus is primarily meant for listening to lifecycle events in which mods should initialize. Many of these events are also ran in parallel so mods can be initialized at the same time. This does mean you can't directly execute code from other mods in these events, use the `InterModEnqueueEvent` and `InterModProcessEvent` events for that.
```
tterrag1098: The mod event bus is used for all events that are only called during the load process. This includes lifecycle events yes, but also registry events, and other misc load-time events like color handlers and model bake etc.

tterrag1098: Also perhaps mention DeferredWorkQueue

tterrag1098 (changes requested):

Aeltumn:
I'm not super familiar with the mod event bus, didn't have a clue which events where fired on which bus before I looked into it today, so the information might be a bit lacklustre.

tterrag1098 (approved):

PR #243, opened 2020-01-08
Update Rendering section

Aeltumn:
Updated the rendering category page about TileEntityItemStackRenderers as the name changed and the method where to set the TEISR moved to Item.Properties.


docs/rendering/teisr.md
```diff
@@ -1,21 +1,21 @@
...
-TileEntityItemStackRenderer is a method to use OpenGL to render on items.  This system is much simpler than the old TESRItemStack system, which required a TileEntity, and did not allow access to the ItemStack.
+ItemStackTileEntityRenderer is a method to use OpenGL to render on items. This system is much simpler than the old TESRItemStack system, which required a TileEntity, and did not allow access to the ItemStack.
 
-Using TileEntityItemStackRenderer
+Using ItemStackTileEntityRenderer
 --------------------------
 
-TileEntityItemStackRenderer allows you to render your item using `public void renderByItem(ItemStack itemStackIn)`.  
+!!! note
+    The ItemStackTileEntityRenderer is abbreviated as TEISR instead of ISTER as it used to be called TileEntityItemStackRenderer in previous versions.
```
tterrag1098: Just fix the abbreviation

tterrag1098 (changes requested):

docs/rendering/teisr.md
Aeltumn: I had kept it as TEISR because the method Item.Properties#setTEISR has the old abbreviation but it's indeed more logical to change it. I haven't submitted anything to the MCP bot before but I'll see if I can suggest changing the method name to setISTER.

tterrag1098 (approved):

PR #244, opened 2020-01-08
Updated Blocks section

Aeltumn:
I've redone the blocks category, it has quite a few changes.

Main differences:
- Explain how Block.Properties works
- Updated class names like ItemBlock to new one, BlockItem
- Added note about that you shouldn't not register blocks based on config
- Filled in the coming soons in the interaction section
- Reorganised interaction section as it had two different layouts
- Updated blockstates explaination to the best of my ability, haven't used blockstates much myself but I believe I updated it correctly looking at the Minecraft blocks using blockstates. The "actual states" part about having non-saved properties doesn't exist anymore, right? I know there is IModelData now which I think is the replacement. ('ll probably add a new page about IModelData eventually in either the Models or the Rendering category)
- Adding some information about the Flattening and encourage users to make more blocks instead of having too many blockstates

### DIFF
--- a/docs/blocks/blocks.md
+++ b/docs/blocks/blocks.md
@@ -8,17 +8,18 @@ Creating a Block
 
 ### Basic Blocks
 
-For simple blocks, which need no special functionality (think cobblestone, wood planks, etc.), a custom class is not necessary. By simply instantiating the `Block` class and calling some of the many setters, one can create many different types of blocks. For instance:
+For simple blocks, which need no special functionality (think cobblestone, wooden planks, etc.), a custom class is not necessary. You can create a block by instantiating the `Block` class with a `Block.Properties` object. This `Block.Properties` object can be made using `Block.Properties#create` and it can be customised by calling its methods. For instance:
 
-- `setHardness` - Controls the time it takes to break the block. It is an arbitrary value. For reference, stone has a hardness of 1.5, and dirt 0.5. If the block should be unbreakable, a convenience method `setBlockUnbreakable` is provided.
-- `setResistance` - Controls the explosion resistance of the block. This is separate from hardness, but `setHardness` will also set the resistance to 5 times the hardness value, if the resistance is any lower than this value.
-- `setSoundType` - Controls the sound the block makes when it is punched, broken, or placed. Requires a `SoundType` argument, see the [sounds][] page for more details.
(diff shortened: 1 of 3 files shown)


docs/blocks/blocks.md
```diff
@@ -8,17 +8,18 @@ Creating a Block
...
+- `sound` - Controls the sound the block makes when it is punched, broken, or placed. Requires a `SoundType` argument, see the [sounds][] page for more details.
+- `lightValue` - Controls the light emission of the block. Takes a value from zero to fifteen.
+- `slipperiness` - Controls how slippery the block is. For reference, ice has a slipperiness of 0.98.
 
-All these methods are *chainable* which means you can call them in series. See `Block#registerBlocks` for examples of this.
+All these methods are *chainable* which means you can call them in series. See the `Blocks` class for examples of this.
+
+!!! Note
+
+    Blocks have no setter for their `ItemGroup`. (formerly Creative Tab) This has been moved to the `BlockItem` and is now its responsibility. Furthermore, there is no setter for translation key. (this is generated based on registry name now)
```
tterrag1098: Remove the parenthetical, or just note it more briefly
```suggestion
    Blocks have no setter for their `ItemGroup`. (formerly Creative Tab) This has been moved to the `BlockItem` and is now its responsibility. Furthermore, there is no setter for translation key as it is now generated from the registry name.
```

docs/blocks/states.md
```diff
@@ -44,68 +44,37 @@ In your Block class, create static final `IProperty<>` objects for every propert
...
-!!! Note
-    Querying `world.getBlockState()` will give you the `IBlockState` representing only the metadata. Thus the returned `IBlockState` will not have data from `getActualState` filled in. If that matters to your code, make sure you call `getActualState`!
-
-[Extended Blockstates][]
-------------------------
-
-Extended blockstates are a way to pass arbitrary data into `IBakedModel`s during the rendering of a block. They are mainly used in the context of rendering, and therefore are documented in the Models section.
+Flattening
+----------
+As of Minecraft 1.13 metadata values have also been removed, instead of creating blocks with many blockstates to set its properties it is now preferred to simply make more blocks. Do you have a new wooden object that should have a variant for every wood type? In the past you'd have used blockstates for this, but now it is preferred to create separate blocks for each wood type.
```
tterrag1098, line 77: Mention the rule of thumb here, which is "if it has a different name, it should be a different block/item"

tterrag1098 (changes requested):

tterrag1098 (approved):

PR #245, opened 2020-01-13
Re-added 1.12 to dropdown list

Hexeption:


diesieben07:
No. [Only 1.14.4 and 1.15.1 are supported](https://www.minecraftforge.net/forum/topic/79304-112114-and-lts-system/).

LexManos:
This is the documents repo, there isn't really a reason to hide the links to the old data. We just don't need to update it anymore

Hexeption:
Like most documentations for most api and such keep an older version of the documentation i feel this should done here also, Yes 1.12 is legacy but it's still useful for some developers

tterrag1098:
The way the dropdown list works right now, having it show all past versions would be unmaintainable, since you must update all version across all branches. Perhaps a better solution would be links to legacy docs pages from the README or some other external source.

tterrag1098:
As noted above, this solution is not correct, so I'm closing this PR.

PR #253, opened 2020-01-18
Add doc for config system

qouteall:


DaemonUmbra (changes requested):
If you must use examples, try to chop it up and explain each bit

qouteall:
Added more explanation

docs/datastorage/config.md
```diff
@@ -11,20 +11,29 @@ public class MyConfig {
...
-    
+    //use defineEnum() to define enum type setting
+    //use defineInRange() to limit number range
+    //use translation() to specify translation key
+    //use worldRestart() to specify that world should reload when config changes
+    //You can also register nested config. Check the code of ForgeConfigSpec.Builder for more information
+  }
+  
+  // Read the config value
+  // Don't read it too early
```
Cadiboo, line 30: Client & Common configs are loaded after registry events and before common setup. Server configs are loaded when the world loads. You might want to mention this.

docs/datastorage/config.md
```diff
@@ -37,12 +46,10 @@ public class ExampleMod {
   public ExampleMod() {
     //...
     
-    // You can also register client only config or server only config
+    // Register the config here
+    // You can also register client only config or server only config by specifying ModConfig.Type.CLIENT or ModConfig.Type.SERVER
```
Cadiboo, line 50: It’s not exactly server only as it is synced to the client. (IMO a better name for Server is actually Common as common is pretty much the same as client but that’s an issue for Forge not here)

docs/datastorage/config.md
```diff
@@ -0,0 +1,48 @@
...
+  public final ForgeConfigSpec.BooleanValue setting;
+    
+  public MyConfig(ForgeConfigSpec.Builder builder) {
+    setting = builder
+      .comment("This is comment")
+      .define("setting", false);
+    }
+    
+  public static boolean getSetting() {
+      return INSTANCE.setting.get();
```
Cadiboo, line 32: You might want to have a section on performance and the advantages of “baking” your config values. While baking you can also perform additional operations, for example turning a list of strings into a set of BlockStates if that’s what the config entry is meant to represent.

Cadiboo (changes requested):

docs/datastorage/config.md
qouteall: What do you mean "bake"? Do you mean cache the config value?
Is there an event or callback so that I can update the cache when config updates?
(Actually, I am not familiar with the config system.)

Cadiboo: Yes, the event. [Here’s](https://cadiboo.github.io/tutorials/1.15.1/forge/3.3-config/) what I’ve got so far (very bare bones, I haven’t given this tutorial as much attention as I usually do), feel free to plagiarise. I’ve found it hard to explain the system and have also had to resort to using examples. I’ll go and change my constant name from “CLIENT” to “INSTANCE”, it’s a much better name. I’ll also get you an example of the transformation I mentioned in a sec

qouteall: It seems that your tutorial is better than this pr. I think that the config doc should be contributed by you-_-

Cadiboo: Uh oh that’s a lot of responsibility. I’ll do it soon:tm:. When I do, do you mind if I steal from your PR?

Issue #254, opened 2020-01-18
Possible category for legacy content


Aeltumn:
Considering the documentation for 1.12.x still exists and there are probably still people who are going to mod for the version: I don't see a problem with having a link somewhere to access the old docs as long as there's a nice big disclaimer that they shouldn't expect any support. I agree #245 wasn't a good solution so I'd suggest the following:

Having a dedicated category for information about legacy versions. Every bit of information that doesn't concern a supported version is banished to this category, this will cause the category to grow in size as new versions keep getting released but I don't think this would be too big of a problem.

This category would contain a main page that has a list of all previous versions with links to the documentation sites for those versions. This main page can have some text about how what the LTS version currently is and how no one should expect support for the other ones.

This section can also have more pages added similar to #216, pages that reword the update primers (just linking them could also do) and explain generally what changed between the versions.

_Note: With "every bit of information that doesn't concern a supported version is banished to this category" I mean that there would be a link to the old version of the documentation site in the main page, not that there would be a subcategory called 1.12.x with everything in there._

tterrag1098:
That sounds fine.

Issue #216, opened 2019-02-06
[Page Request] Updating from 1.12 to 1.13


Aeltumn:
Because of the major differences between 1.12 and 1.13 I feel it'd be helpful for mod authors if there was a page on the documentation to effectively and concretely explain how to update, with links to utilities to help conversion. (site for lang files to json, maybe scripts for auto-renaming files) At the very least a link to the [1.13 primer](https://gist.github.com/williewillus/353c872bcf1a6ace9921189f6100d09a) could really help those trying to update.

I'd glady write a draft if this is something that would fit in the documentation but I'd like to pitch the idea first and collect some opinions.

tterrag1098:
I've long asked @williewillus to incorporate his primers as part of the docs...

williewillus:
I don't like having updating guides in the docs because at some point they're no longer useful and someone has to remove it.

Stable, version-specific info should be there instead ,which I'm starting to do with my PRs (#213, #214 )

daniel-goossens:
I get that, however I still think the information is very important to provide. Maybe there could be a seperate section specifically called "Updating from legacy versions" where future updates could also be put, as in there's a page called 1.12 -> 1.13 and in the future there would be pages called 1.14 -> 1.15 etc. where the pages are written during the update and afterwards never touched again. In that way they are eventually not useful but their inclusion is clear, they don't have to be removed because they are by definition redundant for new users, future users wouldn't in any way get confused by it and noone would expect it to do more than telling them how to specifically update from one specfic version to another.

DarkGuardsman:
I would also like to see a porting from legacy section. Especially for a page linking to or including guides on porting. As I often forget that there is a primers and have to pester people for links. So a consistent place to find these links would be useful. Not just for myself but other developers in the community. 

This would also be a great place to shove all the content the docs have currently referencing older versions. Specifically the difference between the 1.7 and 1.12 way IFluidHandler works.

PR #256, opened 2020-01-21
Added a legacy category, addresses #254 and #216

Aeltumn:
I've set up a concept for a legacy category as suggested in #254. This category also includes a page linking to the 1.12 -> 1.14 primer by @williewillus as suggested in #216 .

I started modding recently after the release of 1.14.x so I don't know anything about the state of the documentation but I assume that the documentation for 1.12 was accurate.

I don't know if my accuracy score for old versions is a good idea but I think it is needed to inform the reader that some documentation versions might not be very accurate. (especially 1.13.x)

### DIFF
--- a/docs/legacy/index.md
+++ b/docs/legacy/index.md
@@ -7,7 +7,9 @@ Forge has existed for years and you can still easily access builds of Forge for 
 
     These old documentation sites are for reference purposes only. Do not ask for help with old versions on the Forge discord or the Forge forums. **You will not receive support when you are using older versions.**
 
+
 ### List of Previously Documented versions
+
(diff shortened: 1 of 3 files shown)


docs/legacy/index.md
```diff
@@ -0,0 +1,16 @@
+Documentation for Legacy Versions
+=================================
+
+Forge has existed for years and you can still easily access builds of Forge for Minecraft versions as old as Minecraft 1.1. There are significant differences between each and every version and it would be an impossible task to support so many different versions. Therefore, Forge uses an LTS system where a previous major Minecraft version is deemed as "LTS" (Long Term Support). Only the latest version and any current LTS versions will have easily accessible documentation and be included in the version dropdown in the sidebar. However, some older versions were LTS once or the latest version at some point and had documentation written, links to old sites with documentation for those versions can be found here.
+
+!!! important
+
+    These old documentation sites are for reference purposes only. Do not ask for help with old versions on the Forge discord or the Forge forums. **You will not receive support when you are using older versions.**
+
+### List of Previously Documented versions
```
tterrag1098, line 11: Need a blank line after this

tterrag1098 (changes requested):

tterrag1098 (approved):

PR #260, opened 2020-02-08
Updated Environment Setup instructions for IntelliJ

TheGreyGhost:
The gradle tasks have been changed since the original instructions were written.  IDEA 2019 also performs most of the setup without manual tweaks required.

### DIFF
--- a/docs/forgedev/index.md
+++ b/docs/forgedev/index.md
@@ -62,11 +62,22 @@ If you're more of a visual person, cpw has uploaded [a video](https://www.youtub
 !!! Note
     These steps will only work reliably from IDEA version 2016 onwards. Older versions didn't have the appropriate Gradle support and did not support Forge development workspaces.
 
+#### IDEA 2019
+
+ 1. Import Forge's `build.gradle` as an IDEA project. For this, simply click `Import Project` from the `Welcome to IntelliJ IDEA` splash screen, then select the `build.gradle` file.
+1. After IDEA is done importing the project and indexing the files, run the Gradle setup task.  Either:
+   1. open the Gradle sidebar on the right hand side of your screen, then open the `forge` project tree, select `Tasks`, then `other` and double-click the `setup` task (may also appear as `MinecraftForge[Setup]`.  Or alternatively:
+   1. tap the CTRL key twice, and type `gradle setup` in the `Run` command window that pops up.
+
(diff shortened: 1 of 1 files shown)


tterrag1098 (changes requested):
I see no reason to keep information that is over a year old. We can expect people to be on a decently new version of IntelliJ. If you want, make a note that it only works on 2019+.

TheGreyGhost:
OK, you mean - delete the "IDEA older versions" section, and just leave the "IDEA 2019" section?

Thanks very much for the docs by the way.  They sure have helped me a lot.

docs/forgedev/index.md
```diff
@@ -62,11 +62,22 @@ If you're more of a visual person, cpw has uploaded [a video](https://www.youtub
 !!! Note
     These steps will only work reliably from IDEA version 2016 onwards. Older versions didn't have the appropriate Gradle support and did not support Forge development workspaces.
 
+#### IDEA 2019
+
+ 1. Import Forge's `build.gradle` as an IDEA project. For this, simply click `Import Project` from the `Welcome to IntelliJ IDEA` splash screen, then select the `build.gradle` file.
+1. After IDEA is done importing the project and indexing the files, run the Gradle setup task.  Either:
+   1. open the Gradle sidebar on the right hand side of your screen, then open the `forge` project tree, select `Tasks`, then `other` and double-click the `setup` task (may also appear as `MinecraftForge[Setup]`.  Or alternatively:
```
EmmanuelMess, line 65: This should be left as an alternative IMO.

docs/forgedev/index.md
```diff
@@ -62,11 +62,22 @@ If you're more of a visual person, cpw has uploaded [a video](https://www.youtub
...
+   1. tap the CTRL key twice, and type `gradle setup` in the `Run` command window that pops up.
+
+You can then run Forge using the `forge_client` gradle task (`Tasks -> fg_runs -> forge_client`): right-click the task and select either `Run` or `Debug` as desired.
+
+#### IDEA older versions
```
EmmanuelMess, line 70: You should edit the disclamer at the top so that "only 2019 and newer versions should be used"

tterrag1098:
> OK, you mean - delete the "IDEA older versions" section, and just leave the "IDEA 2019" section?

Yes. And perhaps futureproof that definition with "2019+"

TheGreyGhost:
Updated as requested...

tterrag1098 (approved):

PR #263, opened 2020-02-14
Fixed critical typo in mods.toml...

realmayus:
...that is crashing the game


LexManos:
We should REALLY be replacing this value from the build script during packaging. So we don't have to remember to bump it every update.

DaemonUmbra:
This particular issue isn't a version change, looks like an erroneous square bracket being removed.
Also this is the RTD, not the MDK

LexManos:
Ah, read the wrong repo name. Sorry.

Aeltumn:
Oops, this is probably my mistake. Why are you copying the mods.toml from the documentation anyways? There is no difference to the one in the mdk.

tterrag1098 (approved):

PR #269, opened 2020-03-24
Change ItemBlock to BlockItem

GloriousAlpaca:
I have also edited the constructor parameters, as there is an additional parameter that was not mentioned.


Aeltumn:
This has already been changed on the 1.14.x branch. It just hasn't been merged into 1.15.x yet.

PR #274, opened 2020-04-11
Docs for global loot modifiers

Draco18s:
Created some documentation regarding the global loot modifiers system.

Wording adjustments or the order of the subsections can be made, I was doing my best to keep a sensible order to things, but as all three tie together so closely I chose "simplest" to "most complex."


tterrag1098:
Please target 1.14.x as it is the current active dev branch.

Draco18s:
I had no idea that was the case.

zhugeo:
Documentation is going to be more helpful with this page.

Aeltumn:
@Draco18s Shouldn't you close this PR considering you've made a seperate PR for 1.14?

Issue #281, opened 2020-04-21
[1.15.x] TileEntity ITickable -> ITickableTileEntity


Terrails:
The "[Ticking TileEntities](https://mcforge.readthedocs.io/en/1.15.x/tileentities/tileentity/#ticking-tileentities)" part is using `net.minecraft.util.ITickable` instead of `net.minecraft.tileentity.ITickableTileEntity`

PR #250, opened 2020-01-16
Fix TileEntity references for 1.13+

Cadiboo:
Fixes references and corrects a couple typos.
`IBlockstate` -> `Blockstate`
`NBTTagCompound` -> `CompoundNBT`
`IBlockstate` -> `Blockstate`
`ITickable` -> `ITickableTileEntity`
`SPacketUpdateTileEntity` -> `SUpdateTileEntityPacket`

PR #282, opened 2020-04-21
[1.15.x] ITickable -> ITickableTileEntity

3TUSK:
In 1.15.+, the correct interface to make tile entity tickable is `ITickableTileEntity`.

Close MinecraftForge/Documentation#281


3TUSK:
Didn't notice that #250 has the same fix. 
However, #250 targets 1.14.x branch while this one targets 1.15.x.

PR #290, opened 2020-05-16
Docs for global loot modifiers [1.14 branch]

Draco18s:
Finally got some time to deal with this. Related to #274


Draco18s:
I hate the hoops I have to jump through to fetch things from upstream. `loot_tables.md` looks like I added the entire file when in reality I only changed line 65.

docs/items/globallootmodifiers.md
```diff
@@ -0,0 +1,129 @@
...
+`replace` causes the cache of modifiers to be cleared fully when this asset loads (mods are loaded in an order that may be specified by a data pack). For modders you will want to use `false` while data pack makers may want to specify their overrides with `true`.
+
+`entries` is an *ordered list* of the modifiers that will be loaded. Any modifier that is not listed will not be loaded and the ones listed are called in the order listed. This is primarily relevant to data pack makers for resolving conflicts between modifiers from separate mods.
+
+The serialized json
+-------------------------------
+
+This file contains all of the potential variables related to your modifier, including the conditions that must be met prior to modifying any loot as well as any other parameters your modifier might have. Avoid hard-coded values where ever possible so that data pack makers can adjust balance if they wish to.
+```
+{
```
TheSilkMiner, line 46: It would be better to also have the `type` field in the example file, since grabbing the type from the file name has been deprecated in 1.15 ([Source](https://github.com/MinecraftForge/MinecraftForge/blob/aa5b9d12ea35f6020d3f1d8117b6b4684740fc59/src/main/java/net/minecraftforge/common/loot/LootModifierManager.java#L131)) and plainly doesn't work in 1.14 ([Source](https://github.com/MinecraftForge/MinecraftForge/blob/90403d18f98db35d5f3980e29bb3bb8ab4965167/src/main/java/net/minecraftforge/common/loot/LootModifierManager.java#L131))

zhugeo (commented):
Cool and good

docs/items/globallootmodifiers.md
Draco18s: 👍 I wrote this stuff up before that change was made and forgot about it.

tterrag1098:
The file you changed was removed as it was extremely outdated. You might want to look over it if you only changed one line, there was a lot of old info there. I'll give this a grammar review soon, but since I don't know the system that well (and you...wrote it) I'll probably just trust that the content is accurate.

Draco18s:
@tterrag1098 Hm, might have to revert that file then. Thanks; source tree was not super helpful about what the merge conflict actually *was.*

docs/items/globallootmodifiers.md
Draco18s: Deleted that file and added the new page to navigation.

docs/items/globallootmodifiers.md
```diff
@@ -0,0 +1,129 @@
...
+===========
+
+Global Loot Modifiers are a data-driven method of handling modification of harvested drops without the need to overwrite dozens to hundreds of vanilla loot tables or to handle effects that would require interactions with another mod's loot tables without knowing what mods may be loaded. Global Loot Modifiers are also stacking, rather than last-load-wins as modifications to loot tables would be.
+
+Registering a Global Loot Modifier
+-------------------------------
+
+You will need 3 things:
+1. Create a global_loot_modifiers.json file at /data/forge/loot_modifiers/
+	This will tell Forge about your modifiers and works similar to [tags].
```
Aeltumn, line 11: I don't think the link to tags will work like this. Don't you need to type it as `[tags][tags]`?

docs/items/globallootmodifiers.md
```diff
@@ -0,0 +1,129 @@
...
+2. A serialized json representing your modifier
+    This will contain all of the data about your modification and allows data packs to tweak your effect.
+3. A class that extends `LootModifier`
+    The operational code that makes your modifier work and associated serializer.
+
+Finally, the serializer for your operational class is [registered] as any other `ForgeRegistryEntry`.
```
Aeltumn, line 17: Same as with `[tags]`, I think this should be `[registered][registered]`

docs/items/globallootmodifiers.md
```diff
@@ -0,0 +1,129 @@
...
+```
+
+
+The critical portion is the `doApply` method.
+
+This method is only called if the `conditions` specified return `true` and the modder is now able to make the modifications they desire. In this case we can see that the number of `itemToCheck` meets or exceeds the `numSeedsToConvert` before modifying the list by adding an `itemReward` and removing any excess `itemToCheck` stacks, matching the previously mentioned effects: When a wheat block is harvested with shears, if enough seeds are generated as loot, they are converted to additional wheat instead.
+
+Also take note of the `read` method in the serializer. The conditions are already deserialized for you and if you have no other data, simply `return new MyModifier(conditionsIn)`. However, the full `JsonObject` is available if needed.
+
+Additional [examples] can be found on the Forge Git repository, including silk touch and smelting effects.
```
Aeltumn, line 127: Same as with the other two, shouldn't it be `[examples][examples]`?

Aeltumn (changes requested):
I don't think I can give any feedback about a system I've never used to the person who made the system, so I'll assume this PR is accurate. 😄
~~Though you do have 3 links that I don't think are written correctly.~~ Sorry, I was wrong about this!

docs/items/globallootmodifiers.md
Draco18s: See: https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#links
The type I'm using is `Or leave it empty and use the [link text itself].`
The URL gets defined a bit further down.

Aeltumn (approved):

tterrag1098:
This has had enough eyes, and you are the author of the system. Good enough for me.

tterrag1098 (approved):

Issue #275, opened 2020-04-13
Outdated info about what RegistryEvents are supported for


RuntimeException6:
The 1.15 docs, under Registries and Registering Things, read:
"RegistryEvents are currently supported for the following types: Block, Item, Potion, Biome, SoundEvent, PotionType, Enchantment, IRecipe, VillagerProfession, EntityEntry"
RegistryEvents aren't supported for IRecipes, and PotionType and EntityEntry don't seem to exist anymore, at least under those names.

PR #297, opened 2020-05-23
[1.14.x] Rewrite docs for Registries

sciwhiz12:
Rewrote [Concepts > Registries]:
- Removed references to `GameRegistry.register`
- Reordered register methods so `DeferredRegister` is first and foremost
- Simplified section on creating registries _(if someone will make their own registry, they'll be capable of researching on their own)_
- Changed `ObjectHolder` rules from block of text to rules list _(TIL that `final` only matters if the class and not the field has `@ObjectHolder`)_

Issue #275 can be closed, and I think the related issue #85 can be closed, since the issue is from long, long ago (2017), and both points are resolved:
 1. `GameRegistry.register` in `preInit` is non-existent at this point; and
 2. `RegistryBuilder` is only used when a modder wants to adds a registry, which is not common I would assume.

### DIFF
--- a/docs/concepts/registries.md
+++ b/docs/concepts/registries.md
@@ -1,34 +1,22 @@
 Registries
 ==========
 
-Registration is the process of taking the objects of a mod (items, blocks, sounds, etc.) and making them known to the game. Registering things is important, as without registration the game will simply not know about these objects in a mod and will exhibit great amounts of unexplainable behavior (and probably crash). Some examples of things that need to be registered are `Block`s, `Item`s, `Biome`s.
+Registration is the process of taking the objects of a mod (such as items, blocks, sounds, etc.) and making them known to the game. Registering things is important, as without registration the game will simply not know about these objects in a mod and will cause unexplainable behavior and crashes. 
 
-Most things that require registration in the game are handled by the Forge registries. A registry is a simple object similar to a map that assigns values to keys. Additionally, they automatically assign integer IDs to values. Forge uses registries with [`ResourceLocation`][ResourceLocation] keys to register objects. This allows the `ResourceLocation` to act like a "registry name" for the object. The registry name for an object may be accessed with `get`/`setRegistryName`. The setter can only ever be called once, and calling it twice results in an exception. Every type of registrable object has its own registry, and names in two different registries will not collide. (E.g. there's a registry for `Block`s, and a registry for `Item`s, and a `Block` and an `Item` may be registered with the same name `mod:example` without colliding. However, if two blocks were registered with that name, an exception would be thrown.)
+Most things that require registration in the game are handled by the Forge registries. A registry is a object similar to a map that assigns values to keys. Forge uses registries with [`ResourceLocation`][ResourceLocation] keys to register objects. This allows the `ResourceLocation` to act as the "registry name" for objects. The registry name for an object may be accessed with `#getRegistryName`/`#setRegistryName`. The setter can only be called once; calling it twice results in an exception. 
 
-Registering Things
+Every type of registrable object has its own registry. To see all registries supported by Forge, see the `ForgeRegistries` class. All registry names within a registry must be unique. However, names in different registries will not collide. For example, there's a `Block` registry, and an `Item` registry. A `Block` and an `Item` may be registered with the same name `example:thing` without colliding; however, if two `Block`s or two `Item`s were registered with the same exact name, an exception will occur.
+
+Methods for Registering
 ------------------
 
-The recommended way to register things is through the `RegistryEvent`s. These [events][] are fired after mod constructors are called and before configs are loaded. In `RegistryEvent.NewRegistry`, registries should be created. Later, `RegistryEvent.Register` is fired once for each registered registry. Because `Register` is a generic event, the event handler should set the type parameter to the type of the object being registered. The event will contain the registry to register things to (`getRegistry`), and things may be registered with `register` (or `registerAll`) on the registry. Here's an example of an event handler that registers blocks:
-
-```java
-@SubscribeEvent
-public void registerBlocks(RegistryEvent.Register<Block> event) {
-    event.getRegistry().registerAll(block1, block2, ...);
-}
-```
-
-The order in which `RegistryEvent.Register` events fire is alphabetically, with the exception that `Block` will *always* fire first, and `Item` will *always* fire second, right after `Block`. After the `Register<Block>` event has fired, all [`ObjectHolder`][ObjectHolder] annotations are refreshed, and after `Register<Item>` has fired they are refreshed again. They are refreshed for a third time after *all* of the other `Register` events have fired.
-
-`RegistryEvent`s are currently supported for the following types: `Block`, `Item`, `Potion`, `Biome`, `SoundEvent`, `PotionType`, `Enchantment`, `IRecipe`, `VillagerProfession`, `EntityEntry`
-
-There is another, older way of registering objects into registries, using `GameRegistry.register`. Anytime something suggests using this method, it should be replaced with an event handler for the appropriate registry event. This method simply finds the registry corresponding to an `IForgeRegistryEntry` with `IForgeRegistryEntry::getRegistryType`, and then registers the object to the registry. There is also a convenience overload that takes an `IForgeRegistryEntry` and a `ResourceLocation`, which is equivalent to calling `IForgeRegistryEntry::setRegistryName`, followed by a `GameRegistry.register` call.
-
-!!! information
-	Registering an `Entity` or `TileEntity` might be a little bit confusing at first as it doesn't use the `Entity` or `TileEntity` classes, but an `EntityType` or `TileEntityType`. These are created by making use of `EntityType.Builder` or `TileEntityType.Builder`.
-	The `String` parameter of the builder's `build` method is a data-fixer id. Data fixers do not work with mods (yet) so you should pass `null` in.
+There are two proper ways to register objects: the `DeferredRegister` class, and the `RegistryEvent.Register` lifecycle event.
 
 ### DeferredRegister
-`DeferredRegister` is a new way of registering objects that replicates the simplicity of static initialisers while supporting registry overrides and avoiding the issues caused by static initialisation. `DeferredRegister` is well documented (check it's javadocs). It simply maintains a list of suppliers for entries and registers the objects from those suppliers them during the proper Register event. These suppliers should return **new** instances every time. Here's an example of a mod that uses `DeferredRegister` to register blocks:
+
+`DeferredRegister` is the newer, documented, and recommeded way to register objects. It allows the use and convenience of static initialisers while avoiding the issues associated with it. It simply maintains a list of suppliers for entries and registers the objects from those suppliers during the proper `Register` event.
+
+An example of a mod registering a custom block:
 
 ```java
 private static final DeferredRegister<Block> BLOCKS = new DeferredRegister<>(ForgeRegistries.BLOCKS, MODID);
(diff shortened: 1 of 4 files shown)


docs/concepts/registries.md
```diff
@@ -1,34 +1,22 @@
...
+Most things that require registration in the game are handled by the Forge registries. A registry is a object similar to a map that assigns values to keys. Forge uses registries with [`ResourceLocation`][ResourceLocation] keys to register objects. This allows the `ResourceLocation` to act as the "registry name" for objects. The registry name for an object may be accessed with `#getRegistryName`/`#setRegistryName`. The setter can only be called once; calling it twice results in an exception.
```
Cyborgmas:
```suggestion
Most things that require registration in the game are handled by the Forge registries. A registry is an object similar to a map that assigns values to keys. Forge uses registries with [`ResourceLocation`][ResourceLocation] keys to register objects. This allows the `ResourceLocation` to act as the "registry name" for objects. The registry name for an object may be accessed with `#getRegistryName`/`#setRegistryName`. The setter can only be called once; calling it twice results in an exception. 
```

docs/concepts/registries.md
```diff
@@ -1,34 +1,22 @@
...
-------------------
-
-The recommended way to register things is through the `RegistryEvent`s. These [events][] are fired after mod constructors are called and before configs are loaded. In `RegistryEvent.NewRegistry`, registries should be created. Later, `RegistryEvent.Register` is fired once for each registered registry. Because `Register` is a generic event, the event handler should set the type parameter to the type of the object being registered. The event will contain the registry to register things to (`getRegistry`), and things may be registered with `register` (or `registerAll`) on the registry. Here's an example of an event handler that registers blocks:
+Every type of registrable object has its own registry. To see all registries supported by Forge, see the `ForgeRegistries` class. All registry names within a registry must be unique. However, names in different registries will not collide. For example, there's a `Block` registry, and an `Item` registry. A `Block` and an `Item` may be registered with the same name `example:thing` without colliding; however, if two `Block`s or two `Item`s were registered with the same exact name, an exception will occur.
```
Cyborgmas: is it necessary to specify that names can be the same between registries? you just said they are act as maps, so this seems like a given to me.

docs/concepts/registries.md
```diff
@@ -1,34 +1,22 @@
...
+### DeferredRegister
 
-!!! information
-	Registering an `Entity` or `TileEntity` might be a little bit confusing at first as it doesn't use the `Entity` or `TileEntity` classes, but an `EntityType` or `TileEntityType`. These are created by making use of `EntityType.Builder` or `TileEntityType.Builder`.
-	The `String` parameter of the builder's `build` method is a data-fixer id. Data fixers do not work with mods (yet) so you should pass `null` in.
+`DeferredRegister` is the newer, documented, and recommeded way to register objects. It allows the use and convenience of static initialisers while avoiding the issues associated with it. It simply maintains a list of suppliers for entries and registers the objects from those suppliers during the proper `Register` event.
```
Cyborgmas:
```suggestion
`DeferredRegister` is the newer, documented, and recommended way to register objects. It allows the use and convenience of static initialisers while avoiding the issues associated with it. It simply maintains a list of suppliers for entries and registers the objects from those suppliers during the proper `Register` event.
```

Cyborgmas: i corrected recommended but i would consider dropping the word altogether. since it is just another way. However it is highly frowned upon to use static inits, that much is true.

docs/concepts/registries.md
```diff
@@ -40,63 +28,133 @@ public ExampleMod() {
 }
 ```
 
-Creating Registries
+### `Register` events
+
+The `RegistryEvent`s are the second and more flexible way to register objects. These [events][] are fired after the mod constructors and before the loading of configs.
+
+The event used in registering objects is the `RegistryEvent.Register<T>`. The type parameter `T` should be set to the type of the object being registered. Calling `#getRegistry` will return the registry, upon which objects are registered with `#register` or `#registerAll`.
```
Cyborgmas, line 35: specify that T has to extend IForgeRegistryEntry? a pure suggestion, not sure.

docs/concepts/registries.md
```diff
@@ -40,63 +28,133 @@ public ExampleMod() {
...
-Creating Registries
+### `Register` events
+
+The `RegistryEvent`s are the second and more flexible way to register objects. These [events][] are fired after the mod constructors and before the loading of configs.
+
+The event used in registering objects is the `RegistryEvent.Register<T>`. The type parameter `T` should be set to the type of the object being registered. Calling `#getRegistry` will return the registry, upon which objects are registered with `#register` or `#registerAll`. 
+
+`RegistryEvent.Register<?>` events are fired in this order: first, the `Block` registry, then the `Item` registry, and then all other registries in alphabetical order. 
+
+Here is an example: (note that the event handler must be registered on the *mod event bus* in the mod constructor)
```
Cyborgmas:
```suggestion
Here is an example: (The RegistryEvents must be added to the *mod event bus*)
```
Not sure about my phrasing, but yours seemed a bit subjective.

docs/concepts/registries.md
```diff
@@ -40,63 +28,133 @@ public ExampleMod() {
...
+
+```java
+@SubscribeEvent
+public void registerBlocks(RegistryEvent.Register<Block> event) {
+    event.getRegistry().registerAll(block1, block2, ...);
+}
+```
+
+!!! note
+    `TileEntity` and `Entity` cannot be registered; instead, `TileEntityType` and `EntityType` are registered, and used in `TileEntity`/`Entity` constructors. These are created through the use of `TileEntityType.Builder` and `EntityType.Builder`, respectively. An example: (`REGISTER` refers to a `DeferredRegister<TileEntityType>`)
```
Cyborgmas: Maybe specify ContainerType as well? Also could include that the "Type" classes are basically factories for the actual objects.

docs/concepts/registries.md
```diff
@@ -40,63 +28,133 @@ public ExampleMod() {
...
+The `Class` of the value of the registry must implement `IForgeRegistryEntry`, which defines that `#setRegistryName` and `#getRegistryName` can be called on the objects of that class. It is recommended to extend `ForgeRegistryEntity`, the default implementation instead of implementing the interface directly. When `#setRegistryName(String)` is called with a string, and that string does not have an explicit namespace, its namespace will be set to the current modid.
```
Cyborgmas:
```suggestion
The `Class` of the value of the registry must implement `IForgeRegistryEntry`, which defines that `#setRegistryName` and `#getRegistryName` can be called on the objects of that class. It is recommended to extend `ForgeRegistryEntry`, the default implementation instead of implementing the interface directly. When `#setRegistryName(String)` is called with a string, and that string does not have an explicit namespace, its namespace will be set to the current modid.
```

docs/concepts/registries.md
```diff
@@ -40,63 +28,133 @@ public ExampleMod() {
...
+It is possible to have Forge inject registered object from registries into the `public static` fields of classes. This is done by annotating classes or fields with `@ObjectHolder` and supplying enough information to construct a `ResourceLocation` that identifies a specific object in a specific registry.
```
Cyborgmas: you dont specify when final is needed

docs/concepts/registries.md
```diff
@@ -1,34 +1,22 @@
...
-    event.getRegistry().registerAll(block1, block2, ...);
-}
-```
-
-The order in which `RegistryEvent.Register` events fire is alphabetically, with the exception that `Block` will *always* fire first, and `Item` will *always* fire second, right after `Block`. After the `Register<Block>` event has fired, all [`ObjectHolder`][ObjectHolder] annotations are refreshed, and after `Register<Item>` has fired they are refreshed again. They are refreshed for a third time after *all* of the other `Register` events have fired.
+Methods for Registering
+------------------
 
-`RegistryEvent`s are currently supported for the following types: `Block`, `Item`, `Potion`, `Biome`, `SoundEvent`, `PotionType`, `Enchantment`, `IRecipe`, `VillagerProfession`, `EntityEntry`
+There are two proper ways to register objects: the `DeferredRegister` class, and the `RegistryEvent.Register` lifecycle event.
```
Cyborgmas, line 13: This much is true i guess, but you should mention ObjectHolders since they are a key part in accessing registry information.

Cyborgmas:
first review, and it seems that i do not have the hang of it. Got most stuff twice in each... sorry for the hassle

docs/concepts/registries.md
sciwhiz12: I wrote this because a) it was in the original docs, and I adapted it, and b) when I was recently coding my mod, I forgot that `Block`s and `TileEntity`s don't need seperate names because they're in seperate registries. 
This is mostly to remind beginners that, yes, you can have a `Block` with 'example:thing', and create its ItemBlock also as `example:thing`, and they won't collide.

tterrag1098: In fact it's necessary for them to collide to make the block->item map pick up the relationship

docs/concepts/registries.md
sciwhiz12: I'll just drop the word, since it is just another way. However, I'm keeping it at the top of the list, to subtly suggest beginners to use it first before trying the `RegistryEvent.Register` events.

docs/concepts/registries.md
sciwhiz12: Telling that to the modder, especially the beginners, means having to make a sentence relaying the use of `IForgeRegistryEntry`. Only modders making custom registries need to know what `IForgeRegistryEntry` means (and there's already a sentence in that section). Better to just keep it as abstract as possible.

docs/concepts/registries.md
sciwhiz12: Reading that, it does seem... off to me.
```suggestion
Here is an example: (the event handler is registered on the *mod event bus*)
```
Might that work instead?

docs/concepts/registries.md
sciwhiz12: Forgot about `ContainerType`, whoops. I'll include that, and that `*Type` classes are factories.

docs/concepts/registries.md
sciwhiz12: As stated in the bulleted rules on `@ObjectHolder`:
```markdown
    * the **enclosing class** has an `@ObjectHolder` annotation, and the field is `final`, then:
```

docs/concepts/registries.md
sciwhiz12: I think that, instead of mentioning `ObjectHolder` again, I'll just move the creating registries section to the bottom of the page, so the section on registering and the section on `ObjectHolder`s are next to each other.

docs/concepts/registries.md
sciwhiz12: Thank you, didn't catch that mistype.

docs/concepts/registries.md
sciwhiz12: Thank you, didn't catch that mistype.

docs/concepts/registries.md
Cyborgmas: sure

docs/concepts/registries.md
```diff
@@ -1,34 +1,22 @@
 Registries
 ==========
 
-Registration is the process of taking the objects of a mod (items, blocks, sounds, etc.) and making them known to the game. Registering things is important, as without registration the game will simply not know about these objects in a mod and will exhibit great amounts of unexplainable behavior (and probably crash). Some examples of things that need to be registered are `Block`s, `Item`s, `Biome`s.
+Registration is the process of taking the objects of a mod (such as items, blocks, sounds, etc.) and making them known to the game. Registering things is important, as without registration the game will simply not know about these objects in a mod and will cause unexplainable behavior and crashes.
```
Aeltumn: > the game will simply not know about these objects in a mod and will cause unexplainable behavior and crashes. 

I would replace `and will` with `which will`

docs/concepts/registries.md
```diff
@@ -40,63 +28,136 @@ public ExampleMod() {
...
 
-Creating Registries
--------------------
+### `Register` events
+
+The `RegistryEvent`s are the second and more flexible way to register objects. These [events][] are fired after the mod constructors and before the loading of configs.
+
+The event used in registering objects is the `RegistryEvent.Register<T>`. The type parameter `T` should be set to the type of the object being registered. Calling `#getRegistry` will return the registry, upon which objects are registered with `#register` or `#registerAll`. 
+
+`RegistryEvent.Register<?>` events are fired in this order: first, the `Block` registry, then the `Item` registry, and then all other registries in alphabetical order.
```
Aeltumn: > and then all other registries in alphabetical order. 

While Block and Item are first the other registries are not ordered in any way. Mods shouldn't break even if the registry events were called in a random order every startup.

Aeltumn (dismissed):
Everything looks pretty good, I've only got two small comments. Although I've never used the @ ObjectHolder so I can't give any feedback on that part.

docs/concepts/registries.md
sciwhiz12:
```java
// GameData.class, fireRegistryEvents(), L951-952, MC 1.14.4, mappings 20190719-1.14.3
// GameData.class, fireRegistryEvents(), L978-979, MC 1.15.2, mappings 20200514-1.15.1
List<ResourceLocation> keys = Lists.newArrayList(RegistryManager.ACTIVE.registries.keySet());
keys.sort((o1, o2) -> String.valueOf(o1).compareToIgnoreCase(String.valueOf(o2)));
```
The javadoc for `String#compareToIgnoreCase` is:
 > Compares two strings lexicographically, ignoring case differences.

So, current registry behavior is that Block and Item are first (evidenced by the lines following the aforementioned code), then all other registries in lexicographical/alphabetical order. This is supported by looking at run logs, which give the stated order (`minecraft:block`, `minecraft:item`, the `forge` registries in alphabetical order, then the `minecraft` registries in alphabetical order). Forge registries run before Minecraft ones because `f`orge is higher than `m`inecraft.

I somewhat agree that mods shoudn't break when registry events after Block and Item are random. However, by having the registries load in lexicographical order, it gives modders confidence that this registry will *always* be run before another. I don't know how that knowledge will benefit the modder, but at least they have it.

docs/concepts/registries.md
sciwhiz12: Will do.

docs/concepts/registries.md
TheSilkMiner: That is the problem: you shouldn't give modders the confidence that registries will load in a certain order, since registries should be completely indipendent of each other. Moreover, as outlined in FAP20, there was an idea to make registry events parallel, meaning that multiple registry events may fire at the same time on different threads.

sciwhiz12: But that idea of parallel registry events are still that: an idea. Until Forge changes from the current behavior, we should still document the current behavior.

Once parallel registry events are introduced and registry loader order cannot be guaranteed, this line will be removed. For now, it exists to inform the modder that there is an load order: Blocks, then Items, then everything else.

To me, it also guides a person from deducing: "So, Blocks are always after Items? So, I can at least reference Blocks in the Items registry event load."

Aeltumn: My comment was based on FAP20 where they talked about how the registries could be called in parallel. Them running in alphabetical order is not a feature, it's just how it happens to work. If it's noted in the documentation someone might base their code on it and find out in a few versions it doesn't work anymore. I think we either shouldn't mention the order in which they are called at all or already mention one should assume the order is random.

sciwhiz12: So, after re-watching that part of FAP20, it does seem that random registry load order may happen in the future. Since I do not want to put an assumption into documentation that currently contradicts current behavior, I'll remove that paragraph altogether.

sciwhiz12:
As 1.14.x is moving into out-of-support, I am closing this PR. See the [updated PR for 1.15.x](https://github.com/MinecraftForge/Documentation/pull/312).

sciwhiz12:
As said to me, 1.14.x is the branch to currently target. Since 1.14.x is essentially the same to 1.15.x in regards to registries, the reviews from the other branch will apply to this branch.

sciwhiz12:
I might have to make a PR once this is merged into 1.15.x, to change the section on `DeferredRegister` from using the constructor to `#create`.

docs/concepts/registries.md
```diff
@@ -1,102 +1,174 @@
...
+    Some classes cannot by themselves be registered; instead, `*Type` classes are registered, and used in the formers' constructors. For example, [`TileEntity`][tileentity] has `TileEntityType`, and `Entity` has `EntityType`. These `*Type` classes are factories that simply create the containing type on demand. 
+    
+    These factories are created through the use of their `*Type.Builder` classes. An example: (`REGISTER` refers to a `DeferredRegister<TileEntityType>`)
+    ```java
+    public static final RegistryObject<TileEntityType<ExampleTile>> EXAMPLE_TILE = REGISTER.register(
+        "example_tile", () -> TileEntityType.Builder.create(ExampleTile::new, EXAMPLE_BLOCK.get()).build(null)
+    );
+    ```
+
+Injecting Values Using @ObjectHolder
```
tterrag1098: I would still like a section on `RegistryObject` before this.

tterrag1098 (changes requested):

docs/concepts/registries.md
sciwhiz12: Done, one section of `RegistryObject` as ordered.

tterrag1098 (approved):

PR #300, opened 2020-05-29
Update logoFile description

puggan:
logoFile is not allowed to be in a subfolder, according to Class ResourcePack
```
   public InputStream getRootResourceStream(String fileName) throws IOException {
      if (!fileName.contains("/") && !fileName.contains("\\")) {
         return this.getInputStream(fileName);
      } else {
         throw new IllegalArgumentException("Root resources can only be filenames, not paths (no / allowed!)");
      }
   }
```
called from `InputStream logoResource = resourcePack.getRootResourceStream(logoFile);` Line 430 in ModListScreen.java


Aeltumn (dismissed):

Yunus1903 (dismissed):

AdrianCollado:
You should probably also update the listed `mods.toml` file, as the `logoFile` value contains a nested path.

https://github.com/MinecraftForge/Documentation/blame/f6f90314bd6d96e1d968decae3858f87bb46ecd6/docs/gettingstarted/structuring.md#L40

Also note that the `mods.toml` within the MDK is correct, so the issue is solely with the documentation.

PR #302, opened 2020-06-13
Added example for exposing capabilities in itemstacks + updated `EnumFacing` to `Direction`

ghost:
Updated `EnumFacing` to `Direction` in the exposing capabilities section to reflect mappings on 1.15.x, and added an example of using `initCapabilties` to expose a capability in an itemstack.

### DIFF
--- a/docs/datastorage/capabilities.md
+++ b/docs/datastorage/capabilities.md
@@ -47,7 +47,7 @@ Once you have your own instance of the capability interface, you will want to no
 
 ```Java
 @Override
-public boolean hasCapability(Capability<?> capability, EnumFacing facing) {
+public boolean hasCapability(Capability<?> capability, Direction facing) {
   if (capability == CapabilityItemHandler.ITEM_HANDLER_CAPABILITY) {
     return true;
   }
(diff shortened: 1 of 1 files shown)


sciwhiz12:
Looking at the docs, it seems to be outdated: references to `hasCapability` which does not currently exist, `getCapability` returning the instance (or `null`) when it now returns `LazyOptional`, references to `NBTBase` and `EnumFacing` (even outside your changes, there still are `EnumFacing`s, such as in 'Creating Your Own Capability'). 

I also believe that the section about `IExtendedEntityProperties` should be removed, or at least moved in some way to the 'Legacy Versions' category.
A thorough update is needed to the docs to reflect the changes in the Capability API.

ghost:
That's a good point.  If I have time I shall rewrite the capabilities page then I guess.

Aeltumn (changes requested):
To add on to @sciwhiz12's comments I agree that the entire file should get updated but that rewrite can be a seperate PR.

In general, please target 1.14.x instead of 1.15.x.

Furthermore, I don't think how you've explained the `initCapabilities` method is sufficient. I'm not 100% sure the way I use capabilities is the correct way but I don't think the `ICapabilityProvider` should be inlined, capability providers are more complicated than that. Also having it in here with the `getCapability` method just returning the superclass' `getCapability` method is not useful at all.

tterrag1098:
This should also target 1.14 for the time being.

sciwhiz12:
It seems that this PR has been superseded in part by #325, which also fixes the page for 1.15.x such as moving to LazyOptional. Closing as superseded.

sciwhiz12:
Could you please update the PR to fix the merge conflicts? And please keep the base branch to 1.15.x, as 1.14 is out-of-support now.
_(If you do not wish to continue this PR, please respond soon.)_

PR #303, opened 2020-06-14
Add Mod Configuration to Docs

iPixeli:
Currently the docs do not cover forge's mod configuration features.

Created this new page to cover it in brief.

### DIFF
--- a/docs/config/intro.md
+++ b/docs/config/intro.md
@@ -6,23 +6,41 @@ Forge has built in support for creating and managing mod configuration files. Th
 Creating
 --------
 
-Create a dedicated class. Use the builder instance to define the configuration structure and data.
+Create a dedicated class. Use the builder instance to define the configuration structure and data. Comments are 100% optional. 
 
 ```java
-private BooleanValue mySetting;
+private BooleanValue myBool;
+public IntValue<THING> myInt;
+public EnumValue<THING> myEnum;
 
 public ServerConfig(ForgeConfigSpec.Builder builder) {
   builder.comment("examplemod settings"); // Comment at the start of the file
-  
-  builder.push("Example Category"); // 'Enter' the Category
-  
-  mySetting = builder
-    .comment("Set this to true to enable something") // Comment on the setting
-    .translation("examplemod.exampleboolserver") // Language translation key
-    .define("exampleboolserver", false); // Setting key and default value
 
+  myBool = builder
+    .comment("Set this to true to enable something") // Comment on the setting
+    .translation("examplemod.examplebool") // Language translation key
+    .define("examplebool", false); // Setting key and default value
+  
+  builder.push("Example Category 1"); // 'Enter' the Category  
+  myInt = builder
+    .translation("examplemod.exampleint")
+    .defineEnum("exampleint", THING.ONE); // Valid settings for enums are all values in the enum set
+  builder.pop(); // 'Exit' the category
+  
+  builder.push("Example Category 2"); // 'Enter' the Category
+  myEnum = builder
+    .comment("Choose which thing")
+    .translation("examplemod.exampleenum")
+    .defineInRange("exampleenum", 5, 0, 30); // For integers, you can specify allowed value range
   builder.pop(); // 'Exit' the category
 }
+```
+
+Default Values
+--------------
+
+```java
+
 ```
 
 Registration
@@ -45,7 +63,7 @@ Simply call the setting's get method.
 
 ```java
 protected final static Config serverConfig = serverConfigPair.getLeft();
-serverConfig.mySetting.get()
+serverConfig.myBool.get()
 ```
 
 Changing Values
--- a/docs/config/intro.md
+++ b/docs/config/intro.md
@@ -1,0 +1,59 @@
+Configuration Files
+===================
+
+Forge has built in support for creating and managing mod configuration files. The purpose of a configuration file is to add flexibility and allow end users to customize their experience.
+
+Creating
+--------
+
+Create a dedicated class. Use the builder instance to define the configuration structure and data.
+
+```java
+private BooleanValue mySetting;
+
+public ServerConfig(ForgeConfigSpec.Builder builder) {
+  builder.comment("examplemod settings"); // Comment at the start of the file
+  
+  builder.push("Example Category"); // 'Enter' the Category
+  
+  mySetting = builder
+    .comment("Set this to true to enable something") // Comment on the setting
+    .translation("examplemod.exampleboolserver") // Language translation key
+    .define("exampleboolserver", false); // Setting key and default value
+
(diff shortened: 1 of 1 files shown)


docs/config/intro.md
```diff
@@ -0,0 +1,59 @@
...
+    .define("exampleboolserver", false); // Setting key and default value
+
+  builder.pop(); // 'Exit' the category
+}
+```
+
+Registration
+------------
+
+Create the Pair instance.
```
Aeltumn, line 49: You should explain why this is a Pair and why we're making this instance. Something like:
`To register the configuration you'll first need to build it, for this you can use the ForgeConfigSpec.Builder#configure method. This will return a Pair object which stores both the configuration builder and your configuration file itself. Then you'll need to register this instance of your configuration file.`
The configure method isn't the only way to do this, you can also create a ForgeConfigSpec.Builder instance manually and build it manually.

docs/config/intro.md
```diff
@@ -0,0 +1,59 @@
...
+Creating
+--------
+
+Create a dedicated class. Use the builder instance to define the configuration structure and data.
+
+```java
+private BooleanValue mySetting;
+
+public ServerConfig(ForgeConfigSpec.Builder builder) {
+  builder.comment("examplemod settings"); // Comment at the start of the file
```
Aeltumn, line 17: You don't need a comment at the start of the file. You should instead mention that you can call comment before push/pull to add comments above/below categories.

docs/config/intro.md
```diff
@@ -0,0 +1,59 @@
...
+  
+  builder.push("Example Category"); // 'Enter' the Category
```
Aeltumn: Having two comments saying 'Enter' the category and 'Exit' the category could better be replaced with one comment explaining how push/pull creates categories. It's not super clear what 'entering' a category means.

docs/config/intro.md
```diff
@@ -0,0 +1,59 @@
...
+------------
+
+Create the Pair instance.
+
+```java
+protected final static Pair<ServerConfig, ForgeConfigSpec> serverConfigPair = new ForgeConfigSpec.Builder().configure(ServerConfig::new);
+```
+
+```java
+ModLoadingContext.get().registerConfig(ModConfig.Type.SERVER, serverConfigPair.getRight());
```
Aeltumn, line 56: It should be explained what the various config types do and where the configuration files are created in the game folder for each type.

Aeltumn (changes requested):
This documentation is a good start but it needs a more detailed explaination.

docs/config/intro.md
iPixeli: Yes, More words could be added here.

docs/config/intro.md
iPixeli: It should be obvious that all comments are not required. My inclusion of comments are purely for demonstration.

docs/config/intro.md
iPixeli: Push and Pop are common terminology across a plethora of programming languages out there.

docs/config/intro.md
iPixeli: There are three. Client, common and server. Forge's Javadocs contains the info you mentioned. Explanation may be overkill.

docs/config/intro.md
iPixeli: Added text " Comments are 100% optional." for clarity.

iPixeli (commented):
The aim of content I submitted is an attempt to adhere to Forge's docs [style guide](https://mcforge.readthedocs.io/en/1.14.x/styleguide/). Most prominently, this is not a java tutorial. What level of detail are you looking for?

docs/config/intro.md
Aeltumn: Yes of course, but these comments feel like you didn't want to not explain anything about the push/pop but now the comments aren't very useful. I'd say to replace the `// 'Enter' the category` and `// 'Exit' the category` comments with one `// Use the push/pop methods to create categories` comment. But this isn't super important.

docs/config/intro.md
Aeltumn: My problem was rather that the comment is not in fact necessarily at the start of the file, not about them being required or not. The comment before a category should explain what kind of settings there are in the category, someone seeing it here would assume it's a generic header comment above the file whilst it isn't, unless you make one big category at the top.

docs/config/intro.md
Aeltumn: There's no harm in mentioning these three types, that common configurations are synced when players log in to the server and where they are saved in the game folder. This is information modders porting from old versions won't know and having the information on the documentation website is better than having it in javadocs. If you don't feel the need to explain it that's fine, but I'd have added it.

tterrag1098 (changes requested):
This needs more content beyond the code examples. I would suggest explaining more about ForgeConfigSpec and ConfigValue rather than jumping straight to examples.

docs/config/intro.md
```diff
@@ -0,0 +1,77 @@
...
+  myInt = builder
+    .translation("examplemod.exampleint")
+    .defineEnum("exampleint", THING.ONE); // Valid settings for enums are all values in the enum set
+  builder.pop(); // 'Exit' the category
+  
+  builder.push("Example Category 2"); // 'Enter' the Category
+  myEnum = builder
+    .comment("Choose which thing")
+    .translation("examplemod.exampleenum")
+    .defineInRange("exampleenum", 5, 0, 30); // For integers, you can specify allowed value range
```
LemADEC, line 34: You're defining an integer range for an enum here. Above, you're defining an enum for an integer.

docs/config/intro.md
```diff
@@ -0,0 +1,77 @@
...
+Forge has built in support for creating and managing mod configuration files. The purpose of a configuration file is to add flexibility and allow end users to customize their experience.
+
+Creating
+--------
+
+Create a dedicated class. Use the builder instance to define the configuration structure and data. Comments are 100% optional. 
+
+```java
+private BooleanValue myBool;
+public IntValue<THING> myInt;
```
LemADEC, line 13: At least on 1.15, there's no parametrization for IntValue. So I guess it's a typo?

docs/config/intro.md
LemADEC: Also, ForgeConfigSpec.Builder.define() returns a ConfigValue<Integer>, while ForgeConfigSpec.Builder.defineInRange() returns a IntValue. So I suggest to change the comment to "For integers, it's recommended to set an allowed value range".

sciwhiz12:
Could you please rebase this PR ontop of the 1.15.x branch instead? And I agree with the previous reviewers, there should be more explanations on how the config system works instead of mostly code examples (which are ripe for copy-pasting).
_(If you do not wish to continue this PR, please respond soon.)_

PR #304, opened 2020-06-19
Remove the mentioning of `GameRegistry.register`

3TUSK:
It has been gone for quite a long while now and leaving it hear serves for no purpose.

### DIFF
--- a/docs/concepts/registries.md
+++ b/docs/concepts/registries.md
@@ -21,8 +21,6 @@ The order in which `RegistryEvent.Register` events fire is alphabetically, with 
 
 `RegistryEvent`s are currently supported for the following types: `Block`, `Item`, `Potion`, `Biome`, `SoundEvent`, `PotionType`, `Enchantment`, `IRecipe`, `VillagerProfession`, `EntityEntry`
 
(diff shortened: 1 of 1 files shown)


Aeltumn (approved):
This changes are fine but I don't know if this information needs to be removed. The documentation already states that this is not the correct method. I don't know if there's any harm in keeping this information in there.

3TUSK:
> The documentation already states that this is not the correct method.

Back in 1.12.x, trying to call this would already give you a compile error (it was `private`). You should have had enough time to migrate to the `RegistryEvent.Register<T>`. FWIW, that paragraph in question was also in the 1.12.x branch; it is fair to assume that that paragraph has finished its mission of smoothing transition.

> I don't know if there's any harm in keeping this information in there.

Personally, I do not see if there is any good, either. One can only hope that no one asks "where did my method go?!" and ignores the rest of article…

sciwhiz12:
Just commenting that https://github.com/MinecraftForge/Documentation/pull/297 also removes the mentioning of `GameRegistry`.

3TUSK:
Interesting, guess I am going to close mine to avoid clash. Thanks for heads up sciwhiz12!

PR #312, opened 2020-06-30
[1.15.x] Rewrite docs for Registries

### LINKED ISSUE #275: Outdated info about what RegistryEvents are supported for

sciwhiz12:
Rewrote the documentation for __[Concepts > Registries]__:
 * Removed outdated references to GameRegistry.register
 * Reordered register methods so DeferredRegister is first and foremost
 * Simplified section on creating registries (if someone makes a custom registry, they'll be researching on their own)
 * Clarified ObjectHolder rules from block of text to rules list
 * Removed mentions of any registry load order (to future-proof when registry events are parallel-dispatched)

This is an update of my [earlier PR for 1.14.x][1.14.xPR]. See that PR for earlier reviews.
Closes #275 and #85.

[1.14.xPR]: https://github.com/MinecraftForge/Documentation/pull/297

### DIFF
--- a/docs/concepts/registries.md
+++ b/docs/concepts/registries.md
@@ -1,34 +1,22 @@
 Registries
 ==========
 
-Registration is the process of taking the objects of a mod (items, blocks, sounds, etc.) and making them known to the game. Registering things is important, as without registration the game will simply not know about these objects in a mod and will exhibit great amounts of unexplainable behavior (and probably crash). Some examples of things that need to be registered are `Block`s, `Item`s, `Biome`s.
+Registration is the process of taking the objects of a mod (such as items, blocks, sounds, etc.) and making them known to the game. Registering things is important, as without registration the game will simply not know about these objects in a mod and will cause unexplainable behavior and crashes. 
 
-Most things that require registration in the game are handled by the Forge registries. A registry is a simple object similar to a map that assigns values to keys. Additionally, they automatically assign integer IDs to values. Forge uses registries with [`ResourceLocation`][ResourceLocation] keys to register objects. This allows the `ResourceLocation` to act like a "registry name" for the object. The registry name for an object may be accessed with `get`/`setRegistryName`. The setter can only ever be called once, and calling it twice results in an exception. Every type of registrable object has its own registry, and names in two different registries will not collide. (E.g. there's a registry for `Block`s, and a registry for `Item`s, and a `Block` and an `Item` may be registered with the same name `mod:example` without colliding. However, if two blocks were registered with that name, an exception would be thrown.)
+Most things that require registration in the game are handled by the Forge registries. A registry is a object similar to a map that assigns values to keys. Forge uses registries with [`ResourceLocation`][ResourceLocation] keys to register objects. This allows the `ResourceLocation` to act as the "registry name" for objects. The registry name for an object may be accessed with `#getRegistryName`/`#setRegistryName`. The setter can only be called once; calling it twice results in an exception. 
 
-Registering Things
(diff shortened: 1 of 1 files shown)


docs/concepts/registries.md
```diff
@@ -1,102 +1,174 @@
...
+Most things that require registration in the game are handled by the Forge registries. A registry is an object similar to a map that assigns values to keys. Forge uses registries with [`ResourceLocation`][ResourceLocation] keys to register objects. This allows the `ResourceLocation` to act as the "registry name" for objects. The registry name for an object may be accessed with `#getRegistryName`/`#setRegistryName`. The setter can only be called once; calling it twice results in an exception. 
 
-Registering Things
```
tterrag1098, line 8: I think this section is linked quite a lot, make sure to find references and update them.

docs/concepts/registries.md
```diff
@@ -1,102 +1,174 @@
...
+Here is an example: (the event handler is registered on the *mod event bus*)
 
-Injecting Registry Values Into Fields
+```java
+@SubscribeEvent
+public void registerBlocks(RegistryEvent.Register<Block> event) {
+    event.getRegistry().registerAll(block1, block2, ...);
```
tterrag1098, line 42: This is a poor example, as registry entries should be constructed at the time of registration. Not held in some field that's initialized at any time. This is the key advantage of `DeferredRegister`. Obviously we need to document both, but the docs should promote the absolute best practices.

docs/concepts/registries.md
```diff
@@ -1,102 +1,174 @@
...
+}
+```
+
+!!! note
+    Some classes cannot by themselves be registered; instead, `*Type` classes are registered, and used in the formers' constructors. For example, [`TileEntity`][tileentity] has `TileEntityType`, and `Entity` has `EntityType`. These `*Type` classes are factories that simply create the containing type on demand.
```
tterrag1098, line 47: I think there needs to be a blank line after the admonition

docs/concepts/registries.md
```diff
@@ -1,102 +1,174 @@
...
+It is possible to have Forge inject registered object from registries into the `public static` fields of classes. This is done by annotating classes or fields with `@ObjectHolder` and supplying enough information to construct a `ResourceLocation` that identifies a specific object in a specific registry.
```
tterrag1098, line 59: This section should take a back seat to a section about `RegistryObject` which should also be documented. If you are injecting third-party entries this is the far superior way to do it.

tterrag1098 (changes requested):
This is good stuff, much needed improvements. Just a couple small things that need addressing and then it's good to go.

tterrag1098:
Also, everything in here is relevant for 1.14.x too, so it should target that branch.

docs/concepts/registries.md
sciwhiz12: Nope, the admonition works fine without a blank line.

docs/concepts/registries.md
sciwhiz12: Hm, I only recently learned that `RegistryObject`s can be instantiated; I thought that it was a `DeferredRegister` class.
I'm going to make a new section on this.

sciwhiz12:
Since 1.14.x is essentially the same as 1.15.x in terms of registries, I'll integrate these reviews into the 1.14.x branch and close this PR. Thank you.

docs/concepts/registries.md
sciwhiz12: Yeah, I didn't see the implications of using `block1` being construed as fields.

docs/concepts/registries.md
sciwhiz12: Found 3 references and updated.

PR #315, opened 2020-07-06
Update capabilities.md

mrp-v2:
-remove hasCapabilty, as it no longer exists
-update getCapability return type and parameters
-explain that hasCapability has been replaced by checking if getCapability != LazyOptional.empty()


mrp-v2:
moving to 1.16 pull request

tterrag1098:
Why? Please target the default branch for any changes not *specific* to newer versions.

mrp-v2:
Ah, I hadn't looked at previous versions to see when this change occurred. I now see that it was between 1.12-1.13. Since this change is already part of #325, no action is necessary.

PR #316, opened 2020-07-08
Fix & Improve Tag in-code documentation

Ktar5:
I fixed a misnamed method in the code samples and also cleaned them up, added some more comments, and tried to make it more understandable after I had some trouble learning them myself. Hopefully this will benefit future users coming to this page.

### DIFF
--- a/docs/utilities/tags.md
+++ b/docs/utilities/tags.md
@@ -1,7 +1,7 @@
 Tags
 ====
-
 Tags are generalized sets of objects in the game, used for grouping related things together and providing fast membership checks.
+Note: Tags have replaced the OreDictionary-- from now on you should (and can only) use tags for all that OreDictionary did. Please see the extra notice at the bottom of this page for more information for transferring from OreDictionary to Tags.
 
 Declaring Your Own Groupings
 ----------------------------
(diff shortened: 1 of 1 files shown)


docs/utilities/tags.md
```diff
@@ -21,17 +21,45 @@ Using Tags In Code
...
+// Tags work through Items, not ItemStacks so we need to get the Item from our ItemStack
 Item unknownItem = stack.getItem();
-boolean isInGroup = ItemTags.getCollection().getOrCreateTag(myTagId).contains(unknownItem);
-// alternatively, can use getTag and perform a null check
+// This line will either retrieve the Tag (referenced by our ResourceLocation) or create it if it is not
+// in the ItemTags collection of all the tags. Then it will check if our Item (again, not ItemStack)
+// exists in the set of Items that belong to this Tag
+boolean isInGroup = ItemTags.getCollection().getOrCreate(myTagId).contains(unknownItem);
+// Alternatively, you can use the getTag and perform a null check to see if the tag exists and do it that way
+Tag retrievedTag = ItemTags.getCollection.get(myTagId);
```
Alpvax: Typo: missing parentheses.
I believe that should be getCollection()

Ktar5: Indeed! Changing it right now.

Ktar5:
Any update on this?

docs/utilities/tags.md
```diff
@@ -21,17 +21,45 @@ Using Tags In Code
...
 Block, Item, and Fluid tags are automatically sent from the server to any remote clients on login and reload. Function tags are not synced.
 
+Tags can be accessed in code through the `Tags.Items` and `Tags.Blocks` classes for all the Vanilla Minecraft tags, and through `ResourceLocation`s for ForgeTags.
+See the examples below for how to use each.
+
 `BlockTags.getCollection()` and `ItemTags.getCollection()` will retrieve the current `TagCollection`, from which you can retrieve a `Tag` object by its ID.
 With a `Tag` object in hand, membership can be tested with `tag.contains(thing)`, or all the objects in the tag queried with `tag.getAllElements()`.
 
-As an example:
+A reminder that `Tag`s contain `Item`s not `ItemStack`s. You can get the `Item` of an `ItemStack` from the stack by calling `<itemStack>.getItem()`.
```
tterrag1098, line 31: This isn't necessary to say.

docs/utilities/tags.md
```diff
@@ -21,17 +21,45 @@ Using Tags In Code
 ------------------
 Block, Item, and Fluid tags are automatically sent from the server to any remote clients on login and reload. Function tags are not synced.
 
+Tags can be accessed in code through the `Tags.Items` and `Tags.Blocks` classes for all the Vanilla Minecraft tags, and through `ResourceLocation`s for ForgeTags.
```
tterrag1098: This isn't true, in fact it's backwards. `Tags` (`net.minecraftforge.common.tags.Tags`) holds the *forge* tags. Vanilla tags come from `ItemTags`, `BlockTags`, etc. And you would never access tags "through `ResourceLocation`s".

docs/utilities/tags.md
```diff
@@ -1,7 +1,7 @@
 Tags
 ====
-
 Tags are generalized sets of objects in the game, used for grouping related things together and providing fast membership checks.
+Note: Tags have replaced the OreDictionary-- from now on you should (and can only) use tags for all that OreDictionary did. Please see the extra notice at the bottom of this page for more information for transferring from OreDictionary to Tags.
```
tterrag1098: Use the admonitions syntax for notes.

docs/utilities/tags.md
```diff
@@ -21,17 +21,45 @@ Using Tags In Code
...
+// There's two methods to do this, doesn't really matter which you choose:
+// Method 1: from the Item's perspective
+inTag = someItemStack.getItem().isIn(tag);
+// Method 2: from the Tag's perspective
+inTag = tag.contains(someItemStack.getItem());
+```
+
+An example of checking if an Item is contained in a certain Forge `Tag`:
+```Java
+// Create a custom resource to use to fetch a tag instance ('mymod:myitemgroup' is the tag)
```
tterrag1098, line 45: This isn't generally correct, the tag should be a static constant somewhere, like all other tags.

docs/utilities/tags.md
```diff
@@ -21,17 +21,45 @@ Using Tags In Code
...
 ResourceLocation myTagId = new ResourceLocation("mymod", "myitemgroup");
+// Tags work through Items, not ItemStacks so we need to get the Item from our ItemStack
 Item unknownItem = stack.getItem();
-boolean isInGroup = ItemTags.getCollection().getOrCreateTag(myTagId).contains(unknownItem);
-// alternatively, can use getTag and perform a null check
+// This line will either retrieve the Tag (referenced by our ResourceLocation) or create it if it is not
+// in the ItemTags collection of all the tags. Then it will check if our Item (again, not ItemStack)
+// exists in the set of Items that belong to this Tag
+boolean isInGroup = ItemTags.getCollection().getOrCreate(myTagId).contains(unknownItem);
+// Alternatively, you can use the getTag and perform a null check to see if the tag exists and do it that way
```
tterrag1098, line 53: This isn't worth mentioning as it should never be done. There is no point, an empty tag and a non-existent tag are treated the same.

docs/utilities/tags.md
```diff
@@ -21,17 +21,45 @@ Using Tags In Code
...
+Tag retrievedTag = ItemTags.getCollection().get(myTagId);
+if(retrievedTag != null){
+    //do something if retrievedTag.contains(unknownItem) or whatever else you'd like to do with that tag
+}
 ```
 
+There is also `ItemTags.getCollection().getOwningTags(item)` that returns all the `Tag`s that contain an `Item`, however this is generally not the way you'd want to handle tags, it's better if you can check if the `Item` is in a specific `Tag` as opposed to retrieving all its `Tag`s and seeing if the `Tag` is in there.
```
tterrag1098, line 60: This is not correct, `IForgeItem#getTags` should be used for this.

tterrag1098 (changes requested):

tterrag1098:
Overall this seems to add a lot of fluff, I'm not sure what the main goal was. The information is quite correct already, this doesn't really correct any misinformation, it just adds some edge cases that are questionably important enough to be in the docs. I'd suggest slimming it down a bit to what's absolutely necessary - more isn't always better.

tterrag1098:
Also, if this information applies to 1.14, it should be targeted there.

docs/utilities/tags.md
Ktar5: I'm not sure what you mean, this is from the previously existing documentation but with extra comments?

tterrag1098: That doesn't make it correct. If the goal here is to *improve* this page, it should be correct.

PR #321, opened 2020-07-17
Add documentation for Access Transformers

### LINKED ISSUE #87: Request for Documentation for Access Transformers
They're not hard to use and are a lot cleaner than reflection in most cases. Your code will break at compile-time if mojang changes something, which is not the case with reflection.

The old docs are here but contain tons of outdated info
https://web.archive.org/web/20160814191351/http://www.minecraftforge.net/wiki/Using_Access_Transformers

sciwhiz12:
_This PR closes #87._

Adds documentation about Access Transformers. Sources include the [AccessTransformer specification][atspec], the [Java Virtual Machine Specification, SE 8, sections 4.3.2 and 4.3.3][jvmspec], and questions from Discord about [AT paths][atpath] and [AT cautionary advice][atvisibility].

[atspec]: https://github.com/MinecraftForge/AccessTransformers/blob/master/FMLAT.md
[jvmspec]: https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-4.html#jvms-4.3.2
[atpath]: https://discordapp.com/channels/313125603924639766/454376090362970122/733757461550596146
[atvisibility]: https://discordapp.com/channels/313125603924639766/454376090362970122/733781382815285289

### DIFF
--- a/docs/advanced/accesstransformers.md
+++ b/docs/advanced/accesstransformers.md
@@ -1,0 +1,111 @@
+Access Transformers
+===================
+
+Access Transformers (ATs for short) allows for changing the visibility and `final` flags of classes and their methods and fields. They allow modders to access and modify otherwise `private` members in classes outside their control.
+
+The [specification document][specs] can be viewed on the Minecraft Forge GitHub.
+
+!!! note
+    An access transformer preprocessor is required to compile against access-transformed code. ForgeGradle fulfills this purpose for Forge development.
+
+Adding ATs
+----------
+
+Adding an Access Transformer to your mod project is as simple as adding a single line into your `build.gradle`:
+
+```groovy
+// This block is where your mappings version is also specified
+minecraft {
+    accessTransformer = file('src/main/resources/META-INF/accesstransformer.cfg')
+}
+```
+
+During development, the AT file can be anywhere specified by the line above. However, when loading in a non-development environment, Forge will only search for the exact path of `META-INF/accesstransformer.cfg` in your JAR file.
+
+Comments
+--------
+
+All text after a `#` until the end of the line will be treated as a comment and will not be parsed.
+
+Access Modifiers
+----------------
+
+Access modifiers specify which member visibility the given target will be transformed to. In decreasing order of visibility:
+
+  * `public` - visible to all classes inside and outside its package
+  * `protected` - visible only to classes inside the package and subclasses
+  * `default` - visible only to classes inside the package
+  * `private` - visible only to inside the class
+
+A special modifier `+f` and `-f` can be appended to the aforementioned modifiers to either add or remove respectively the `final` modifier, which prevents subclassing, method overriding, or field modification when applied.
+
+!!! warning
+    Access transformers can **_never_** make a member less visible; they can only widen visibility. The JVM will throw an `IllegalAccessError` if any compiled code references a method or field which cannot be accessed due to restrictive transformations.
+    
+    Directives only modify the method they directly reference; any overriding methods will _not_ be access-transformed. Therefore, **never** transform any non-`public` or non-`static` method, because any overriding methods might still use the restrictive visibility, which will result in the JVM throwing a `VerifyError`. 
+
+Targets and Directives
+------------------
+
+!!! Information
+    When using Access Transformers on Minecraft classes, the SRG name must be used for fields and methods.
+
+### Classes
+To target classes:
+```
+<access modifier> <fully qualified class name>
+```
+Inner classes are denoted by combining the fully qualified name of the outer class and the name of the inner class with a `$` as separator.
+
+### Fields
+To target fields:
+```
+<access modifier> <fully qualified class name> <field name>
+```
+
+### Methods
+Targeting methods require a special syntax to denote the method parameters and return type:
+```
+<access modifier> <fully qualified class name> <method name>(<parameter types>)<return type>
+```
+
+#### Specifying Types
+
+Also called "descriptors": see the [Java Virtual Machine Specification, SE 8, sections 4.3.2 and 4.3.3][jvmdescriptors] for more technical details.
+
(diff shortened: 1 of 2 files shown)


Yunus1903 (dismissed):

docs/advanced/accesstransformers.md
```diff
@@ -0,0 +1,111 @@
+Access Transformers
+===================
+
+Access Transformers (ATs for short) allows for changing the visibility and `final` flags of classes and their methods and fields. They allow modders to access and modify otherwise `private` members in classes outside their control.
```
diesieben07: Not just `private`, visibility can be elevated.

docs/advanced/accesstransformers.md
```diff
@@ -0,0 +1,111 @@
...
+
+The [specification document][specs] can be viewed on the Minecraft Forge GitHub.
+
+!!! note
+    An access transformer preprocessor is required to compile against access-transformed code. ForgeGradle fulfills this purpose for Forge development.
```
diesieben07: Is there a reason for this somewhat strange wording? These are docs for Forge, after all.

docs/advanced/accesstransformers.md
```diff
@@ -0,0 +1,111 @@
...
+  * `protected` - visible only to classes inside the package and subclasses
+  * `default` - visible only to classes inside the package
+  * `private` - visible only to inside the class
+
+A special modifier `+f` and `-f` can be appended to the aforementioned modifiers to either add or remove respectively the `final` modifier, which prevents subclassing, method overriding, or field modification when applied.
+
+!!! warning
+    Access transformers can **_never_** make a member less visible; they can only widen visibility. The JVM will throw an `IllegalAccessError` if any compiled code references a method or field which cannot be accessed due to restrictive transformations.
+    
+    Directives only modify the method they directly reference; any overriding methods will _not_ be access-transformed. Therefore, **never** transform any non-`public` or non-`static` method, because any overriding methods might still use the restrictive visibility, which will result in the JVM throwing a `VerifyError`.
```
diesieben07: You can make private methods public to call them, because they cannot be overridden anywhere.

docs/advanced/accesstransformers.md
```diff
@@ -0,0 +1,111 @@
...
+  * `D` - 'double', a double-precision floating-point value
+  * `F` - `float`, a single-precision floating-point value
+  * `I` - `integer`, a 32-bit integer
+  * `J` - `long`, a 64-bit integer
+  * `S` - `short`, a signed short
+  * `Z` - `boolean`, a `true` or `false` value
+  * `[` - references one dimension of an array
+    * Example: `[[S` refers to `short[][]`
+  * `L<class name>;` - references a reference type
+    * Example: `Ljava/lang/String;` refers to `java.lang.String` reference type _(note the use of parentheses)_
```
diesieben07: You mean slashes, not parentheses, right?

diesieben07 (changes requested):

docs/advanced/accesstransformers.md
sciwhiz12: My thought was to make the opening paragraph simple to understand. The second sentence states that it allows access to `private` members, because to modders, this is the most (and often the only) usecase they have. The first sentence already covers any visiblity. Though, reading at it again, I might change 'changing' to 'widening', since 'changing' could mean restricting the visibility.

docs/advanced/accesstransformers.md
XFactHD: Same applies to final methods, no matter the visibility.

docs/advanced/accesstransformers.md
sciwhiz12: Yeah, I was and still am unsure of the wording. The warning was lifted from the FMLAT.md docs. Will reword that.

docs/advanced/accesstransformers.md
sciwhiz12: Ah, must've been confused when writing that piece. WIll correct it.

docs/advanced/accesstransformers.md
sciwhiz12: Yep, will correct that mistake.

docs/advanced/accesstransformers.md
sciwhiz12: > Same applies to final methods, no matter the visibility.

Yeah, I'll need to think through the logic again. Thanks to both of you for correcting this.

docs/advanced/accesstransformers.md
sciwhiz12: I think I'll change it to `otherwise inaccesible`.

docs/advanced/accesstransformers.md
```diff
@@ -0,0 +1,111 @@
...
+
+#### Specifying Types
+
+Also called "descriptors": see the [Java Virtual Machine Specification, SE 8, sections 4.3.2 and 4.3.3][jvmdescriptors] for more technical details.
+
+If the return type is void, or the method has no parameters, then it does not need to be specified.
+
+  * `B` - `byte`, a signed byte
+  * `C` - `char`, a Unicode character code point in UTF-16
+  * `D` - 'double', a double-precision floating-point value
```
TeamSpen210: Any reason this isn't using backticks?

sciwhiz12: No reason, just didn't see it. Thank you.

docs/advanced/accesstransformers.md
sciwhiz12: Actually, I'll just remove that paragraph, since I've now just realized that only Forge has reason to use ATs.

Yunus1903 (commented):
Maybe point out somewhere that gradle needs to be refreshed after adding an AT

tterrag1098 (changes requested):
Looks good, but this needs to target 1.14.x

tterrag1098 (approved):

PR #313, opened 2020-07-02
1.16.x Documentation Start

ChampionAsh5357:
This contains about a 75%-80% update to the current forge docs written for 1.16.x as of update 32.0.38. I accidentally pulled the 1.14.x instead of the 1.15.x for a base so there might be some commits from others or information that I forgot to remove. The main issues such as updating methods and fields to their current counterparts is approximately done say for a few pages which I will discuss later.

I know I am merging this to the 1.15.x docs. There isn't one for 1.16.x as of yet.

**Format Standards**
There was a lot of formatting standards that need to be addressed and probably still do, specifically dealing with the use of `Class/Interface Names`. Some methods are referenced via the class/interface they extended/implemented instead of the class they were implemented (e.g. `IForgeBakedModel` and `IBakedModel`). I would like to know in which cases the method should be referenced from which class/interface.

**Removal**
I have removed a lot of text pertaining to 1.8 and 1.9 mentions as the system prior is almost completely inapplicable in the current standard. There are still a few mentions left which I think should be redone to mention the current system in slightly more depth. All other text removals have to do with a change in implementation.

**Obfuscation**
There are a few obfuscated methods scattered around the documentation due to its recent release. I will update them as new mappings get released.

**BlockState**
Currently, the documentation for BlockState uses DyeColor for a PropertyEnum. As of 1.13, the flattening removed the use of this property since the block registry was now expressively infinite. For this reason, I replaced the use of DyeColor as an example. In regards to the subset example on PropertyEnum, DyeColor still remains the best example for the subset over all current enums. So, I might have to create a specific example to show this change without creating a property that should be expressed as different blocks.

Also, the documentation specifies overwriting getDefaultState(). I do believe this should be specified to calling setDefaultState(baseState) within the constructor to better clarify. I did not change this yet as I wanted the opinion of the reviewer.

**Model**
I could not accurately express the model page when converting from the current language had many mappings  within the current system. I did my best job to try and convert the names successfully. A deep review is needed to make sure there is nothing I missed or need to fix.

**Animation and WorldSavedData**
These two systems have not been well documented since their 1.12 versions. I have done very little editing on them since I still need to verify how to work the current implementation within 1.16. Until then, I will leave them untouched until I am more fluent with them.

Please let me know if there is anything I can do to improve the current condition of the docs. As this version is incomplete, it still provides a better understanding of the current system in place over what the docs has right now.

PR #324, opened 2020-07-19
1.16.x Updates

ChampionAsh5357:
This contains an accurate update to forge docs as of 32.0.70 with mapping snapshot 20200514-1.16. Once again, I know I am merging this to the 1.15.x docs. There isn't one for 1.16.x.

This PR is made to address the issues and recommendations of tterrag1098 in #313. In line with that idea, I have split this PR into the three versions it changes: 1.14.x/1.15.x (#325) and this current one.

This PR only contains 1.16.x specific changes. That's why it may seems as alot of the files still reflect their original counterparts. All the version specific changes are each within their own PR which would combine together to make the original completed version.

Most of the changes in the original PR have been changed in 1.14.x. 1.15.x contained rendering changes and some minor fixes here and there that occurred between 1.14.x and 1.16.x. This PR focuses on the changing of mapping names, the use of AbstractBlock, the replacement of IProperty with Property, and The property override system becoming its own separate client.

To also do a complete match, the example within oak_log.json has been replaced with its data generated form. Since there is no primer written for 1.16 yet, the porting file has been removed until then.

Let me know if there is anything that needs updating or fixing.


docs/legacy/index.md
```diff
@@ -16,3 +16,4 @@ Unfortunately not all versions were used for a significant amount of time and th
 |:-------------:|:----------:|:------------------------------------------|
 |    1.12.x     |   100%     | https://mcforge.readthedocs.io/en/1.12.x/ |
 |    1.13.x     |    10%     | https://mcforge.readthedocs.io/en/1.13.x/ |
+|    1.15.x     |   100%     | https://mcforge.readthedocs.io/en/1.15.x/ |
```
tterrag1098: 1.15.x is not legacy yet.

docs/legacy/porting1214.md
```diff
@@ -1,6 +0,0 @@
-Porting from Minecraft 1.12 to Minecraft 1.13/1.14
-==================================================
-
-Due to major rewrites to Forge happening during the lifetime of 1.13 it never saw much usage and instead most mods ported directly to 1.14 skipping over 1.13. Therefore this guide will focus on porting from 1.12 directly to 1.14.
-
-_Full documentation on porting from 1.12 to 1.14 is still work in progress, for now we suggest having a look at [a primer written by williewillus](https://gist.github.com/williewillus/353c872bcf1a6ace9921189f6100d09a)!_
```
tterrag1098: I don't see the point in removing this without replacement. These guides are still useful to someone coming from versions before 1.15.

docs/models/advanced/ibakedmodel.md
```diff
@@ -22,7 +22,7 @@ If the model is being rendered as an item in an inventory, on the ground as an e
...
 
-When rendering this as an item, returning `true` causes the model to not be rendered, instead falling back to `TileEntityItemStackRenderer::renderItem`. For certain vanilla items such as chests and banners, this method is hardcoded to copy data from the item into a `TileEntity`, before using a `TileEntitySpecialRenderer` to render that TE in place of the item. For all other items, it will use the `TileEntityItemStackRenderer` instance provided by `Item#setTileEntityItemStackRenderer`; refer to [TileEntityItemStackRenderer][teisr] page for more information.
+When rendering this as an item, returning `true` causes the model to not be rendered, instead falling back to `ItemStackTileEntityRenderer::func_239207_a_`. For certain vanilla items such as chests and banners, this method is hardcoded to copy data from the item into a `TileEntity`, before using a `TileEntitySpecialRenderer` to render that TE in place of the item. For all other items, it will use the `TileEntityItemStackRenderer` instance provided by `Item#setTileEntityItemStackRenderer`; refer to [TileEntityItemStackRenderer][teisr] page for more information.
```
tterrag1098: I don't think this change is a good idea, it will be remapped pretty soon and then we'll have to change it back. Let's hold off on baking in new mappings until we have a RB.

docs/models/advanced/itemoverridelist.md
```diff
@@ -11,9 +11,9 @@ Given a list of `ItemOverride`s, the constructor copies that list and stores the
 
 This is a deprecated vanilla method. It is only called in the vanilla `handleItemState`, and in almost all cases can be safely ignored.
 
-### `handleItemState`
+### `func_239290_a_`
```
tterrag1098: Same here

tterrag1098 (changes requested):
I think it might be best to hold off on some of the name changes until 1.16 has stabilized a bit, perhaps this PR can remain open as an omnibus for these kinds of changes until RB is out.

ChampionAsh5357:
Is this also the case for mappings in ItemModelProperties replacing addPropertyOverride? Or should I just leave them be and update when mapped since it does replace the entire system.

tterrag1098:
As 1.16.x now has a RB out, this can be accepted once it's up to date.

ChampionAsh5357:
@tterrag1098 Should be already to go whenever you can. Most of the information that's changed is still present in the other PR I have open, so it might be a good idea to review and merge that first before this one.

tterrag1098 (approved):

PR #327, opened 2020-07-19
Move and update the docs on lifecycle events

sciwhiz12:
Moves the documentation about lifecycle events from [Conventions >> Loading Stages] to [Concepts >> Mod Lifecycle], and updates the references, adds the datagenerators event, `GatherDataEvent`, and removes mention of the configs in `FMLCommonSetupEvent` (proper way is registering to `ConfigTracker` during mod construction).

The move reflects the importance of the lifecycle events in mods, since they are how all mod register their stuff,

_I also removed the reference to commands, since it should be tackled in another article._


MichaelHillcox (dismissed):
All looks good, I think it's a good idea to move the Lifecycle bits into the concepts rather than conventions as it is a core element of modding with forge.

tterrag1098 (changes requested):
This looks fine, but it needs to be rebased onto 1.15.x.

tterrag1098 (approved):

PR #332, opened 2020-08-08
[1.16.x] Add documentation for biomes

thedarkcolour:
Talks about different properties of biomes and how biome ambience works.

### DIFF
--- a/docs/biomes/biomes.md
+++ b/docs/biomes/biomes.md
@@ -1,0 +1,16 @@
+Biomes
+======
+
+Biomes are different regions that appear around the world. They have their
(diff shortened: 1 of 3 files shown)


Yunus1903:
I'd wait a bit with this PR, as biomes are changing a bit in 1.16.2. I don't know how much they will change though other than that they are becoming data-driven.

tterrag1098:
Yes, a lot of this can be adapted but it all needs to be updated to be relevant to 1.16.3.

ChampionAsh5357:
Please update the information in this PR to reflect the current information relating to biomes.
If you do not want to continue this PR, please let us now as soon as possible.

PR #333, opened 2020-08-13
Fix TranslationTextComponent class name

VedaRePowered:
The TranslationTextComponent was referred to as "TextComponentTranslation", which is not the actual type name.


VedaRePowered:
Oops duplicate of #331, sorry.

PR #336, opened 2020-09-25
1.16.x - Add information on TER classes and rendering

robotgryphon:
Adding some information I've dug up from various places on increasing TER render distance as well as how to set up Vertex Builder states.

### DIFF
--- a/docs/tileentities/tesr.md
+++ b/docs/tileentities/tesr.md
@@ -26,3 +26,10 @@ Registering a TER
 -----------------
 
 In order to register a TESR, call `ClientRegistry#bindTileEntitySpecialRenderer` passing the tile entity class to be renderer with this TER and the instance of the TER to use to render all TEs of this class.
+
+Increasing render distance
+--------------------------
+
+For large TESR renderings (multiblock renders, for example) will stop rendering if the primary tile entity goes outside of the player's view (view frustum culling).
+
+In order to address this, change the `AxisAlignedBB` that is returned from `TileEntity#getRenderBoundingBox`. This value should contain your main tile entity bounds, as well as any additional bounds needed to render the full structure.
--- a/docs/tileentities/tesr.md
+++ b/docs/tileentities/tesr.md
@@ -27,6 +27,25 @@ Registering a TER
 
 In order to register a TESR, call `ClientRegistry#bindTileEntitySpecialRenderer` passing the tile entity class to be renderer with this TER and the instance of the TER to use to render all TEs of this class.
 
+IRenderTypeBuffer - Vertex Builders
+-----------------------------------
+This class gives you access to `IRenderTypeBuffer`, which in turn provides a series of `IVertexBuilder` instances via the `IRenderTypeBuffer#getBuffer` method.
+
+To get an example vertex builder (VB), see the `RenderTypes` class for a set of prebuilt builders.
+
+There are various vertex builders that Minecraft provides by default; here is a small overview of the main VB instances:
+
+| Render Type |  Example Purpose
+| :---------- | :--------------
+| SOLID | Rendering block faces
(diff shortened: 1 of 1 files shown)


docs/tileentities/tesr.md
```diff
@@ -26,3 +26,29 @@ Registering a TER
...
+-----------------------------------
+This class gives you access to `IRenderTypeBuffer`, which in turn provides a series of `IVertexBuilder` instances via the `IRenderTypeBuffer#getBuffer` method.
+
+To get an example vertex builder (VB), see the `RenderTypes` class for a set of prebuilt builders.
+
+There are various vertex builders that Minecraft provides by default; here is a small overview of the main VB instances:
+
+| Render Type |  Example Purpose
+| :---------- | :--------------
+| SOLID | Rendering block faces
```
ChampionAsh5357, line 40: As these types are private, it would be better to refer to them by their getters (e.g. `#getSolid`, `#getCutoutMipped`, etc.)

docs/tileentities/tesr.md
```diff
@@ -26,3 +26,29 @@ Registering a TER
...
+| CUTOUT_MIPPED | Rendering parts of a block, or additional, dynamic models
+| CUTOUT | See above, difference is texture sheet usage
+| LINES | Directional lines
```
ChampionAsh5357, line 43: Needs a better description than what's provided. For example, the `LINES` render type can not only draw directional lines, but can also be colored, supports an alpha channel, and targets the item entity framebuffer.

docs/tileentities/tesr.md
```diff
@@ -26,3 +26,29 @@ Registering a TER
...
+
+The best way to get started with these is to look at the `RenderType` class and inspect the various definitions at the top.
+
+If you need a custom VB implementation, the easiest way is to subclass `RenderType` and declare custom instances of `RenderType` in the same manner as defined above. Look at the various builder properties to get a feel for how Minecraft natively sets up a GL state; each method effectively has a setup and teardown function defined.
```
ChampionAsh5357, line 47: `RenderType` _should_ be abstractly subclassed for a custom type as all vanilla defined `RenderState`s are protected. Unless we want to define an entirely new list of possible states, there should be no reason not to subclass it.

Also, it should be mentioned that there should be no direct GL calls when creating the setup and teardown state. They all should be already abstracted away within `RenderSystem`.

docs/tileentities/tesr.md
```diff
@@ -26,3 +26,29 @@ Registering a TER
...
+
+Increasing render distance
```
ChampionAsh5357, line 49: This does not increase the render distance of the tile entity.  Probably more appropriate as `Increasing Visible Scope`

docs/tileentities/tesr.md
```diff
@@ -26,3 +26,29 @@ Registering a TER
...
+--------------------------
+
+For large TESR renderings (multiblock renders, for example) will stop rendering if the primary tile entity goes outside of the player's view (view frustum culling).
```
ChampionAsh5357, line 52: TESR -> TER

An explanation on view frustum culling instead of just a parenthetical citation would give more detail. Also, its the collision box of the block that determines the view frustum.

ChampionAsh5357 (changes requested):
None of this documentation is really specific to the TER. `IRenderTypeBuffer` is a way to abstract away the GL calls and is used in a variety of other places. This would be more appropriate as a separate documentation that outlines everything within the new render system.

This information is also not-specific to 1.16.x and as such, should be rebased and PR-ed into the 1.15.x branch.

sciwhiz12:
Given that the version this PR modifies is of an unsupported version, as well as the age of the PR and its longstanding unaddressed comments, I am closing this PR. Please feel free to open a new PR to a supported version with updated information.

PR #339, opened 2020-10-01
Added recipe condition documentation.

Darkhax:
This PR adds basic info and examples on how to create a new recipe condition. It also adds documentation for the conditions bundled with Forge and basic examples.


TeamSpen210:
I’d suggest swapping the order of the creating/using sections. That way the reader is first taught how the conditions are used, and only then told how to implement one. It might also be a good idea to mention the `forge:conditional` recipe type, which lets you switch between several alternative recipes depending on conditions.

PR #325, opened 2020-07-19
1.14.x -> 1.15.x Update

ChampionAsh5357:
This PR is an overhaul of #313 into 1.14.x as specified by tterag1098. About 90% of the changes are present here. There are very few differences like some formatting fixes. The only major thing of note is commenting out Animation and WorldSavedData within the navigation bar due to its lack of updated information.

The original PR in #322 was pulled incorrectly. Since then, there have been a number of updates including fixes in BlockStates for properties. Any updated changes has been removed and little to no cleanup has been done on those files.

Please feel free to let me know if there is anything I need to add or fix from transferring over.

PR #340, opened 2020-10-03
Updated Item property override in 1.16

PellucidDice:


docs/models/overrides.md
```diff
@@ -37,15 +39,18 @@ The format of an override can be seen on the [wiki][format], and a good example
...
+And here's a hypothetical snippet from the supporting code. Unlike the older versions (lower than 1.16.x), this needs to be done on client side only because ItemModelsProperties does not exist on server.
 
 ```java
-item.addPropertyOverride(new IItemPropertyGetter() {
-  @SideOnly(Side.CLIENT)
-  @Override
-  public float apply(ItemStack stack, @Nullable World world, @Nullable EntityLivingBase entity) {
-    return (float)getPowerLevel(stack) / (float)getMaxPower(stack); // Some external methods
-  }
+private void setup(final FMLCommonSetupEvent event)
```
TeamSpen210: Shouldn't this be `FMLClientSetupEvent`?
```suggestion
private void setup(final FMLClientSetupEvent event)
```

docs/models/overrides.md
```diff
@@ -1,14 +1,16 @@
...
+`ItemModelsProperties::registerGlobalProperty` is used to add a property to all items, it does not take `Item` as it's parameter. `ItemModelsProperties::registerProperty` is used to add a property to a certain item. The `Item` parameter is the item the property is attaching to (e.g. ExampleItems.APPLE). The `ResourceLocation` parameter is the name given to the property (e.g. `new ResourceLocation("pull")`). The `IItemPropertyGetter` is a function that takes the `ItemStack`, the `ClientWorld` it's in (may be null), and the `LivingEntity` that holds it (may be null), returning the `float` value for the property. For modded item properties, it is recommended that the modid of the mod is used as the namespace (e.g. `examplemod:property` and not just `property`, as that really means `minecraft:property`). These should be done in FMLClientSetupEvent.
+!!! important
+    Use DeferredWorkQueue to proceed the task, since the data structures in ItemModelsProperties are not threadsafe.
```
desht: `DeferredWorkQueue` has been replaced by `event.enqueueWork()`

docs/models/overrides.md
```diff
@@ -37,15 +39,18 @@ The format of an override can be seen on the [wiki][format], and a good example
...
 ```java
-item.addPropertyOverride(new IItemPropertyGetter() {
-  @SideOnly(Side.CLIENT)
-  @Override
-  public float apply(ItemStack stack, @Nullable World world, @Nullable EntityLivingBase entity) {
-    return (float)getPowerLevel(stack) / (float)getMaxPower(stack); // Some external methods
-  }
+private void setup(final FMLClientSetupEvent event)
+{
+  DeferredWorkQueue.runLater(() ->
```
desht: As above; use `event.enqueueWork()` here

docs/models/overrides.md
```diff
@@ -37,15 +39,18 @@ The format of an override can be seen on the [wiki][format], and a good example
...
+  {
+    ItemModelsProperties.func_239418_a_(ExampleItems.APPLE,
```
desht: this is `registerProperty()` in recent mappings

PellucidDice:
Thank you for pointing these out.

docs/models/overrides.md
```diff
@@ -1,14 +1,16 @@
...
+An item property assigns a certain `float` value to every `ItemStack` it is registered for, and vanilla item model definitions can use these values to define "overrides", where an item defaults to a certain model, but if an override matches, it overrides the model and uses another. They are useful mainly because they are continuous. For example, bows use item properties to define their pull animation. Since the value of the property is a `float`, it increases continuously from 0 to 1. This allows resource packs to add as many models as they want for the bow pulling animation along that spectrum, instead of being stuck with four "slots" for their models in the animation. The same is true of the compass and clock.
```
sciwhiz12: Could you add some clarification/explicit info. that the `float` value is not limited to `0.0`-`1.0`, and can be any value in the range of a `float`?
> Since the value of the property is a `float`, it increases continuously from 0 to 1.

This implies that the value is restricted to `0.0`-`1.0`, which can be confusing when encountering this for the first time.

docs/models/overrides.md
```diff
@@ -1,14 +1,16 @@
...
+`ItemModelsProperties::registerGlobalProperty` is used to add a property to all items, it does not take `Item` as it's parameter. `ItemModelsProperties::registerProperty` is used to add a property to a certain item. The `Item` parameter is the item the property is attaching to (e.g. ExampleItems.APPLE). The `ResourceLocation` parameter is the name given to the property (e.g. `new ResourceLocation("pull")`). The `IItemPropertyGetter` is a function that takes the `ItemStack`, the `ClientWorld` it's in (may be null), and the `LivingEntity` that holds it (may be null), returning the `float` value for the property. For modded item properties, it is recommended that the modid of the mod is used as the namespace (e.g. `examplemod:property` and not just `property`, as that really means `minecraft:property`). These should be done in FMLClientSetupEvent.
```
sciwhiz12: > `ItemModelsProperties::registerGlobalProperty` is used to add a property to all items, it does not take `Item` as it's parameter. 

The method is by default `private static`. Should either be removed, or moved to the bottom and clarified that it is `private`.

> The `IItemPropertyGetter` is a function that **...**
It's more of a functional interface. Might want to mention that too.

docs/models/overrides.md
```diff
@@ -37,15 +39,18 @@ The format of an override can be seen on the [wiki][format], and a good example
...
-  @Override
-  public float apply(ItemStack stack, @Nullable World world, @Nullable EntityLivingBase entity) {
-    return (float)getPowerLevel(stack) / (float)getMaxPower(stack); // Some external methods
-  }
+private void setup(final FMLClientSetupEvent event)
+{
+  event.enqueueWork(() ->
+  {
+    ItemModelsProperties.registerProperty(ExampleItems.APPLE, 
+      new ResourceLocation(ExampleMod.MODID, pulling"), (stack, world, living) -> {
```
sciwhiz12: > ```java
>     new ResourceLocation(ExampleMod.MODID, pulling"), (stack, world, living) -> {
> ```
Missing starting `"` for the string .

sciwhiz12 (changes requested):
A few changes are in order, then this'll be ready.

If you have the time too, could you also add to the `Using Overrides` section that if multiple overrides match, the one nearest to the end of the list is used.

docs/models/overrides.md
```diff
@@ -1,19 +1,22 @@
...
+`ItemModelsProperties::registerProperty` is used to add a property to a certain item. The `Item` parameter is the item the property is attaching to (e.g. ExampleItems.APPLE). The `ResourceLocation` parameter is the name given to the property (e.g. `new ResourceLocation("pull")`). The `IItemPropertyGetter` is a functional interface that takes the `ItemStack`, the `ClientWorld` it's in (may be null), and the `LivingEntity` that holds it (may be null), returning the `float` value for the property. For modded item properties. It is recommended that the modid of the mod is used as the namespace (e.g. `examplemod:property` and not just `property`, as that really means `minecraft:property`). These should be done in FMLClientSetupEvent.
+There's also another private method `ItemModelsProperties::registerGlobalProperty` that is used to add properties to all items, and it does not take `Item` as it's parameter since all items will apply this property.
```
sciwhiz12: Minor nitpick: 
```suggestion
There's also another private method `ItemModelsProperties::registerGlobalProperty` that is used to add properties to all items, and it does not take `Item` as its parameter since all items will apply this property.
```

docs/models/overrides.md
```diff
@@ -1,19 +1,22 @@
...
+`ItemModelsProperties::registerProperty` is used to add a property to a certain item. The `Item` parameter is the item the property is attaching to (e.g. ExampleItems.APPLE). The `ResourceLocation` parameter is the name given to the property (e.g. `new ResourceLocation("pull")`). The `IItemPropertyGetter` is a functional interface that takes the `ItemStack`, the `ClientWorld` it's in (may be null), and the `LivingEntity` that holds it (may be null), returning the `float` value for the property. For modded item properties. It is recommended that the modid of the mod is used as the namespace (e.g. `examplemod:property` and not just `property`, as that really means `minecraft:property`). These should be done in FMLClientSetupEvent.
```
sciwhiz12: Minor nitpick:
```suggestion
`ItemModelsProperties::registerProperty` is used to add a property to a certain item. The `Item` parameter is the item the property is attaching to (e.g. ExampleItems.APPLE). The `ResourceLocation` parameter is the name given to the property (e.g. `new ResourceLocation("pull")`). The `IItemPropertyGetter` is a functional interface that takes the `ItemStack`, the `ClientWorld` it's in (may be null), and the `LivingEntity` that holds it (may be null), returning the `float` value for the property. For modded item properties, it is recommended that the modid of the mod is used as the namespace (e.g. `examplemod:property` and not just `property`, as that really means `minecraft:property`). These should be done in FMLClientSetupEvent.
```

sciwhiz12 (changes requested):
Just two minor nitpicky changes, then it's okay to me.

PellucidDice:
Thank you for reviewing.

sciwhiz12 (dismissed):
I will note that #325 does touch the same file, so there's a possibility that we'll need to have this rebased if that PR gets merged in first, if there's a merge conflict.

docs/models/overrides.md
```diff
@@ -1,19 +1,22 @@
...
+There's also another private method `ItemModelsProperties::registerGlobalProperty` that is used to add properties to all items, and it does not take `Item` as its parameter since all items will apply this property.
+!!! important
+    Use FMLClientSetupEvent::enqueueWork to proceed with the tasks, since the data structures in ItemModelsProperties are not threadsafe.
 
 Using Overrides
 ---------------
 
-The format of an override can be seen on the [wiki][format], and a good example can be found in `model/item/bow.json`. For reference, here is a hypothetical example of an item with an `examplemod:power` property. If the values have no match, the default is the current model.
+The format of an override can be seen on the [wiki][format], and a good example can be found in `model/item/bow.json`. For reference, here is a hypothetical example of an item with an `examplemod:power` property. If the values have no match, the default is the current model, but if there are multiple matches the predicate that is nearest to the end of the list will be used.
```
tterrag1098: This addition is a bit wordy.
```suggestion
The format of an override can be seen on the [wiki][format], and a good example can be found in `model/item/bow.json`. For reference, here is a hypothetical example of an item with an `examplemod:power` property. If the values have no match, the default is the current model, but if there are multiple matches last match in the list will be selected.
```

docs/models/overrides.md
```diff
@@ -1,19 +1,22 @@
...
+An item property assigns a certain `float` value to every `ItemStack` it is registered for, and vanilla item model definitions can use these values to define "overrides", where an item defaults to a certain model, but if an override matches, it overrides the model and uses another. They are useful mainly because they are continuous. For example, bows use item properties to define their pull animation. The item models are decided by the 'float' number predicates, it is not limited but in this case we use values between 0.0F and 1.0F as the example. This allows resource packs to add as many models as they want for the bow pulling animation along that spectrum, instead of being stuck with four "slots" for their models in the animation. The same is true of the compass and clock.
```
tterrag1098: In *this* case? Which case? This should probably just say "is not limited but generally between..."
```suggestion
An item property assigns a certain `float` value to every `ItemStack` it is registered for, and vanilla item model definitions can use these values to define "overrides", where an item defaults to a certain model, but if an override matches, it overrides the model and uses another. They are useful mainly because they are continuous. For example, bows use item properties to define their pull animation. The item models are decided by the 'float' number predicates, it is not limited but generally between `0.0F` and `1.0F`. This allows resource packs to add as many models as they want for the bow pulling animation along that spectrum, instead of being stuck with four "slots" for their models in the animation. The same is true of the compass and clock.
```

docs/models/overrides.md
```diff
@@ -1,19 +1,22 @@
...
+`ItemModelsProperties::registerProperty` is used to add a property to a certain item. The `Item` parameter is the item the property is attaching to (e.g. ExampleItems.APPLE). The `ResourceLocation` parameter is the name given to the property (e.g. `new ResourceLocation("pull")`). The `IItemPropertyGetter` is a functional interface that takes the `ItemStack`, the `ClientWorld` it's in (may be null), and the `LivingEntity` that holds it (may be null), returning the `float` value for the property. For modded item properties, it is recommended that the modid of the mod is used as the namespace (e.g. `examplemod:property` and not just `property`, as that really means `minecraft:property`). These should be done in FMLClientSetupEvent.
+There's also another private method `ItemModelsProperties::registerGlobalProperty` that is used to add properties to all items, and it does not take `Item` as its parameter since all items will apply this property.
+!!! important
```
tterrag1098, line 14: Blank line before an admonition

tterrag1098 (dismissed):

docs/models/overrides.md
PellucidDice: It's using vanilla bow as the example.

docs/models/overrides.md
```diff
@@ -1,19 +1,23 @@
...
+There's also another private method `ItemModelsProperties::registerGlobalProperty` that is used to add properties to all items, and it does not take `Item` as its parameter since all items will apply this property.
+
+!!! important
+    Use FMLClientSetupEvent::enqueueWork to proceed with the tasks, since the data structures in ItemModelsProperties are not threadsafe.
 
 Using Overrides
 ---------------
 
-The format of an override can be seen on the [wiki][format], and a good example can be found in `model/item/bow.json`. For reference, here is a hypothetical example of an item with an `examplemod:power` property. If the values have no match, the default is the current model.
+The format of an override can be seen on the [wiki][format], and a good example can be found in `model/item/bow.json`. For reference, here is a hypothetical example of an item with an `examplemod:power` property. If the values have no match, the default is the current model, but if there are multiple matches last match in the list will be selected.
```
sciwhiz12: Minor adjustment:
```suggestion
The format of an override can be seen on the [wiki][format], and a good example can be found in `model/item/bow.json`. For reference, here is a hypothetical example of an item with an `examplemod:power` property. If the values have no match, the default is the current model, but if there are multiple matches, the last match in the list will be selected.
```

docs/models/overrides.md
```diff
@@ -37,15 +41,18 @@ The format of an override can be seen on the [wiki][format], and a good example
 }
 ```
 
-And here's a hypothetical snippet from the supporting code. (This does not have to be client-only; it will work on a server too. In vanilla, properties are registered in the item's constructor.)
+And here's a hypothetical snippet from the supporting code. Unlike the older versions (lower than 1.16.x), this needs to be done on client side only because ItemModelsProperties does not exist on server.
```
sciwhiz12: Another minor formatting adjustment:
```suggestion
And here's a hypothetical snippet from the supporting code. Unlike the older versions (lower than 1.16.x), this needs to be done on client side only because `ItemModelsProperties` does not exist on server.
```

sciwhiz12 (changes requested):
Just two more minor formatting changes, then it should be fine now.

PellucidDice:
How would I go about with the conflicts?

ChampionAsh5357:
You can rebase your branch on top of the 1.16.x upstream or merge the 1.16.x upstream with your current branch. Essentially, the file you are using is outdated as a current update to the branch modified the file in some way. If you happen to modify a portion of the file that was changed, you get a conflict. This will explain how to do this within the [command line](https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/resolving-a-merge-conflict-using-the-command-line). If you're using Github Desktop, it should pop up with a notification when trying to rebase/merge that will allow you to resolve the files in your selected text editor.

PellucidDice:
Sorry for my unclear question, what I meant to ask was which content you think is better/easier for the users? I'm not quite sure should I override the conflict section with his/her content, or override with mine.

ChampionAsh5357:
#324 was written back for base 1.16 with relatively few changes to the source and rather updating method calls that were used within the new system. As such, it did not aim to expand or clarify some details, just to make the information being presented accurate. Since your content expands upon the information already available and clarifies some resources in the example source, I would say overriding the current changes is fine.

sciwhiz12 (approved):

TheCurle (approved):

PR #347, opened 2020-11-08
Fix documentation for obstinate modders to more easily follow CPW rec…

JaisDK:
…ommended practices and hopefully prevent 1.16 from being another shitshow/failure.

### DIFF
--- a/docs/gettingstarted/structuring.md
+++ b/docs/gettingstarted/structuring.md
@@ -24,8 +24,8 @@ The `mods.toml` file is formatted as [TOML](https://github.com/toml-lang/toml), 
     # The name of the mod loader type to load - for regular FML @Mod mods it should be javafml
     modLoader="javafml"
     # A version range to match for said mod loader - for regular FML @Mod it will be the forge version
-    # Forge for 1.15.2 is version 31
-    loaderVersion="[31,)"
+    # Forge for 1.16.3 is version 34*. It is recommended to leave a comma at the end to support future versions
+    loaderVersion="[34,)"
     # A URL to refer people to when problems occur with this mod
     issueTrackerURL="github.com/MinecraftForge/MinecraftForge/issues"
     # If the mods defined in this file should show as seperate resource packs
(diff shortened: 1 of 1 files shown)


docs/gettingstarted/structuring.md
```diff
@@ -24,8 +24,8 @@ The `mods.toml` file is formatted as [TOML](https://github.com/toml-lang/toml),
     # The name of the mod loader type to load - for regular FML @Mod mods it should be javafml
     modLoader="javafml"
     # A version range to match for said mod loader - for regular FML @Mod it will be the forge version
-    # Forge for 1.15.2 is version 31
-    loaderVersion="[31,)"
+    # Forge for 1.16.3 is version 34*. It is recommended to leave a comma at the end to support future versions
```
tterrag1098, line 27: Why the random asterisk?
```suggestion
    # Forge for 1.16.3 is version 34. It is recommended to leave a comma at the end to support future versions
```

JaisDK: It is to point the reader to the note below `* All version ranges use...`

gigaherz: No idea how markdown/rtd deals with symbols but would a little superscript ¹ do the job there? Assuming there's no more superscripts in the page. Otherwise ²³⁴⁵⁶⁷⁸⁹⁰ are also available :P

tterrag1098: No, that's just confusing. This is a code example. If the note below is important it should be in an admonition, not using custom HTML.

docs/gettingstarted/structuring.md
```diff
@@ -47,18 +47,20 @@ The `mods.toml` file is formatted as [TOML](https://github.com/toml-lang/toml),
...
-        versionRange="[31,)"
+        versionRange="[34,)"
         ordering="NONE"
         side="BOTH"
 
       [[dependencies.examplemod]]
         modId="minecraft"
         mandatory=true
-        versionRange="[1.15.2]"
+        versionRange="[1.16.3,]"
```
pupnewfster, line 57: Shouldn't this be like cpw made the mdk, and end in `,1.17)`

JaisDK: In his tweet he specifically mentioned having the trailing slash, so I don't know why the mdk has it differently. The result is the same right?

PR #348, opened 2020-11-12
Removed MapStorage stuff since its outdated

Joshi234:


ChampionAsh5357:
This does not update the information relating to the documentation. `WorldSavedData` still exists and is relevant, it just needs to be updated to reflect the new information. I will close this PR for now as it seems there is no intention to rewrite this section.

PR #349, opened 2020-11-19
Distinguish MDK from source distribution

jpaugh:
Also added to the latest branch (1.16.x). See PR #350

### DIFF
--- a/docs/gettingstarted/index.md
+++ b/docs/gettingstarted/index.md
@@ -6,7 +6,7 @@ This is a simple guide to get you from nothing to a basic mod. The rest of this 
 From Zero to Modding
 --------------------
 
-1. Obtain the MCForge Development Kit (MDK) from forge's [files][] site. (Look for the MDK file type)
+1. Obtain the Mod Development Kit (MDK) from forge's [files][] site. (Look for the MDK file type)
 2. Extract the downloaded MDK an empty directory. You should see a bunch of files, and an example mod is placed in `src/main/java` for you to look at. Only a few of these files are strictly necessary for mod development, and you may reuse these files for all your projects These files are:
     * `build.gradle`
     * `gradlew.bat`
--- a/docs/gettingstarted/index.md
+++ b/docs/gettingstarted/index.md
@@ -6,8 +6,8 @@ This is a simple guide to get you from nothing to a basic mod. The rest of this 
 From Zero to Modding
 --------------------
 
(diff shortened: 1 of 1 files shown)


docs/gettingstarted/index.md
```diff
@@ -6,8 +6,8 @@ This is a simple guide to get you from nothing to a basic mod. The rest of this
 From Zero to Modding
 --------------------
 
-1. Obtain a source distribution from forge's [files][] site. (Look for the MDK file type)
-2. Extract the downloaded source distribution to an empty directory. You should see a bunch of files, and an example mod is placed in `src/main/java` for you to look at. Only a few of these files are strictly necessary for mod development, and you may reuse these files for all your projects These files are:
+1. Obtain the MCForge Development Kit (MDK) from forge's [files][] site. (Look for the MDK file type)
```
LexManos: Mod Development Kit

jpaugh: You're right! That reads way better! I've updated both PRs to fix this.

docs/gettingstarted/index.md
```diff
@@ -6,8 +6,8 @@ This is a simple guide to get you from nothing to a basic mod. The rest of this
 From Zero to Modding
 --------------------
 
-1. Obtain a source distribution from forge's [files][] site. (Look for the MDK file type)
-2. Extract the downloaded source distribution to an empty directory. You should see a bunch of files, and an example mod is placed in `src/main/java` for you to look at. Only a few of these files are strictly necessary for mod development, and you may reuse these files for all your projects These files are:
+1. Obtain the Mod Development Kit (MDK) from forge's [files][] site. (Look for the MDK file type)
+2. Extract the downloaded MDK an empty directory. You should see a bunch of files, and an example mod is placed in `src/main/java` for you to look at. Only a few of these files are strictly necessary for mod development, and you may reuse these files for all your projects These files are:
```
sciwhiz12: Missing a `to` there.
```suggestion
2. Extract the downloaded MDK to an empty directory. You should see a bunch of files, and an example mod is placed in `src/main/java` for you to look at. Only a few of these files are strictly necessary for mod development, and you may reuse these files for all your projects These files are:
```

sciwhiz12 (changes requested):
One small minor correction to the grammar, then it looks good to me.

TheCurle (approved):
I wonder how long this has been wrong.

PR #350, opened 2020-11-19
Distinguish MDK from source distribution

jpaugh:
Also added to 1.15.x branch as a hotfix. See PR #349


sciwhiz12:
The main branch of the documentation is 1.15.x, and , and only PRs that add 1.16-specific content should be PR-ed specifically to 1.16.x.
Closing this in favor of #349.

PR #355, opened 2021-01-16
Fix path of example logoFile in mods.toml

ravenbuilder934:
Right now, the logoFile line in the example mods.toml on the Structuring page reads `      logoFile="assets/examplemod/textures/logo.png"`. However, later in the page it says the logofile "must be placed in the root resource folder, not in a subfolder." I confirmed that if you put a logofile in the mods.toml with a path, it crashes. So this PR changes the logoFile line in the example mods.toml to read `logoFile="logo.png"`.


TheCurle:
If you're going to touch this file, at least add information about the license field, since it's required and not mentioned ANYWHERE.

PR #357, opened 2021-02-07
Update index.md

EvanMachusak:
Added note about ensuring you're using a 64-bit JVM.

### DIFF
--- a/docs/gettingstarted/index.md
+++ b/docs/gettingstarted/index.md
@@ -6,6 +6,8 @@ This is a simple guide to get you from nothing to a basic mod. The rest of this 
 From Zero to Modding
 --------------------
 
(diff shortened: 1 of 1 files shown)


ChampionAsh5357:
I would hold off on this until we have an omnibus PR as this only changes one line.

Also, it would be better to specify a Java 8 jdk with a 64-bit jvm which can be obtained from https://adoptopenjdk.net/. This is not to mention the command specified does not support all usecases. Any default path that uses Java 10+ will error when that command is used as it no longer exists. Using `java -version` isn't the best way to check if Java 8 is installed anyways since it depends on the first jdk instance mentioned on the path which does not need to be Java 8. It's possible to have Java 8 run in development and for gradle tasks without having it be the jdk version used to run applications.

tterrag1098:
This needs to be manually pulled into the standing Omnibus PR.

ChampionAsh5357:
Merged in the omnibus #364.

PR #368, opened 2021-04-08
Removed misplaced colon

RedNicStone:
Removed a misplaced colon causing the script to violate json syntax.


ChampionAsh5357:
This should be PRed into the omnibus.

TheCurle:
The Omnibus has an equivalent modification, so this should be closed.

PR #374, opened 2021-05-04
1.16.x

icedchai:
documentation for forge tags


TheCurle:
There's a) almost no usable information here, and b) no way to actually get to this page.. 
Consider putting some information on what these tags actually do, rather than just saying "they're important. use them".

Additionally, linking out to some random file on a random repository from a random user is poor taste.
The Forge repository has the full list of tags, as we're the ones that register them...

icedchai:
How should I make it accessible? I already have it in a folder

TheCurle:
Nothing links to it.

icedchai:
i see

TheCurle:
Please stop spamming open/close..

docs/forgetags/tags.md
```diff
@@ -0,0 +1,43 @@
...
+    "replace": false,//this means that you will add to the tag, instead of overriding it completely
+    "values": [
+      "MODID:ruby"
+    ]
+  }
+```
+To add these tags to your own recipes, you should go to the directory data/MODID/recipes. Inside it, create a recipe named ruby_sword.json, and write inside it:
+```
+{
+  "type": "minecraft:crafting_shaped",//this means the crafting will take on a specific pattern
```
TheCurle, line 20: The tag parser doesn't use lenient mode by default, this is invalid json

ChampionAsh5357:
In my opinion, this shouldn't be separate from the tags page already existing within the docs. Not to mention, everything in this page described so far applies to both forge and vanilla tags. There is no difference being exemplified here and seems like a tutorial rather than actual documentation.

Questions to ask yourself:
1. What point are you trying to get across?
2. Has this point been explained already?
3. Does the documentation, if already present, provide an example?
4. Does it generalize such that it can be applied in multiple scenarios/use cases?
5. Is the information accurate?

There are probably other questions to ask, but these five are a good starting point.

TheCurle:
This change hasn't been justified or updated since we last got in touch, so i'm gonna close this.

PR #392, opened 2021-10-16
[1.16] Improve TileEntityType Explanation

ChampionAsh5357:
Originally, `TileEntityType`s were briefly glanced over as the object for registration of a `TileEntity`. While this may be the case, the explanation provided for understanding was lacking in clarity. This more clearly explains each of the methods used to create a `TileEntityType` and what it is used for.

### DIFF
--- a/docs/tileentities/tileentity.md
+++ b/docs/tileentities/tileentity.md
@@ -13,7 +13,7 @@ More advanced examples exist in mods, such as quarries, sorting machines, pipes,
 
 Tile Entities are created and removed dynamically and as such are not registry objects on their own. 
 
-In order to create a `TileEntity`, you need to extend the `TileEntity` class. As such, a factory type is registered instead to easily create the dynamic object. For a `TileEntity`, these are known as `TileEntityType`s.
+In order to create a `TileEntity`, you need to extend the `TileEntity` class. As such, another object is registered instead to easily create and refer to the *type* of the dynamic object. For a `TileEntity`, these are known as `TileEntityType`s.
 
 A `TileEntityType` can be [registered][registration] like any other registry object. To construct a `TileEntityType`, its builder form can be used via `TileEntityType$Builder#of`. This takes in two arguments: a `factory` representing some supplied method to create a new instance of the associated `TileEntity`, and a varargs of `validBlocks` which this `TileEntity` can be attached to. Building the `TileEntityType` is done by calling `TileEntityType$Builder#build`. This takes in a `Type` which represents the type-safe reference used to refer to this registry object in a `DataFixer`. Since `DataFixer`s are an optional system to use for mods, this can be passed as `null`.
 
--- a/docs/tileentities/tileentity.md
+++ b/docs/tileentities/tileentity.md
@@ -9,23 +9,22 @@ More advanced examples exist in mods, such as quarries, sorting machines, pipes,
     `TileEntities` aren't a solution for everything and they can cause lag when used wrongly.
     When possible, try to avoid them.
 
+## Registering
+
+Tile Entities are created and removed dynamically and as such are not registry objects on their own. 
+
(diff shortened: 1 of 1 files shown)


docs/tileentities/tileentity.md
```diff
@@ -9,23 +9,22 @@ More advanced examples exist in mods, such as quarries, sorting machines, pipes,
...
     When possible, try to avoid them.
 
-## Creating a `TileEntity`
+## Registering
+
+Tile Entities are created and removed dynamically and as such are not registry objects on their own. 
+
+In order to create a `TileEntity`, you need to extend the `TileEntity` class. As such, another object is registered instead to easily create and refer to the *type* of the dynamic object. For a `TileEntity`, these are known as `TileEntityType`s.
+
+A `TileEntityType` can be [registered][registration] like any other registry object. To construct a `TileEntityType`, its builder form can be used via `TileEntityType$Builder#of`. This takes in two arguments: a `factory` representing some supplied method to create a new instance of the associated `TileEntity`, and a varargs of `validBlocks` which this `TileEntity` can be attached to. Building the `TileEntityType` is done by calling `TileEntityType$Builder#build`. This takes in a `Type` which represents the type-safe reference used to refer to this registry object in a `DataFixer`. Since `DataFixer`s are an optional system to use for mods, this can be passed as `null`.
```
sciwhiz12, line 18: I have to admit, I'm a bit wary of using the `$` syntax to refer to inner classes, given that it's not used in documentation (but rather, by Java compiled code as a convention to my knowledge).

docs/tileentities/tileentity.md
```diff
@@ -9,23 +9,22 @@ More advanced examples exist in mods, such as quarries, sorting machines, pipes,
...
-To register it, listen for the appropriate registry event and create a `TileEntityType`:
 ```java
-@SubscribeEvent
-public static void registerTE(RegistryEvent.Register<TileEntityType<?>> evt) {
-  TileEntityType<?> type = TileEntityType.Builder.of(factory, validBlocks).build(null);
-  type.setRegistryName("mymod", "myte");
-  evt.getRegistry().register(type);
-}
+// For some DeferredRegister<TileEntityType<?>> REGISTER
+public static final RegistryObject<TileEntityType<MyTE>> MY_TE = REGISTER.register("myte", () -> TileEntityType.Builder.of(factory, validBlocks).build(null));
```
sciwhiz12, line 22: Could we have an admonition note to inform why a `null` is passed to the `#build` function?

docs/tileentities/tileentity.md
```diff
@@ -9,23 +9,22 @@ More advanced examples exist in mods, such as quarries, sorting machines, pipes,
...
 ```
-In this example, `factory` is a function that creates a new instance of your TileEntity. A method reference or a lambda is commonly used. The variable `validBlocks` is one or more blocks (`TileEntityType$Builder#of` is varargs) that the tile entity can exist for.
 
-## Attaching a `TileEntity` to a `Block`
+## Creating a `TileEntity`
 
-To attach your new `TileEntity` to a `Block`, you need to override 2 (two) methods within your `Block` subclass:
+To create a `TileEntity` and attach it to a `Block`, you need to override 2 (two) methods within your `Block` subclass:
```
sciwhiz12: Just something which I feel like should be done since this line is already being modified: interchange the `2` and `two`, so that part reads `[...] you need to override two (2) methods [...]` (or perhaps even omit the numeral).

docs/tileentities/tileentity.md
ChampionAsh5357: True, but that's the style guide I created for the docs 6 months ago. So, that would be more of a formatting change across the entire doc.

docs/tileentities/tileentity.md
ChampionAsh5357: Read the last two lines of the explanation above the example. Should I move that down instead?

TheCurle (approved):

PR #400, opened 2022-01-20
Change WorldSavedData to SavedData for 1.18.x

D0CTOR-ZED:
If I understand correctly, the class has a new name with the new mappings being used for 1.18.x.


ChampionAsh5357 (approved):
Seems to be just something missed when updating to 1.17, apologies.

TheCurle (approved):

PR #409, opened 2022-02-12
Add info for registering to vanilla registries

eerussianguy:
I see this question come up relatively often in the discord, and the errors produced aren't always instructive for what you need to do. There are quite a few registries that are not forge registries so it's useful to have something to point people to.

Some big examples of this are recipe types, placement modifier types, configured and placed features, and block predicates.

### DIFF
--- a/docs/concepts/registries.md
+++ b/docs/concepts/registries.md
@@ -10,7 +10,7 @@ Every type of registrable object has its own registry. To see all registries sup
 Methods for Registering
 ------------------
 
-There are three proper ways to register objects: the `DeferredRegister` class, the `RegistryEvent$Register` lifecycle event, as well as `Registry#register` for vanilla-only registries.
+There are two proper ways to register objects: the `DeferredRegister` class, and the `RegistryEvent$Register` lifecycle event.
 
 ### DeferredRegister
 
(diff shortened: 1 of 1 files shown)


docs/concepts/registries.md
```diff
@@ -43,6 +43,37 @@ public void registerBlocks(RegistryEvent.Register<Block> event) {
...
 
+### Registries that aren't Forge Registries
+
+Due to some peculiarities of vanilla code, not all registries are wrapped by Forge. One of the more common examples of this is with some worldgen registries, such as configured features and placed features.
+
+It is safe to use these registries in your mods when a forge variant isn't available. However, without the benefit of registry objects, we have to be careful of when we do the registration and how.
+
+Here is an example. Here, we just use vanilla's method `Registry#register` while being sure to populate it at a later date.
+
+```java
```
TheCurle: Can you explain this without copypasta code? 

It's a giant pain for us receiving support requests, and having to maintain this snippet when the internals of vanilla code changes.

docs/concepts/registries.md
```diff
@@ -10,7 +10,7 @@ Every type of registrable object has its own registry. To see all registries sup
 Methods for Registering
 ------------------
 
-There are two proper ways to register objects: the `DeferredRegister` class, and the `RegistryEvent$Register` lifecycle event.
+There are three proper ways to register objects: the `DeferredRegister` class, the `RegistryEvent$Register` lifecycle event, as well as `Registry#register` for vanilla-only registries.
```
ChampionAsh5357: This should really not be a combined statement. First, a list followed by a context means that all elements in the list are in relation to said context. Second, these two statements should be separated. Vanilla-only registries is a special case when no Forge wrapped registry is present. There are a few reasons why this may happen, but in general it should be its own section and not wrapped.

docs/concepts/registries.md
eerussianguy: Fine with me, though I would argue that this is a relatively stable way of doing things

ChampionAsh5357 (changes requested):
Since Curle already took my other comment, I will talk about something else.

There is some nuances of vanilla registries that are left out here. Specifically, this tends to lump all non-Forge wrapped registries into vanilla, which is not the case. Vanilla registries are broken into static and dynamic registries as well.

If you would like to mention vanilla registries in general, you need to separate these two cases and explain its appropriateness. For example, dynamic registries which include most world gen registries should not be registered in code unless a pre-existing registry object requires them. Additionally, dynamic registry objects in code should only be a dummy instance with their actual object being supplied within its serialized form for overriding.

docs/concepts/registries.md
```diff
@@ -43,6 +43,15 @@ public void registerBlocks(RegistryEvent.Register<Block> event) {
 }
 ```
 
+### Registries that aren't Forge Registries
+
+Due to some peculiarities of vanilla code, not all registries are wrapped by Forge. These can be static registries, like `RecipeType`, which are safe to use. There are also dynamic registries, like `ConfiguredFeature` and some other worldgen registries, which are typically represented in JSON. These objects should only be registered this way if there is another registry object that requires it. 
+
+Without the benefit of registry objects, we have to be careful of when we do the registration and how.
```
ChampionAsh5357: Try to avoid using first person pronouns in sentences. This is not heavily enforced, but I would eventually like to transition the documentation away from making it seem like a narrative banter.

docs/concepts/registries.md
```diff
@@ -43,6 +43,15 @@ public void registerBlocks(RegistryEvent.Register<Block> event) {
...
+
+Outside your main mod class, have a static field store a `Lazy` of a call to `Registry#register` containing the object you want to register. Later, in FMLCommonSetupEvent, enqueue a call to `Supplier#get` on the object, thus populating it.
```
ChampionAsh5357: What is a `Lazy`? Why must the field be within a `Lazy`? What does `Registry#register` take in ? Why must this be done in the common setup event? How did `Lazy` become a `Supplier#get`?

Telling someone to use a particular piece of code without some explanation as to reasonings behind it make it difficult for people to understand the specifications and how to adapt them.

docs/concepts/registries.md
```diff
@@ -43,6 +43,15 @@ public void registerBlocks(RegistryEvent.Register<Block> event) {
...
+
+
```
ChampionAsh5357, line 55: Extra newline.

ChampionAsh5357 (changes requested):
Mainly just want some more explanation on the usage sentence. It seems a bit too barebones to get the importance of its meaning across.

ChampionAsh5357 (approved):

sciwhiz12 (approved):

TheCurle (approved):

PR #411, opened 2022-02-16
[1.18.x] Overhaul Global Loot Modifiers

ChampionAsh5357:
Global loot modifiers are structured as a tutorial around the first test case ever implemented in Forge for GLMs. As such, it does not provide a 'documentation'-like information point as to the meanings behind certain classes and what do they do. It additionally provides copy-paste code which shouldn't need to be maintained.

This removes the example code and replaces it with pseudocode, along with outlining all processes of a modder's implementation of a GLM.


docs/items/globallootmodifiers.md
```diff
@@ -7,133 +7,142 @@ Registering a Global Loot Modifier
 -------------------------------
 
 You will need 4 things:
-1. Create a `global_loot_modifiers.json` file at `/data/forge/loot_modifiers/`.
-    This will tell Forge about your modifiers and works similar to [tags][].
+1. Create a `global_loot_modifiers.json`.
+    This will tell Forge about your modifiers and works similar to [tags].
```
sciwhiz12: Please add a space between the paragraph and the list, so it renders correctly.

docs/items/globallootmodifiers.md
```diff
@@ -7,133 +7,142 @@ Registering a Global Loot Modifier
...
     The operational code that makes your modifier work. Most modders can extend `LootModifier` as it supplies base functionality.
-4. Finally, the serializer for your operational class.
+4. Finally, a class that extends `GlobalLootModifierSerializer` for your operational class.
     This is [registered] as any other `IForgeRegistryEntry`.
 
 The `global_loot_modifiers.json`
 -------------------------------
 
-All you need to add here are the file names of your loot modifiers. The [ResourceLocation][resloc] specified points to `data/<namespace>/loot_modifiers/<path>.json`.
+The `global_loot_modifiers.json` represents all loot modifiers to be loaded into the game. This file must be placed within `data/forge/loot_modifiers/global_loot_modifiers.json`.
```
sciwhiz12: Would it be worth reinforcing the idea that the file _must_ be under the `forge` namespace rather than under the mod's namespace?

docs/items/globallootmodifiers.md
```diff
@@ -7,133 +7,142 @@ Registering a Global Loot Modifier
...
 
-```json
+`entries` is an *ordered list* of the modifiers that will be loaded. This means that any modifier that not listed will not be loaded, and the ones listed are their written order. The [ResourceLocation][resloc]s specified points to their associated entry within `data/<namespace>/loot_modifiers/<path>.json`. This is primarily relevant to data pack makers for resolving conflicts between modifiers from separate mods.
```
sciwhiz12: I think it is not needed to mention that those modifiers not on the list are not loaded.

docs/items/globallootmodifiers.md
```diff
@@ -7,133 +7,142 @@ Registering a Global Loot Modifier
...
+
+`replace` causes the cache of modifiers to be cleared fully when this asset loads (mods are loaded in an order that may be specified by a data pack). For modders, you will want to use `false`. Data pack makers may want to specify their overrides with `true`.
```
sciwhiz12: I do wonder if this could be reworded more simply as "this changes the behavior from appending loot modifiers to the cache/list to replacing the current list entirely".

docs/items/globallootmodifiers.md
```diff
@@ -7,133 +7,142 @@ Registering a Global Loot Modifier
...
-  "replacement": "minecraft:wheat"
+  "prop1": "val1",
+  "prop2": 10,
+  "prop3": "minecraft:dirt"
 }
 ```
 
-In the above example, the modification only happens if an entity harvests wheat when using shears (specified by the two `conditions` which are automatically `AND`ed together). The `seedsItem` and `numSeeds` values are then used to count how many seeds were generated by the vanilla loot table, and if matched, are substituted for an additional `replacement` item instead. The operation code will be shown below.
-`conditions` is the only object needed by the system specification, everything else is the mod maker's data.
+# TODO: Document
```
sciwhiz12: Leftover TODO comment.

docs/items/globallootmodifiers.md
```diff
@@ -7,133 +7,142 @@ Registering a Global Loot Modifier
...
 
-Also take note of the `read` method in the serializer. The conditions are already deserialized for you and if you have no other data, simply `return new MyModifier(conditions)`. However, the full `JsonObject` is available if needed. The `write` method, on the other hand, is used for if you want to utilize `GlobalLootModifierProvider` for [data generation][datagen].
+`#write` is responsible for turning the defined loot modifier and writing it to a `JsonObject`. This requires that all conditions along with any custom properties must be written. For ease of convenience `#makeConditions` can be called to create a new `JsonObject` with the conditions already serialized within. Then, any additional properties can be added to this `JsonObject`. This is utilized for [data generation][datagen] of the associated loot modifier.
```
sciwhiz12: Pedantry comment of a missing comma and a bit of rewording:
```suggestion
`#write` is responsible for turning the defined loot modifier and writing it to a `JsonObject`. This requires that all conditions along with any custom properties must be written. For ease of convenience, `#makeConditions` can be called to create a new `JsonObject` with the conditions already serialized within. Any additional properties to be serialized can then be added to this `JsonObject`. This is utilized for [data generation][datagen] of the associated loot modifier.
```

sciwhiz12 (changes requested):

docs/items/globallootmodifiers.md
```diff
@@ -8,22 +8,33 @@ Registering a Global Loot Modifier
 
 You will need 4 things:
 1. Create a `global_loot_modifiers.json`.
```
sciwhiz12:
```suggestion
You will need 4 things:

1. Create a `global_loot_modifiers.json`.
```

sciwhiz12 (changes requested):

sciwhiz12 (dismissed):

docs/items/globallootmodifiers.md
```diff
@@ -7,133 +7,152 @@ Registering a Global Loot Modifier
...
+`#write` is responsible for turning the defined loot modifier and writing it to a `JsonObject`. This requires that all conditions along with any custom properties must be written. For ease of convenience, `#makeConditions` can be called to create a new `JsonObject` with the conditions already serialized within. Any additional properties to be serialized can then be added to this `JsonObject`. This is utilized for [data generation][datagen] of the associated loot modifier.
+
+```java
+public ExampleModifierSerializer extends GlobalLootModifierSerializer<ExampleModifier> {
+
+    @Override
+    public ExampleModifier read(ResourceLocation location, JsonObject object, LootItemCondition[] ailootcondition) {
+        String prop1 = GsonHelper.getAsString(object, "prop1");
+        // Deserializer other properties
+        return new ExampleModifier(ailootcondition, prop1, prop2, prop3);
```
sciwhiz12: Just noticed this while skim-reviewing:
```suggestion
    public ExampleModifier read(ResourceLocation location, JsonObject object, LootItemCondition[] conditions) {
        String prop1 = GsonHelper.getAsString(object, "prop1");
        // Deserializer other properties
        return new ExampleModifier(conditions, prop1, prop2, prop3);
```

sciwhiz12 (approved):

TheCurle (approved):

PR #413, opened 2022-02-23
[1.16.x] Upstream #409: Add info for registering to vanilla registries

ChampionAsh5357:
Port of #409 to 1.16.

> There are quite a few registries that are not forge registries so it's useful to have something to point people to.

### DIFF
--- a/docs/concepts/registries.md
+++ b/docs/concepts/registries.md
@@ -43,6 +43,16 @@ public void registerBlocks(RegistryEvent.Register<Block> event) {
 }
 ```
 
+### Registries that aren't Forge Registries
+
+Due to some peculiarities of vanilla code, not all registries are wrapped by Forge. These can be static registries, like `RecipeType`, which are safe to use. There are also dynamic registries, like `ConfiguredFeature` and some other worldgen registries, which are typically represented in JSON. These objects should only be registered this way if there is another registry object that requires it. 
+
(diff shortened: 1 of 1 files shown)


docs/concepts/registries.md
```diff
@@ -43,6 +43,16 @@ public void registerBlocks(RegistryEvent.Register<Block> event) {
 }
 ```
 
+### Registries that aren't Forge Registries
+
+Due to some peculiarities of vanilla code, not all registries are wrapped by Forge. These can be static registries, like `RecipeType`, which are safe to use. There are also dynamic registries, like `ConfiguredFeature` and some other worldgen registries, which are typically represented in JSON. These objects should only be registered this way if there is another registry object that requires it.
```
sciwhiz12: Change use of Mojmap class name to MCP class name:
```suggestion
Due to some peculiarities of vanilla code, not all registries are wrapped by Forge. These can be static registries, like `IRecipeType`, which are safe to use. There are also dynamic registries, like `ConfiguredFeature` and some other worldgen registries, which are typically represented in JSON. These objects should only be registered this way if there is another registry object that requires it. 
```

sciwhiz12 (changes requested):

sciwhiz12 (approved):

TheCurle (approved):

PR #416, opened 2022-03-02
[1.18.x] Add Admonition Convention

ChampionAsh5357:
Admonitions, as added by the [admonition markdown extension](https://github.com/Python-Markdown/markdown/blob/master/markdown/extensions/admonition.py), tend to be formatted incorrectly causing bad rendering within the docs.

This PR aims to address that by fixing all current bugs and adding an additional convention to the style guide linking the documentation needed to add a proper admonition.

### DIFF
--- a/docs/advanced/accesstransformers.md
+++ b/docs/advanced/accesstransformers.md
@@ -46,7 +46,7 @@ A special modifier `+f` and `-f` can be appended to the aforementioned modifiers
 Targets and Directives
 ----------------------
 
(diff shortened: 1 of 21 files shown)


sciwhiz12 (dismissed):

TheCurle (dismissed):

TheCurle:
This one's been on my TODO list for a while.

TheCurle (approved):

PR #418, opened 2022-03-03
1.18.2 Update

ChampionAsh5357:
Updates the necessary pages to 1.18.2 (really only tags in this case).

### DIFF
--- a/docs/utilities/tags.md
+++ b/docs/utilities/tags.md
@@ -5,7 +5,7 @@ Tags are generalized sets of objects in the game used for grouping related thing
 
 Declaring Your Own Groupings
 ----------------------------
-Tags are declared in your mod's [datapack][datapack]. For example, `/data/<modid>/tags/blocks/foo/tagname.json` will declare a `Tag$Named<Block>` with ID `modid:foo/tagname`.
+Tags are declared in your mod's [datapack][datapack]. For example, `/data/<modid>/tags/blocks/foo/tagname.json` will declare a `TagKey<Block>` with ID `modid:foo/tagname`.
 Similarly, you may append to or override tags declared in other domains, such as Vanilla, by declaring your own JSONs.
 For example, to add your own mod's saplings to the Vanilla sapling tag, you would specify it in `/data/minecraft/tags/blocks/saplings.json`, and Vanilla will merge everything into one tag at reload, if the `replace` option is false.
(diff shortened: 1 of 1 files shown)


docs/utilities/tags.md
```diff
@@ -5,7 +5,7 @@ Tags are generalized sets of objects in the game used for grouping related thing
 
 Declaring Your Own Groupings
 ----------------------------
-Tags are declared in your mod's [datapack][datapack]. For example, `/data/<modid>/tags/blocks/foo/tagname.json` will declare a `Tag$Named<Block>` with ID `modid:foo/tagname`.
+Tags are declared in your mod's [datapack][datapack]. For example, `/data/<modid>/tags/blocks/foo/tagname.json` will declare a `TagKey<Block>` with ID `modid:foo/tagname`.
```
sciwhiz12: Perhaps it is more accurate to reverse this definition: that a `TagKey` of a given location leads to a certain file location in the datapack.

Also, we may want to add a section describing the names of the tag folders, especially mentioning the special case of the IIRC five previous registries with tags in previous versions retaining their tags folder in the plural rather than following the registry's name.

docs/utilities/tags.md
```diff
@@ -33,24 +33,24 @@ You may declare a `remove` array of the same format as the `values` array. Any v
 
 Using Tags In Code
 ------------------
-Block, Item, Fluid, and EntityType tags are automatically sent from the server to any remote clients on login and reload. In addition, Forge adds BlockEntityType, Potion, and Enchantment tags. Function tags are not synced.
+Tags for all registries are automatically sent from the server to any remote clients on login and reload. However, only tags for `Block`s, `Item`s, `EntityType`s, `Fluid`s, and `GameEvent`s can be held by their registry objects.
```
sciwhiz12: I believe this is slightly inaccurate; it is not that only these five objects hold their tags, but rather that the tags in which they are contained in are accessible through the `Holder` available through `#builtInRegistryHolder()`. However, we may not want to mention this method, given that these are marked deprecated, presumably because Mojang plans to eliminate them in a future version.

sciwhiz12 (changes requested):

sciwhiz12 (approved):

TheCurle (approved):

PR #419, opened 2022-03-06
Fix tag creation to use ResourceLocation

MelanX:
Creating new tags requires `ResourceLocation`s instead of strings.


sciwhiz12 (approved):

sciwhiz12:
Superseded by #421 (author credit will be preserved).

PR #420, opened 2022-03-18
[1.18.x] Hitchhiker's Guide to All Things Recipes

### LINKED ISSUE #379: Adding custom Recipe Types
I would like to see a page about how to create custom Recipe Types.
I already did some research and I'm really confused aubout those serializers.

ChampionAsh5357:
This PR rewrites all the recipe documentation within Forge and expands upon all aspects relating to the recipe subsystem within Minecraft.

Closes Issues:
Closes #379

Major Changes:
- Moved docs/utilities/recipes.md -> doc/utilities/recipe/index.md
- Removed Vanilla pack-specific information and redirected to the Minecraft wiki
- Added discussion on RecipeManager, conditional recipes, larger crafting grids, ingredients, custom recipes, brewing recipes, anvil recipes, and data generation

Not Yet Implemented due to Issues/PRs:
- [x] Loom Recipes MinecraftForge/MinecraftForge#8518
  - Will fix in a later PR once this is merged.
- [x] New Ingredient Types MinecraftForge/MinecraftForge#8517
- [x] Recipe Book MinecraftForge/MinecraftForge#8028
  - Will defer to whenever the menu documentation is written up
- [x] Data Fixers MinecraftForge/MinecraftForge#8501
  - Will defer to whenever the data fixer documentation is merged and written up

Will not be Covered due to Lack of Implementation:
- Cartography Recipes


docs/utilities/recipes/ingredients.md
```diff
@@ -0,0 +1,173 @@
...
+  "item": "examplemod:example_item",
+  "nbt": {
+    // Add nbt data (must match exactly what is on the stack)
+  }
+}
+```
+
+### PartialNBTIngredient
+
+`PartialNBTIngredient`s are a looser version of [`NBTIngredient`][nbt] as they compare against a single or set of items and only keys specified within the tag. This can be used by specifying the `type` as `forge:partial_nbt`.
```
KnightMiner: Worth mentioning this also compares the share tag, ensures its consistent on client and server

docs/utilities/recipes/ingredients.md
```diff
@@ -0,0 +1,173 @@
...
+
+### Ingredient Subclass
+
+There are three important methods to implement for each ingredient subclass:
+
+Method | Description
+:---: | :---
+getSerializer | Returns the [serializer] used to read and write the ingredient.
+test | Returns true if the input is valid for this ingredient.
+isSimple | Returns false if the data needed from the stack is mutable (such as tag or capability data). This should not be overridden if returning true.
```
KnightMiner: I don't think that is accurate. My understanding is `isSimple` returns `false` if the ingredient matchs on based on NBT, and `true` if you can match entirely based on a set of IDs. As such, an ingredient matching on tags is considered simple, but an NBT ingredient is not.

Also worth nothing `AbstractIngredient` currently abstracts `isSimple` as I assumed all custom ingredients need to consider whether to return `true` or `false`, might be worth making that not abstract if you think there is a case for keeping the default value (defaulting an abstract method shouldn't be a breaking change).

ChampionAsh5357: > I don't think that is accurate. My understanding is isSimple returns false if the ingredient matchs on based on NBT, and true if you can match entirely based on a set of IDs. As such, an ingredient matching on tags is considered simple, but an NBT ingredient is not.

Yeah, I considered tag and items to be immutable in checking, but that would make it more clear.

> Might be worth making that not abstract if you think there is a case for keeping the default value (defaulting an abstract method shouldn't be a breaking change).

That's fine. This was written before 40.0.30, so this made the assumption that everyone would be subclassing `Ingredient` instead. A modder should explicitly state this imo.

docs/resources/server/recipes/index.md
```diff
@@ -0,0 +1,99 @@
...
+=======
+
+Recipes are a way to transform some number of objects into other objects within a Minecraft world. Although the vanilla system deals purely with item transformations, the system as a whole can be expanded to use any object the programmer creates.
+
+Data-Driven Recipes
+-------------------
+
+Most recipe implementations within vanilla are data driven via JSON. This means that a mod is not necessary to create a new recipe, only a [Data pack][datapack]. A full list on how to create and put these recipes within the mod's `resources` folder can be found on the [Minecraft Wiki][wiki].
+
+A recipe can be obtained within the Recipe Book as a reward for completing an [advancement][advancement]. Recipe advancements always parent `minecraft:recipes/root` so as not to appear on the Advancement screen and that they are treated as standalone. In addition, the criteria to unlock a recipe always checks if the user has unlocked the recipe from using it once:
```
sciwhiz12: The wording here is a bit awkward to me. Also, what does it mean for an advancement to be "treated as standalone"? (I omitted that part from my suggestion -- feel free to re-add it back with a sufficient explanation.)
```suggestion
A recipe can be obtained within the Recipe Book as a reward for completing an [advancement][advancement]. Recipe advancements always have `minecraft:recipes/root` as their parent, to not to appear on the advancement screen. The default criteria to gain the recipe advancement is a check if the user has unlocked the recipe from using it once or receiving it through a command like `/recipe`:
```

docs/resources/server/recipes/index.md
```diff
@@ -0,0 +1,99 @@
...
+
+Recipe Manager
+--------------
+
+Recipes are loaded and stored via the `RecipeManager`. Any operations relating to getting available recipe(s) are handled by this manager. There are two important methods to know of:
+
+Method|Description
+:---: | :---
+getRecipeFor | Gets the first recipe that matches the current input.
+getRecipesFor | Gets all recipes that match the current input.
```
sciwhiz12: Misalignment on the column separators. Also, we may want to have the method names in monospace.
```suggestion
 Method         | Description
 :---:          | :---
`getRecipeFor`  | Gets the first recipe that matches the current input.
`getRecipesFor` | Gets all recipes that match the current input.
```

docs/resources/server/recipes/index.md
```diff
@@ -0,0 +1,99 @@
...
+
+Each method takes in a `RecipeType`, which denotes what method is being applied to use the recipe (crafting, smelting, etc.), a `Container` which holds the configuration of the inputs, and the current level which is passed to `Recipe#matches` along with the container.
+
+!!! important
+    When using an [item handler capability][cap] as the configuration inputs, it must be wrapped via `RecipeWrapper` and passed in as `IItemHandler` and its subtypes are not instances of `Container`:
```
sciwhiz12: I think this can be worded more helpfully by reducing the imperative command of "must be wrapped" into mentioning the use of `RecipeWrapper` for wrapping `IItemHandler`s:
```suggestion
    Forge provides the `RecipeWrapper` utility class which extends `Container` for wrapping around `IItemHandler`s and passing them to methods which requires a `Container` parameter.
```

docs/resources/server/recipes/index.md
```diff
@@ -0,0 +1,99 @@
...
+    ```
+
+Additional Features
+-------------------
+
+Forge provides some additional behavior to the recipe schema and its implementations for greater control of the system.
+
+### Recipe ItemStack Result
+
+Except for `minecraft:stonecutting` recipes, all recipe serializers expand the `result` tag to take in a full `ItemStack` as a `JsonObject` instead of just the item name and amount in some cases.
```
sciwhiz12: All recipe serializers, or all _vanilla_ recipe serializers?

docs/resources/server/recipes/index.md
```diff
@@ -0,0 +1,99 @@
...
+// In some recipe JSON
+"result": {
+  // The name of the registry item to give as a result
+  "item": "examplemod:example_item",
+  // The number of items to return
+  "count": 4,
+  // The tag data of the stack, can also be a string
+  "nbt": {
+      // Add tag data here
+  }
```
sciwhiz12, line 72: We should mention somehow that the `nbt` tag can either be a JSON object, or a string containing SNBT (stringified NBT) for NBT which cannot be represented as a JSON object (such as those involving `IntArrayTag`s).

docs/resources/server/conditional.md
```diff
@@ -0,0 +1,177 @@
...
+### Tag Empty
+
+`TagEmptyCondition` returns true whenever the given item tag has no items within it. This is represented by `forge:tag_empty`.
+
+```json5
+// For some condition
+{
+  "type": "forge:tag_empty",
+   // Returns true if 'examplemod:example_tag' is an item tag with no entries
+  "item": "examplemod:example_tag"
```
sciwhiz12: The tag empty condition checks for `tag`, not `item`.
```suggestion
  "tag": "examplemod:example_tag"
```

docs/resources/server/conditional.md
```diff
@@ -0,0 +1,177 @@
...
+Custom conditions can be created by implementing `ICondition` and its associated `IConditionSerializer`.
+
+### ICondition
+
+Any condition only need to implement two methods:
+
+Method | Description
+:---: | :---
+getID | The registry name of the condition. Must be equivalent to [`IConditionSerializer#getID`][serializer]. Used only for [data generation][datagen].
+test | Returns true if the condition has been satisfied.
```
sciwhiz12: Misaligned column separators:
```suggestion
Method | Description
:---:  | :---
getID  | The registry name of the condition. Must be equivalent to [`IConditionSerializer#getID`][serializer]. Used only for [data generation][datagen].
test   | Returns true if the condition has been satisfied.
```

docs/resources/server/conditional.md
```diff
@@ -0,0 +1,177 @@
...
+
+### IConditionSerializer
+
+Serializers need to implement three methods:
+
+Method | Description
+:---: | :---
+getID | The registry name of the condition. Must be equivalent to [`ICondition#getID`][condition].
+read | Reads the condition data from JSON.
+write | Writes the given condition data to JSON.
```
sciwhiz12: Misaligned column separators:
```suggestion
Method | Description
:---:  | :---
getID  | The registry name of the condition. Must be equivalent to [`ICondition#getID`][condition].
read   | Reads the condition data from JSON.
write  | Writes the given condition data to JSON.
```

docs/resources/server/recipes/ingredients.md
```diff
@@ -0,0 +1,176 @@
...
+
+Custom ingredients can be specified by setting `type` to the name of the [ingredient's serializer][serializer] with only [one exception][compound]. When no type is specified, `type` defaults to the vanilla ingredient `minecraft:item`. Custom ingredients can also easily be used in [data generation][datagen].
+
+### Forge Types
+
+Forge provides a few additional `Ingredient` types for programmers to implement. 
+
+#### CompoundIngredient
+
+Though they are functionally identical, Compound ingredients replaces the way one would implement a list of ingredients would in a recipe. They work as a set OR where the passed in stack must be within at least one of the supplied ingredients. This change was made to allow custom ingredients to work correctly within lists. As such, **NO TYPE** needs to be specified.
```
sciwhiz12: No need to use all-caps when using bold text:
```suggestion
Though they are functionally identical, Compound ingredients replaces the way one would implement a list of ingredients would in a recipe. They work as a set OR where the passed in stack must be within at least one of the supplied ingredients. This change was made to allow custom ingredients to work correctly within lists. As such, **no type** needs to be specified.
```

docs/resources/server/recipes/ingredients.md
```diff
@@ -0,0 +1,176 @@
+Ingredients
+===========
+
+`Ingredient`s are predicate handlers for item-based inputs which check whether a certain `ItemStack` meets the condition to be a valid input in a recipe. All [vanilla recipes][recipes] that take inputs use an `Ingredient` or a list of `Ingredient`s, which is then merged into a single `Ingredient`.
+
+Custom Ingredients
+------------------
+
+Custom ingredients can be specified by setting `type` to the name of the [ingredient's serializer][serializer] with only [one exception][compound]. When no type is specified, `type` defaults to the vanilla ingredient `minecraft:item`. Custom ingredients can also easily be used in [data generation][datagen].
```
sciwhiz12: The part about "only one exception" should include what the exception is after it, to avoid needing to hover over or click the link to figure out what that exception is.
```suggestion
Custom ingredients can be specified by setting `type` to the name of the [ingredient's serializer][serializer], with the exception of [compound ingredients][compound]. When no type is specified, `type` defaults to the vanilla ingredient `minecraft:item`. Custom ingredients can also easily be used in [data generation][datagen].
```

docs/resources/server/recipes/ingredients.md
```diff
@@ -0,0 +1,176 @@
...
+Forge provides a few additional `Ingredient` types for programmers to implement. 
+
+#### CompoundIngredient
+
+Though they are functionally identical, Compound ingredients replaces the way one would implement a list of ingredients would in a recipe. They work as a set OR where the passed in stack must be within at least one of the supplied ingredients. This change was made to allow custom ingredients to work correctly within lists. As such, **NO TYPE** needs to be specified.
+
+```json5
+// For some input
+[
+  // At least one of these ingredients must return true to succeed
```
sciwhiz12: Probably better to use "must match" instead of the developer-oriented phrase "must return true":
```suggestion
  // At least one of these ingredients must match to succeed
```

docs/resources/server/recipes/ingredients.md
```diff
@@ -0,0 +1,176 @@
...
+  {
+    // Custom ingredient
+    "type": "examplemod:example_ingredient"
+  }
+]
+```
+
+#### NBTIngredient
+
+`NBTIngredient`s compare the item, damage, and the share tags (as defined by `IForgeItem#getShareTag`) on an `ItemStack` for equivalency. This can be used by specifying the `type` as `forge:nbt`.
```
sciwhiz12: The paragraph should note that `NBTIngredient` compares the share NBT tags _exactly_, to contrast with `PartialNBTIngredient` (and to avoid keeping important details in the example).

docs/resources/server/recipes/ingredients.md
```diff
@@ -0,0 +1,176 @@
...
+      // Ingredient 2
+    }
+    // ...
+  ]
+}
+```
+
+### DifferenceIngredient
+
+`DifferenceIngredient`s work as a set subtraction (SUB) where the passed in stack must be within the first ingredient but not in the second ingredient. This can be used by specifying the `type` as `forge:difference`.
```
sciwhiz12: Perhaps we can use the terminology "matches" rather than "within".
```suggestion
`DifferenceIngredient`s work as a set subtraction (SUB) where the passed in stack must match the first ingredient but must not match the second ingredient. This can be used by specifying the `type` as `forge:difference`.
```

docs/resources/server/recipes/ingredients.md
```diff
@@ -0,0 +1,176 @@
...
+    "key2": {
+      // Data 2
+    }
+  }
+}
+```
+
+### IntersectionIngredient
+
+`IntersectionIngredient`s work as a set AND where the passed in stack must be within all supplied ingredients. There must be at least two ingredients supplied to this. This can be used by specifying the `type` as `forge:intersection`.
```
sciwhiz12: Perhaps we can use the terminology "matches" rather than "within".
```suggestion
`IntersectionIngredient`s work as a set AND where the passed in stack must match all supplied ingredients. There must be at least two ingredients supplied to this. This can be used by specifying the `type` as `forge:intersection`.
```

docs/resources/server/recipes/ingredients.md
```diff
@@ -0,0 +1,176 @@
...
+
+### IIngredientSerializer
+
+`IIngredientSerializer` subtypes must implement three methods:
+
+Method | Description
+:---: | :---
+parse (JSON) | Converts a `JsonObject` to an `Ingredient`.
+parse (Network) | Reads the network buffer to decode an `Ingredient`.
+write | Writes an `Ingredient` to the network buffer.
```
sciwhiz12: Table column separators should be aligned.
```suggestion
 Method         | Description
 :---:          | :---
parse (JSON)    | Converts a `JsonObject` to an `Ingredient`.
parse (Network) | Reads the network buffer to decode an `Ingredient`.
write           | Writes an `Ingredient` to the network buffer.
```

docs/resources/server/recipes/ingredients.md
```diff
@@ -0,0 +1,176 @@
...
+
+### Ingredient Subclass
+
+There are three important methods to implement for each ingredient subclass:
+
+Method | Description
+:---: | :---
+getSerializer | Returns the [serializer] used to read and write the ingredient.
+test | Returns true if the input is valid for this ingredient.
+isSimple | Returns false if the ingredient matches on the stack's tag. `AbstractIngredient` subclasses will need to define this behavior, while `Ingredient` subclasses return `true` by default.
```
sciwhiz12: Table column separators should be aligned.
```suggestion
 Method       | Description
 :---:        | :---
getSerializer | Returns the [serializer] used to read and write the ingredient.
test          | Returns true if the input is valid for this ingredient.
isSimple      | Returns false if the ingredient matches on the stack's tag. `AbstractIngredient` subclasses will need to define this behavior, while `Ingredient` subclasses return `true` by default.
```

docs/resources/server/recipes/ingredients.md
```diff
@@ -0,0 +1,176 @@
...
+Afterwards, a static instance should be declared to hold the initialized serializer and then registered using `CraftingHelper#register` either during the `RegistryEvent$Register` for `RecipeSerializer`s or during `FMLCommonSetupEvent`. The `Ingredient` subclass return the static instance of the serializer in `Ingredient#getSerializer`.
+
+```java
+// In some serializer class
+public static final ExampleIngredientSerializer INSTANCE = new ExampleIngredientSerializer();
+
+// In some handler class
+public void registerSerializers(RegistryEvent.Register<RecipeSerializer<?>> event) {
+  CraftingHelper.register(registryName, INSTANCE);
+}
```
sciwhiz12, line 157: Not a change request, but we should keep note of this as the registry events will change come 1.19 with the registry rework PR at MinecraftForge/MinecraftForge#8538 -- specifically, the replacement of `RegistryEvent.Register` with a new `RegisterEvent`.

docs/resources/server/recipes/ingredients.md
```diff
@@ -0,0 +1,176 @@
...
+
+// In some ingredient subclass
+@Override
+public IIngredientSerializer<? extends Ingredient> getSerializer() {
+  return INSTANCE;
+}
+```
+
+!!! important
+    If using `FMLCommonSetupEvent` to register an ingredient serializer, it must be enqueued to the synchronous work queue via `FMLCommonSetupEvent#enqueueWork` as `CraftingHelper#register` is not thread-safe.
```
sciwhiz12: Perhaps this can be a tip instead of an important admonition, given that we first recommend the use of the registry event to mark when to register ingredient serializers.

docs/resources/server/recipes/ingredients.md
```diff
@@ -0,0 +1,176 @@
...
+Method | Description
+:---: | :---
+parse (JSON) | Converts a `JsonObject` to an `Ingredient`.
+parse (Network) | Reads the network buffer to decode an `Ingredient`.
+write | Writes an `Ingredient` to the network buffer.
+
+Additionally, `Ingredient` subclasses should implement `Ingredient#toJson` for use with [data generation][datagen].
+
+!!! important
+    `Ingredient#toJson` is made abstract if using an `AbstractIngredient` subclass and must be implemented.
```
sciwhiz12: I feel like either both of these paragraphs should be in the admonition, or the use of the admonition removed and its contents be placed as regular text (perhaps combined into a single paragraph).

mkdocs.yml
```diff
@@ -47,14 +47,20 @@ nav:
...
       - Introduction: 'resources/server/index.md'
-      - Recipes: 'resources/server/recipes.md'
+      - Recipes:
+        - Introduction: 'resources/server/recipes/index.md'
+        - Custom Recipes: 'resources/server/recipes/custom.md'
+        - Ingredients: 'resources/server/recipes/ingredients.md'
+        - Non-Datapack Recipes:  'resources/server/recipes/incode.md'
       - Global Loot Modifiers: 'resources/server/glm.md'
       - Tags: 'resources/server/tags.md'
+      - Conditionally-Loaded Data: 'resources/server/conditional.md'
```
sciwhiz12, line 57: Is the use of the hyphen appropriate? I'm unsure myself.

docs/resources/server/recipes/incode.md
```diff
@@ -0,0 +1,69 @@
...
+
+Brewing is one of the few recipes that still exist in code. Brewing recipes are added as part of a bootstrap within `PotionBrewing` for their containers, container recipes, and potion mixes. To expand upon the existing system, Forge allows brewing recipes to be added by calling `BrewingRecipeRegistry#addRecipe` in `FMLCommonSetupEvent`.
+
+!!! warning
+    `BrewingRecipeRegistry#addRecipe` must be called within the synchronous work queue via `#enqueueWork` as the method is not thread-safe.
+
+The default implementation takes in an input ingredient, a catalyst ingredient, and a stack output for a standard implementation. Additionally, an `IBrewingRecipe` instance can be supplied instead to do the transformations.
+
+!!! note
+    There is no wrapper for adding additional potion containers or potion mixes similar to vanilla. A new `IBrewingRecipe` implementation will need to be added to replicate this behavior.
```
sciwhiz12: Perhaps this admonition could be moved to be under the section for `IBrewingRecipe`, and be lowered from an admonition?

docs/resources/server/recipes/incode.md
```diff
@@ -0,0 +1,69 @@
...
+
+The update event must be [attached] to the Forge event bus.
+
+Loom Recipes
+------------
+
+Looms are responsible for applying a dye and pattern (either from the loom or from an item) to a banner. While the banner and the dye must be a `BannerItem` or `DyeItem` respectively, custom patterns can be created and applied in the loom. Banner Patterns can be created by calling `BannerPattern#create` during mod construction.
+
+!!! warning
+    `BannerPattern`s which return `true` for `#hasPatternItem` do not appear as an option in the loom. These patterns must have an accompanying `BannerPatternItem` to be used.
```
sciwhiz12: This doesn't seem like a warning, but more of an important admonition.

docs/datagen/server/recipes.md
```diff
@@ -0,0 +1,185 @@
...
+
+After implementation, the provider must be [added][datagen] to the `DataGenerator`.
+
+`RecipeBuilder`
+---------------
+
+`RecipeBuilder` is a convenience implementation for creating `FinishedRecipe`s to generate. It provides basic definitions for unlocking, grouping, saving, and getting the result of a recipe. This is done through `#unlockedBy`, `#group`, `#save`, and `#getResult` respectively.
+
+!!! important
+    [`ItemStack` outputs][stack] in recipes is not supported within vanilla recipe builders for `RecipeProvider`. A `FinishedRecipe` must be built in a different manner for existing vanilla recipe serializers to generate this data.
```
sciwhiz12: Small grammatical issue:
```suggestion
    [`ItemStack` outputs][stack] in recipes are not supported within vanilla recipe builders for `RecipeProvider`. A `FinishedRecipe` must be built in a different manner for existing vanilla recipe serializers to generate this data.
```

docs/datagen/server/recipes.md
```diff
@@ -0,0 +1,185 @@
...
+
+!!! warning
+    The item results being generated must have a valid `CreativeModeTab` specified; otherwise, a `NullPointerException` will be thrown. If no category accurately represents the item and no tab should be created, set the `#tab` property during item initialization to `CreativeModeTab#TAB_SEARCH`.
```
sciwhiz12, line 17: Why `TAB_SEARCH`?

docs/datagen/server/recipes.md
```diff
@@ -0,0 +1,185 @@
...
+
+### ShapedRecipeBuilder
+
+`ShapedRecipeBuilder` is used to generate shaped recipes. The builder can be initialized via `#shaped`. The recipe group, input symbol pattern, symbol definition of ingredients, and how the recipe is unlocked can be specified before saving.
+
+The advancement criteria for having the recipe is added and ORed together with all other added criteria.
```
sciwhiz12: Are there no other more natural-sounding alternatives to `ORed`?

docs/datagen/server/recipes.md
```diff
@@ -0,0 +1,185 @@
...
+// Add data to builder
+builder.save(writer);
+```
+
+#### Validation Checks
+
+* A pattern must be defined and take in more than one item.
+* All pattern rows must be the same width.
+* A symbol cannot be defined more than once.
+* The whitespace (' ') symbol is reserved and, as such, cannot be defined.
```
sciwhiz12: The term "whitespace" is taken to represent any blank vertical or horizontal space, which includes tabs and spaces. In this case, it is better to specify this as the "space character" instead. Additionally, the space character in quotes should be in monospace and after the word symbol/character.

docs/datagen/server/recipes.md
```diff
@@ -0,0 +1,185 @@
...
+* A pattern must only use all defined symbols.
```
sciwhiz12: The word "all" makes the meaning of the sentence a bit unclear to me: does this mean that the pattern must only use those symbols defined previously, or that the pattern must make use of all symbols defined previously?

docs/datagen/server/recipes.md
```diff
@@ -0,0 +1,185 @@
...
+* A criteria besides from using the recipe must be specified to unlock the recipe.
```
sciwhiz12, line 46: There should be a note or tip about the conventional use of `InventoryChangeTrigger` as this criteria with the result item.

docs/datagen/server/recipes.md
```diff
@@ -0,0 +1,185 @@
...
+// In RecipeProvider#buildCraftingRecipes(writer)
+ShapedRecipeBuilder builder = ShapedRecipeBuilder.shaped(result);
+// Add data to builder
+builder.save(writer);
+```
+
+#### Validation Checks
+
+* A pattern must be defined and take in more than one item.
+* All pattern rows must be the same width.
```
sciwhiz12, line 42: It might or might not be worth specifying that the pattern row strings only need be the same -- that is, a pattern row may not be fully specified (`" # "`) or even blank, as long as the definition of the row as a string literal matches the length of all other rows as string literals.

docs/datagen/server/recipes.md
```diff
@@ -0,0 +1,185 @@
...
+The advancement criteria for having the recipe is added and ORed together with all other added criteria.
+
+```java
+// In RecipeProvider#buildCraftingRecipes(writer)
+ShapelessRecipeBuilder builder = ShapelessRecipeBuilder.shapeless(result);
+// Add data to builder
+builder.save(writer);
+```
+
+#### Validation Checks
```
sciwhiz12: Should these be individual sections, or a paragraph followed by a list in the parent section? Since most of these sections (with the notable exception of the first section) contain one bullet point, the use of a whole section sounds a bit too much to me.

docs/datagen/server/recipes.md
```diff
@@ -0,0 +1,185 @@
...
+
+!!! important
+    [`ItemStack` outputs][stack] in recipes is not supported within vanilla recipe builders for `RecipeProvider`. A `FinishedRecipe` must be built in a different manner for existing vanilla recipe serializers to generate this data.
+
+!!! warning
+    The item results being generated must have a valid `CreativeModeTab` specified; otherwise, a `NullPointerException` will be thrown. If no category accurately represents the item and no tab should be created, set the `#tab` property during item initialization to `CreativeModeTab#TAB_SEARCH`.
+
+### ShapedRecipeBuilder
+
+`ShapedRecipeBuilder` is used to generate shaped recipes. The builder can be initialized via `#shaped`. The recipe group, input symbol pattern, symbol definition of ingredients, and how the recipe is unlocked can be specified before saving.
```
sciwhiz12: Perhaps it may be good to note somewhere here that these factory methods are meant to be statically imported.

Also, adding parallelism for that third item in the last sentence would be good for readability.
```suggestion
`ShapedRecipeBuilder` is used to generate shaped recipes. The builder can be initialized via `#shaped`. The recipe group, input symbol pattern, symbol definition of ingredients, and the recipe unlock criteria can be specified before saving.
```

docs/datagen/server/recipes.md
```diff
@@ -0,0 +1,185 @@
...
+
+The advancement criteria for having the recipe is added and ORed together with all other added criteria.
+
+```java
+// In RecipeProvider#buildCraftingRecipes(writer)
+ShapedRecipeBuilder builder = ShapedRecipeBuilder.shaped(result);
+// Add data to builder
+builder.save(writer);
+```
```
sciwhiz12: Perhaps this should be formatted to show the fluent design of the recipe builder methods, which allow for chaining calls up to the final `save` call.

docs/datagen/server/recipes.md
```diff
@@ -0,0 +1,185 @@
...
+
+#### Validation Checks
+
+* A criteria besides from using the recipe must be specified to unlock the recipe.
+
+Non-`RecipeBuilder` Builders
+----------------------------
+
+Some recipe builders do not implement `RecipeBuilder` due to lacking  features used by all previously mentioned recipes.
```
sciwhiz12: Extra whitespace:
```suggestion
Some recipe builders do not implement `RecipeBuilder` due to lacking features used by all previously mentioned recipes.
```

docs/datagen/server/recipes.md
```diff
@@ -0,0 +1,185 @@
...
+To simplify adding conditions to conditional recipes without having to construct the instances of each condition instance manually, the extended `RecipeProvider` can implement `IConditionBuilder`. The interface adds methods to easily construct condition instances.
+
+```java
+// In ConditionalRecipe$Builder#addCondition
+(
+  // If either 'examplemod:example_item'
+  // OR 'examplemod:example_item2' exists
+  // AND
+  // NOT FALSE
+  and(
```
sciwhiz12: There should probably be a note that these are static imports of the methods in `IConditionBuilder`.

docs/resources/server/recipes/custom.md
```diff
@@ -0,0 +1,134 @@
...
+!!! tip
+    `#assemble` should always produce a unique `ItemStack`. If unsure whether `#assemble` does so, call `ItemStack#copy` on the result before returning.
+
+Most of the other methods are purely for integration with the recipe book.
+
+```java
+public record ExampleRecipe(Ingredient input, int data, ItemStack output) implements Recipe<Container> {
+  // Implement methods here
+}
+```
```
sciwhiz12, line 27: Perhaps it is prudent to note that while one may use a record for a recipe, it is in no way required.

docs/resources/server/recipes/custom.md
```diff
@@ -0,0 +1,134 @@
...
+
+RecipeType
+----------
+
+`RecipeType` is responsible for defining the category or context the recipe will be used within. For example, if a recipe was going to be smelted in a furnace, it would have a type of `RecipeType#SMELTING`.  Being blasted in a blast furnace would have a type of `RecipeType#BLASTING`.
```
sciwhiz12: Extra whitespace:
```suggestion
`RecipeType` is responsible for defining the category or context the recipe will be used within. For example, if a recipe was going to be smelted in a furnace, it would have a type of `RecipeType#SMELTING`. Being blasted in a blast furnace would have a type of `RecipeType#BLASTING`.
```

docs/resources/server/recipes/custom.md
```diff
@@ -0,0 +1,134 @@
...
+
+A `RecipeSerializer` is responsible for decoding JSONs and communicating across the network for an associated `Recipe` subtype. Each recipe decoded by the serializer is saved as a unique instance within the `RecipeManager`.
+
+Only three methods need to be implemented for a `RecipeSerializer`:
+
+Method | Description
+:---: | :---
+fromJson | Decodes a JSON into the `Recipe` subtype.
+toNetwork | Encodes a `Recipe` to the buffer to send to the client. The recipe identifier does not need to be encoded.
+fromNetwork | Decodes a `Recipe` from the buffer sent from the server. The recipe identifier does not need to be decoded.
```
sciwhiz12: Table column seperators should be aligned.
```suggestion
 Method     | Description
 :---:      | :---
fromJson    | Decodes a JSON into the `Recipe` subtype.
toNetwork   | Encodes a `Recipe` to the buffer to send to the client. The recipe identifier does not need to be encoded.
fromNetwork | Decodes a `Recipe` from the buffer sent from the server. The recipe identifier does not need to be decoded.
```

docs/resources/server/recipes/custom.md
```diff
@@ -0,0 +1,134 @@
...
+
+A `RecipeSerializer` must be [registered][forge].
```
sciwhiz12: This standalone paragraph could be combined with the first paragraph.

docs/resources/server/recipes/custom.md
```diff
@@ -0,0 +1,134 @@
...
+    For ease of convenience, the `RecipeSerializer` subtype can extend `ForgeRegistryEntry` to implement the methods within `IForgeRegistryEntry`.
+
+```java
+public class ExampleSerializer extends ForgeRegistryEntry<RecipeSerializer<?>> implements RecipeSerializer<ExampleRecipe> {
+  // Implement methods here
+}
+```
+
+!!! tip
+    There are some useful methods to make reading and writing data for recipes easier. `Ingredient`s can use `#fromJson`, `#toNetwork`, and `#fromNetwork` while `ItemStack`s can use `CraftingHelper#getItemStack`, `FriendlyByteBuf#writeItem`, and `FriendlyByteBuf# readItem`.
```
sciwhiz12: For the sake of spacing out admonitions, please move this to the end of the section.

docs/resources/server/recipes/custom.md
```diff
@@ -0,0 +1,134 @@
...
+// Checks the block at the position to see if it matches the stored data
+boolean matches(Level level, BlockPos pos);
+
+// Creates the block state to set the block at the specified position to
+BlockState assemble();
+
+// In some manager class
+public Optional<ExampleRecipe> getRecipeFor(Level level, BlockPos pos) {
+  return level.getRecipeManager().getAllRecipesFor(exampleRecipeType)
+    .stream().filter(recipe -> recipe.matches(level, pos)).findFirst();
```
sciwhiz12: Perhaps for readability, each operation in this stream should be on its own separate line.

sciwhiz12 (changes requested):

docs/datagen/server/recipes.md
ChampionAsh5357: It is the only category that can view all items within the creative inventory. As such, if no item is applicable for any tab and the user doesn't want to create a custom tab, it would make sense that it would be found in at least that location, hence the setting.

docs/datagen/server/recipes.md
ChampionAsh5357: Probably, I just need to swap around the words.

mkdocs.yml
ChampionAsh5357: Grammatically, it can go either way. It's just a matter of preference.

docs/datagen/server/recipes.md
ChampionAsh5357: I feel like that's more confusing. The pattern is still the same width regardless and would be fully specified. And the point below about whitespace would clear up the definition to count it towards the width of the pattern.

docs/datagen/server/recipes.md
ChampionAsh5357: A pattern must use all symbols defined by the user. If the user does not use a symbol or does not define a symbol, then an error is thrown. Fixed the wording a bit to clarify.

docs/datagen/server/recipes.md
ChampionAsh5357: They are not static imports, they are defaulted methods as a result of implentation.

docs/resources/server/recipes/ingredients.md
ChampionAsh5357: Added to my recipe notes.

docs/datagen/server/recipes.md
sciwhiz12: Are there any possibly unwanted side-effects with using the search tab?

docs/datagen/server/recipes.md
```diff
@@ -0,0 +1,173 @@
...
+* A pattern must be defined and take in more than one item.
+* All pattern rows must be the same width.
+* A symbol cannot be defined more than once.
+* The space character (`' '`) is reserved for representing no item in a slot and, as such, cannot be defined.
+* A pattern must use all symbols defined by the user.
+* A criteria besides from using the recipe must be specified to unlock the recipe.
+
+### ShapelessRecipeBuilder
+
+`ShapelessRecipeBuilder` is used to generate shapeless recipes. The builder can be initialized via `#shapeless`. The recipe group, input ingredients, and how the recipe is unlocked can be specified before saving.
```
sciwhiz12: Parallelism as per https://github.com/MinecraftForge/Documentation/pull/420#discussion_r845640222 (please also repeat this for the following sections):
```suggestion
`ShapelessRecipeBuilder` is used to generate shapeless recipes. The builder can be initialized via `#shapeless`. The recipe group, input ingredients, and the recipe unlock criteria can be specified before saving.
```

sciwhiz12 (changes requested):

docs/datagen/server/recipes.md
ChampionAsh5357: No, creative tabs are populated using `Item#fillItemCategory` which performs a check called `#allowedIn` which determines whether or not the item can appear in the creative inventory. If the creative tab is not null, then it will populate with the item if the tab is equal to the set category or the search tab. As such, the only thing this would affect is that the OR check for creative tab would essentially be the same.

You can verify this is a workspace as well.

sciwhiz12: Hmmm, looking at the code which generates the recipes shows that the output location of the generated recipe JSON uses the creative mode tab of the resulting item for convenient grouping: `new ResourceLocation(recipeId.getNamespace(), "recipes/" + this.result.getItemCategory().getRecipeFolderName() + "/" + recipeId.getPath())`.

This means that a recipe whose result item's tab is set to the `SEARCH` creative mode tab will be placed under `recipes/search/`, which may be an unwanted side-effect.

ChampionAsh5357: Unwanted or not, you cannot generate a recipe using the builder without a creative tab, which I agree should be the case. If it gets generated into that folder as a result, it would make sense that it would belong there.

sciwhiz12: Is it good practice to add items to the search tab, such that the item can only appear in the creative menu through the search tab and no other tab? Are there cases in which a mod has an item which does not belong on normal creative tabs (whether vanilla or modded), but they wish to add a crafting recipe for?

ChampionAsh5357: This is specifically dealing with a case where a modder willingly decides to hide their item from the creative inventory but makes it obtainable in survival, so this is my opinion on how to handle this unrecommended edge case. As for the second point, I address that in the [PR that attempts to change that behavior](https://github.com/MinecraftForge/MinecraftForge/pull/8511#issuecomment-1066870648).

sciwhiz12 (approved):

PR #421, opened 2022-03-22
[1.18.x] Tags 40.0.18

ChampionAsh5357:
Updates tags to include additional changes made in 40.0.18.

Supersedes #419.

### DIFF
--- a/docs/resources/server/tags.md
+++ b/docs/resources/server/tags.md
@@ -35,6 +35,9 @@ Using Tags In Code
 ------------------
 Tags for all registries are automatically sent from the server to any remote clients on login and reload. `Block`s, `Item`s, `EntityType`s, `Fluid`s, and `GameEvent`s are special cased as they have `Holder`s allowing for available tags to be accessible through the object itself.
 
+!!! note
+    Intrusive `Holder`s may be removed in a future version of Minecraft. If they are, the below methods can be used instead to query the associated `Holder`s.
+
 ### ITagManager
 
 Forge wrapped registries provide an additional helper for creating and managing tags through `ITagManager` which can be obtained via `IForgeRegistry#tags`. Tags can be created using using `#createTagKey` or `#createOptionalTagKey`. Tags or registry objects can also be checked for either or using `#getTag` or `#getReverseTag` respectively.
@@ -69,14 +72,16 @@ public static final TagKey<Item> myItemTag = ItemTags.create(new ResourceLocatio
 public static final TagKey<Potion> myPotionTag = ForgeRegistries.POTIONS.tags().createTagKey(new ResourceLocation("mymod", "mypotiongroup"));
 
 public static final TagKey<VillagerType> myVillagerTypeTag = TagKey.create(Registry.VILLAGER_TYPE, new ResourceLocation("mymod", "myvillagertypegroup"));
-    
-// In some method where stack is an ItemStack
+
+// In some method:
+
+ItemStack stack = /*...*/;
 boolean isInItemGroup = stack.is(myItemTag);
 
-// In some method where potion is a Potion
+Potion potion = /*...*/;
 boolean isInPotionGroup  = ForgeRegistries.POTIONS.tags().getTag(myPotionTag).contains(potion);
 
-// In some method where villagerTypeKey is a ResourceKey<VillagerType>
+ResourceKey<VillagerType> villagerTypeKey = /*...*/;
 boolean isInVillagerTypeGroup = Registry.VILLAGER_TYPE.getHolder(villagerTypeKey).map(holder -> holder.is(myVillagerTypeTag)).orElse(false);
 ```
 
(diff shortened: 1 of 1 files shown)


docs/utilities/tags.md
```diff
@@ -35,25 +35,49 @@ Using Tags In Code
...
+### ITagManager
 
-Registry objects need to grab their associated holder using either `Registry#getHolder` or `Registry#getHolderOrThrow` and then compare if the registry object has a tag using `Holder#is`. Tag-holding registry objects contain a method called `#is` in either their registry object or state-aware class to check whether the object belongs to a certain tag.
+Forge wrapped registries provide an additional helper for creating and managing tags through `ITagManager` which can be obtained via `IForgeRegistry#tags`. Tags can be created using using `#createTagKey` or `#createOptionalTagKey`. Tags or registry objects can also be checked for either or using `#getTag` or `#getReverseTag` respectively.
+
+#### Custom Registries
+
+Custom registries can create tags when constructing their `DeferredRegister` via `#createTagKey` or `#createOptionalTagKey` respectively. Their tags or registry objects can then becked for either using the `IForgeRegistry` obtained by calling `DeferredRegister#makeRegistry`.
```
sciwhiz12: Looks like a typo of `checked` to me:
```suggestion
Custom registries can create tags when constructing their `DeferredRegister` via `#createTagKey` or `#createOptionalTagKey` respectively. Their tags or registry objects can then checked for either using the `IForgeRegistry` obtained by calling `DeferredRegister#makeRegistry`.
```

docs/utilities/tags.md
```diff
@@ -35,25 +35,49 @@ Using Tags In Code
...
+### Referencing Tags
+
+There are four methods of creating a tag wrapper:
+
+Method | For
+:---: | :---
+`*Tags#create` | `Block`, `Item`, `EntityType`, `Fluid`, and `Biome` where `*` represents one of these types.
+`ITagManager#createTagKey` | Forge wrapped vanilla registries, registries can be obtained from `ForgeRegistries`.
+`DeferredRegister#createTagKey` | Custom forge registries.
+`TagKey#create` | Vanilla registries without forge wrappers, registries can be obtained from `Registry`.
```
sciwhiz12: The style guide is silent on the matter of table formatting and spacing. Should we keep it as compact as possible here, or format it so it is somewhat readable in raw text?

docs/utilities/tags.md
```diff
@@ -35,25 +35,49 @@ Using Tags In Code
...
 // In some method where stack is an ItemStack
 boolean isInItemGroup = stack.is(myItemTag);
 
-// In some method where potionKey is a ResourceKey<Potion>
-boolean isInItemGroup = Registry.POTION.getHolder(potionKey).map(holder -> holder.is(myPotionTag)).orElse(false);
+// In some method where potion is a Potion
+boolean isInPotionGroup  = ForgeRegistries.POTIONS.tags().getTag(myPotionTag).contains(potion);
+
+// In some method where villagerTypeKey is a ResourceKey<VillagerType>
+boolean isInVillagerTypeGroup = Registry.VILLAGER_TYPE.getHolder(villagerTypeKey).map(holder -> holder.is(myVillagerTypeTag)).orElse(false);
```
sciwhiz12: Perhaps it may be better to represent variables of a particular type as, for example, `Potion potion = /* ... */;`, instead of a comment.

Also, could move the `In some method` part as a 'header' comment (`// In some method: //`) before each block instead of being repeated in each comment.

docs/utilities/tags.md
```diff
@@ -35,25 +35,49 @@ Using Tags In Code
 ------------------
 Tags for all registries are automatically sent from the server to any remote clients on login and reload. `Block`s, `Item`s, `EntityType`s, `Fluid`s, and `GameEvent`s are special cased as they have `Holder`s allowing for available tags to be accessible through the object itself.
```
sciwhiz12: To my knowledge, the instrusive `Holder.Reference` held by instances of the foregoing are likely to be removed in some future version of Minecraft. This may be worth noting so modders query the registry instead, therefore saving themselves the hassle when the removal is done and they port their mod forward.

sciwhiz12 (changes requested):

docs/utilities/tags.md
ChampionAsh5357: Personally, I usually compress as much as possible since preview tools on text editors usually will look fine. Throwing this question to @TheCurle.

TheCurle: There's a middle ground - we don't want the docs to be too hostile to edit, and we don't really have size constraints here. Don't compress for the sake of "small text tho", but don't have massive line breaks for no reason.

Personally, i find it easier for everyone involved if we try to keep it consistent and easy to read in raw form.

ChampionAsh5357: Ok, then while it won't be a convention, I'll still modify the table so it lines up within a text editor to better read in raw form.

sciwhiz12 (dismissed):

sciwhiz12 (approved):

PR #426, opened 2022-03-29
[1.18.x] Configurable Configurations

ChampionAsh5357:
Adds documentation for mod configurations. Anything that has been heavily documented by Forge or directly part of night-config has been delegated to their own library or documentation.

Supersedes and closes #303.

### DIFF
--- a/docs/misc/config.md
+++ b/docs/misc/config.md
@@ -1,0 +1,95 @@
+Configuration
+=============
+
+Configurations define settings and consumer preferences that can be applied to a mod instance. Forge uses a configuration system using [TOML][toml] files and read with [night-config][nightconfig].
+
+Creating a Configuration
+------------------------
+
+A configuration can be created using a subtype of `IConfigSpec`. Forge implements the type via `ForgeConfigSpec` and enables its construction through `ForgeConfigSpec#Builder`. The builder can separate the config values into sections via `Builder#push` to create a section and `Builder#pop` to leave a section. Afterwards, the configuration can be built using one of two methods:
+
+Method    | Description
+:---      | :---
+build     | Creates the `ForgeConfigSpec`.
+configure | Creates a pair of the class holding the config values and the `ForgeConfigSpec`.
+
+Each config value can be supplied with additional context to provide additional behavior. Contexts must be defined before the config value is fully built:
+
+Method       | Description
+:---         | :---
+comment      | Provides a description of what the config value does. Can provide multiple strings for a multiline comment.
+translation  | Provides a translation key for the name of the config value.
+worldRestart | The world must be restarted before the config value can be changed.
+
+### ConfigValue
+
+Config values can be built with the provided contexts (if defined) using any of the `#define` methods. The methods take in four components: a path representing the name of the variable, the default value when no valid configuration is present, a validator to make sure the deserialized object is valid, and a class representing the data type of the config value.
+
+!!! note
+    The path is a `.` separated string representing the sections the config value is in.
+
+```java
+// For some ForgeConfigSpec$Builder builder
+ConfigValue<T> value = builder.comment("Comment")
+  .define("config_value_name", defaultValue);
+```
+
+The values themselves can be obtained using `ConfigValue#get`. The values are additionally cached to prevent multiple readings from files.
+
+#### Range Values
+
+Config values that are specified within a range can be defined using `#defineInRange`. These expect the value to be a subtype of `Comparable`. The method is broken down into five components: a path representing the name of the variable, the default value when no valid configuration is present, the minimum and maximum value, and a class representing the data type of the config value.
+
+#### Whitelisted Values
+
+Config values that must be a part of a specific whitelist can be defined using `#defineInList`. The method is broken down into three components: a path representing the name of the variable, the default value when no valid configuration is present, and a collection of the allowed values the configuration can be.
+
+#### List Values
+
+Config values that can take in a list of entries can be defined using `#defineList`. If the list provided by the consumer can be empty, `#defineListAllowEmpty` should be used instead. The method is broken down into three components: a path representing the name of the variable, the default value when no valid configuration is present, and a validator to make sure a deserialized element from the list is valid.
+
+#### Enum Values
+
+Config values that represent an `Enum` can be defined by `EnumValue`s via `#defineEnum`. The method is broken down into four components: a path representing the name of the variable, the default value when no valid configuration is present, a getter to convert a string or integer into an enum, and a collection of the allowed values the configuration can be.
+
+#### Boolean Values
+
+Config values that represent a `boolean` can be defined by `BooleanValue`s via `#define`. The method is broken down into two components: a path representing the name of the variable and the default value when no valid configuration is present.
+
+#### Double Values
+
+Config values that represent a `double` can be defined by `DoubleValue`s via `#defineInRange`. The method is broken down into four components: a path representing the name of the variable, the default value when no valid configuration is present, and the minimum and maximum value.
+
+#### Int Values
+
+Config values that represent an `int` can be defined by `IntValue`s via `#defineInRange`. The method is broken down into four components: a path representing the name of the variable, the default value when no valid configuration is present, and the minimum and maximum value.
+
+#### Long Values
+
+Config values that represent a `long` can be defined by `LongValue`s via `#defineInRange`. The method is broken down into four components: a path representing the name of the variable, the default value when no valid configuration is present, and the minimum and maximum value.
+
+Registering a Configuration
+---------------------------
+
+Once a `ForgeConfigSpec` has been built, it must be registered to allow Forge to load, track, and sync the configuration settings as required. Configurations should be registered in the mod constructor via `ModLoadingContext#registerConfig`. A configuration can be registered with a given type representing the side the config belongs to (`CLIENT`, `COMMON`, `SERVER`), the `ForgeConfigSpec`, and optionally a specific file name for the configuration.
+
+```java
+// In the mod constructor with a ForgeConfigSpec CONFIG
+ModLoadingContext.get().registerConfig(Type.COMMON, CONFIG);
+```
+
+!!! tip
+    The definition of each config type and what they are used for can be found as part of the [constant's javadoc][type].
+
+Reading a Configuration
+-----------------------
+
+Operations that occur whenever a config is loaded or reloaded can be done using the `ModConfigEvent$Loading` and `ModConfigEvent$Reloading` events. The events must be [registered][events] to the mod event bus.
+
+!!! warning
+    These events are called for all configurations for the mod; the `ModConfig` object provided should be used to denote which configuration is being loaded or reloaded.
+
+[toml]: https://toml.io/
+[nightconfig]: https://github.com/TheElectronWill/night-config
+[type]: https://github.com/MinecraftForge/MinecraftForge/blob/1.18.x/fmlcore/src/main/java/net/minecraftforge/fml/config/ModConfig.java#L108-L136
+[events]: ../concepts/events.md#creating-an-event-handler
--- a/docs/misc/config.md
+++ b/docs/misc/config.md
@@ -8,25 +8,25 @@ Creating a Configuration
 
 A configuration can be created using a subtype of `IConfigSpec`. Forge implements the type via `ForgeConfigSpec` and enables its construction through `ForgeConfigSpec$Builder`. The builder can separate the config values into sections via `Builder#push` to create a section and `Builder#pop` to leave a section. Afterwards, the configuration can be built using one of two methods:
 
-Method    | Description
-:---      | :---
+ Method     | Description
+ :---       | :---
 `build`     | Creates the `ForgeConfigSpec`.
 `configure` | Creates a pair of the class holding the config values and the `ForgeConfigSpec`.
 
 !!! note
-    `ForgeConfigSpec$Builder#configure` is typically used with a `static` block and a class that takes in `ForgeCondifSpec$Builder` as part of its constructor to attach and hold the values:
+    `ForgeConfigSpec$Builder#configure` is typically used with a `static` block and a class that takes in `ForgeConfigSpec$Builder` as part of its constructor to attach and hold the values:
 
     ```java
     // In some config class
     ExampleConfig(ForgeConfigSpec.Builder builder) {
-        // Define values here in final fields
+      // Define values here in final fields
     }
 
     // Somewhere the constructor is accessible
     static {
-        Pair<ExampleConfig, ForgeConfigSpec> pair = new ForgeConfigSpec.Builder()
-            .configure(ExampleConfig::new);
-        // Store pair values in some constant field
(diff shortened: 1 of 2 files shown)


docs/misc/config.md
```diff
@@ -0,0 +1,95 @@
+Configuration
+=============
+
+Configurations define settings and consumer preferences that can be applied to a mod instance. Forge uses a configuration system using [TOML][toml] files and read with [night-config][nightconfig].
```
sciwhiz12: According to the [README](https://github.com/TheElectronWill/night-config#readme), the official name of the project is "NightConfig".
```suggestion
Configurations define settings and consumer preferences that can be applied to a mod instance. Forge uses a configuration system using [TOML][toml] files and read with [NightConfig][nightconfig].
```

docs/misc/config.md
```diff
@@ -0,0 +1,95 @@
...
+
+Creating a Configuration
+------------------------
+
+A configuration can be created using a subtype of `IConfigSpec`. Forge implements the type via `ForgeConfigSpec` and enables its construction through `ForgeConfigSpec#Builder`. The builder can separate the config values into sections via `Builder#push` to create a section and `Builder#pop` to leave a section. Afterwards, the configuration can be built using one of two methods:
```
sciwhiz12: Do we want to expose the `IConfigSpec` type in the docs? It may be worth simplifying down to just the existence of `ForgeConfigSpec`.

docs/misc/config.md
```diff
@@ -0,0 +1,95 @@
...
+build     | Creates the `ForgeConfigSpec`.
+configure | Creates a pair of the class holding the config values and the `ForgeConfigSpec`.
+
+Each config value can be supplied with additional context to provide additional behavior. Contexts must be defined before the config value is fully built:
+
+Method       | Description
+:---         | :---
+comment      | Provides a description of what the config value does. Can provide multiple strings for a multiline comment.
+translation  | Provides a translation key for the name of the config value.
+worldRestart | The world must be restarted before the config value can be changed.
```
sciwhiz12: The method names should probably be in monospace.

docs/misc/config.md
```diff
@@ -0,0 +1,95 @@
...
+
+### ConfigValue
+
+Config values can be built with the provided contexts (if defined) using any of the `#define` methods. The methods take in four components: a path representing the name of the variable, the default value when no valid configuration is present, a validator to make sure the deserialized object is valid, and a class representing the data type of the config value.
+
+!!! note
+    The path is a `.` separated string representing the sections the config value is in.
```
sciwhiz12: Perhaps it may be better to make these four components as an unordered list, and inlining the note about the path in the list.

docs/misc/config.md
```diff
@@ -0,0 +1,95 @@
...
+
+Creating a Configuration
+------------------------
+
+A configuration can be created using a subtype of `IConfigSpec`. Forge implements the type via `ForgeConfigSpec` and enables its construction through `ForgeConfigSpec#Builder`. The builder can separate the config values into sections via `Builder#push` to create a section and `Builder#pop` to leave a section. Afterwards, the configuration can be built using one of two methods:
+
+Method    | Description
+:---      | :---
+build     | Creates the `ForgeConfigSpec`.
+configure | Creates a pair of the class holding the config values and the `ForgeConfigSpec`.
```
sciwhiz12: An admonition note regarding the use of `configure` would be good to see.

docs/misc/config.md
```diff
@@ -0,0 +1,95 @@
...
+
+Config values that represent a `double` can be defined by `DoubleValue`s via `#defineInRange`. The method is broken down into four components: a path representing the name of the variable, the default value when no valid configuration is present, and the minimum and maximum value.
+
+#### Int Values
+
+Config values that represent an `int` can be defined by `IntValue`s via `#defineInRange`. The method is broken down into four components: a path representing the name of the variable, the default value when no valid configuration is present, and the minimum and maximum value.
+
+#### Long Values
+
+Config values that represent a `long` can be defined by `LongValue`s via `#defineInRange`. The method is broken down into four components: a path representing the name of the variable, the default value when no valid configuration is present, and the minimum and maximum value.
```
sciwhiz12: These look very repetitive, especially the first one and last four sections. The sections on `double`, `int`, and `long` values can be defined into a single "Numerical Values" section. The information regarding the path of the config value and the default value should be deduplicated. Perhaps it may be worth converting these into lists instead of sections?

docs/misc/config.md
```diff
@@ -0,0 +1,95 @@
...
+
+Once a `ForgeConfigSpec` has been built, it must be registered to allow Forge to load, track, and sync the configuration settings as required. Configurations should be registered in the mod constructor via `ModLoadingContext#registerConfig`. A configuration can be registered with a given type representing the side the config belongs to (`CLIENT`, `COMMON`, `SERVER`), the `ForgeConfigSpec`, and optionally a specific file name for the configuration.
+
+```java
+// In the mod constructor with a ForgeConfigSpec CONFIG
+ModLoadingContext.get().registerConfig(Type.COMMON, CONFIG);
+```
+
+!!! tip
+    The definition of each config type and what they are used for can be found as part of the [constant's javadoc][type].
```
sciwhiz12: As the details of each config type are important to the selection and behavior of the config, those types should be documented here rather than linking to the javadocs.

docs/misc/config.md
```diff
@@ -0,0 +1,95 @@
...
+
+Reading a Configuration
+-----------------------
```
sciwhiz12: "Reading a Configuration" is too generic and unrelated a name for a section whose contents are related to listening for changes to the configuration through events.

sciwhiz12 (changes requested):
There should be a part in the docs (whether as a paragraph/admonition in the lead section or as a new section) which explains when it is appropriate to expose configurable settings as mod configuration files vs data files (in a datapack).

docs/misc/config.md
ChampionAsh5357: Technically, Forge exposes `IConfigSpec` as a result of `ModLoadingContext#registerConfig`. That's why I breifly mention the configuration interface. Although, that's all the documentation mentions as it more specifically uses `ForgeConfigSpec`.

docs/misc/config.md
```diff
@@ -0,0 +1,131 @@
...
+
+A configuration can be created using a subtype of `IConfigSpec`. Forge implements the type via `ForgeConfigSpec` and enables its construction through `ForgeConfigSpec$Builder`. The builder can separate the config values into sections via `Builder#push` to create a section and `Builder#pop` to leave a section. Afterwards, the configuration can be built using one of two methods:
+
+Method    | Description
+:---      | :---
+`build`     | Creates the `ForgeConfigSpec`.
+`configure` | Creates a pair of the class holding the config values and the `ForgeConfigSpec`.
+
+!!! note
+    `ForgeConfigSpec$Builder#configure` is typically used with a `static` block and a class that takes in `ForgeCondifSpec$Builder` as part of its constructor to attach and hold the values:
```
sciwhiz12: Typo:
```suggestion
    `ForgeConfigSpec$Builder#configure` is typically used with a `static` block and a class that takes in `ForgeConfigSpec$Builder` as part of its constructor to attach and hold the values:
```

docs/misc/config.md
```diff
@@ -0,0 +1,131 @@
...
+
+Creating a Configuration
+------------------------
+
+A configuration can be created using a subtype of `IConfigSpec`. Forge implements the type via `ForgeConfigSpec` and enables its construction through `ForgeConfigSpec$Builder`. The builder can separate the config values into sections via `Builder#push` to create a section and `Builder#pop` to leave a section. Afterwards, the configuration can be built using one of two methods:
+
+Method    | Description
+:---      | :---
+`build`     | Creates the `ForgeConfigSpec`.
+`configure` | Creates a pair of the class holding the config values and the `ForgeConfigSpec`.
```
sciwhiz12: Misalignment of column separators:
```suggestion
 Method     | Description
 :---       | :---
`build`     | Creates the `ForgeConfigSpec`.
`configure` | Creates a pair of the class holding the config values and the `ForgeConfigSpec`.
```

docs/misc/config.md
```diff
@@ -0,0 +1,131 @@
...
+    ExampleConfig(ForgeConfigSpec.Builder builder) {
+        // Define values here in final fields
+    }
+
+    // Somewhere the constructor is accessible
+    static {
+        Pair<ExampleConfig, ForgeConfigSpec> pair = new ForgeConfigSpec.Builder()
+            .configure(ExampleConfig::new);
+        // Store pair values in some constant field
+    }
```
sciwhiz12: Per the style guide, indentation should be two spaces:
```suggestion
    ExampleConfig(ForgeConfigSpec.Builder builder) {
      // Define values here in final fields
    }

    // Somewhere the constructor is accessible
    static {
      Pair<ExampleConfig, ForgeConfigSpec> pair = new ForgeConfigSpec.Builder()
        .configure(ExampleConfig::new);
      // Store pair values in some constant field
    }
```

docs/misc/config.md
```diff
@@ -0,0 +1,131 @@
...
+  * Class Type: `Boolean`
+  * Method Name: `#define`
+
+Registering a Configuration
+---------------------------
+
+Once a `ForgeConfigSpec` has been built, it must be registered to allow Forge to load, track, and sync the configuration settings as required. Configurations should be registered in the mod constructor via `ModLoadingContext#registerConfig`. A configuration can be registered with a given type representing the side the config belongs to, the `ForgeConfigSpec`, and optionally a specific file name for the configuration.
+
+Type   | Loaded | Synced to Client | Stored                         | Default Suffix
+:---:  | :---:  | :---:            | :---:                          | :---
```
sciwhiz12: Something which we may want to look into the cause and solution is the misalignment between the column header and its values. The column values follow the alignment as configured, but the header remains left aligned.

image

docs/misc/config.md
```diff
@@ -0,0 +1,131 @@
...
+CLIENT | Client | ❌               | `.minecraft/config`            | `-client`
+COMMON | Both   | ❌               | `.minecraft/config`            | `-common`
+SERVER | Server | ✔️               | `<server_folder>/serverconfig` | `-server`
```
sciwhiz12: Perhaps, for acccessibility, we want to include the text "Yes" and "No" beside the emoji indicators?

We may want to expand the values text in the `Loaded` column, to something like "CLient side only", "On both sides", "Server side only"?

The last column could be renamed as `Default file suffix` for clarity.

The storage locations for `COMMON` and `SERVER` are a bit inaccurate: `COMMON` needs to include the dedicated server folder as well, and `SERVER` needs to be under the world/level's folder (named on the client, `world` on the dedicated server).

docs/misc/config.md
```diff
@@ -0,0 +1,131 @@
...
+
+!!! tip
+    Forge documents the [config types][type] within their codebase.
+
+```java
+// In the mod constructor with a ForgeConfigSpec CONFIG
+ModLoadingContext.get().registerConfig(Type.COMMON, CONFIG);
+```
```
sciwhiz12, line 111: Perhaps it may be better to move this code block example to before the table, since it seems a bit out of place after the admonition.

docs/misc/config.md
```diff
@@ -0,0 +1,131 @@
...
+
+Listening for Configuration Loading/Reloading
+---------------------------------------------
```
sciwhiz12: Perhaps a shorter section header would be `Configuration Events`?

docs/misc/config.md
```diff
@@ -0,0 +1,131 @@
...
+  * Class Type: `Enum<T>`
+  * Method Name: `#defineEnum`
+  * Additional Components:
+    * A getter to convert a string or integer into an enum
+    * A collection of the allowed values the configuration can be
+
+* **Boolean Values**
+  * Description: A `boolean` value
+  * Class Type: `Boolean`
+  * Method Name: `#define`
```
sciwhiz12: Looks like the second-level lists don't work with two-space indentation:

image

docs/misc/config.md
```diff
@@ -0,0 +1,131 @@
...
+
+Config values can be built with the provided contexts (if defined) using any of the `#define` methods.
+
+All config value methods take in at least two components:
+* A path representing the name of the variable: a `.` separated string representing the sections the config value is in
+* The default value when no valid configuration is present
+
+The `ConfigValue` specific methods take in two additional components:
+* A validator to make sure the deserialized object is valid
+* A class representing the data type of the config value
```
sciwhiz12: The absence of a space between the paragraph and the list is causing the list to not render properly.
image
```suggestion
All config value methods take in at least two components:

* A path representing the name of the variable: a `.` separated string representing the sections the config value is in
* The default value when no valid configuration is present

The `ConfigValue` specific methods take in two additional components:

* A validator to make sure the deserialized object is valid
* A class representing the data type of the config value
```

docs/misc/config.md
```diff
@@ -0,0 +1,131 @@
...
+---------------------------------------------
+
+Operations that occur whenever a config is loaded or reloaded can be done using the `ModConfigEvent$Loading` and `ModConfigEvent$Reloading` events. The events must be [registered][events] to the mod event bus.
+
+!!! warning
+    These events are called for all configurations for the mod; the `ModConfig` object provided should be used to denote which configuration is being loaded or reloaded.
+
+[toml]: https://toml.io/
+[nightconfig]: https://github.com/TheElectronWill/night-config
+[type]: https://github.com/MinecraftForge/MinecraftForge/blob/1.18.x/fmlcore/src/main/java/net/minecraftforge/fml/config/ModConfig.java#L108-L136
```
sciwhiz12: It may be better to make this a permalink to the line (commit rather than branch), for future-proofing.

sciwhiz12 (changes requested):

docs/misc/config.md
ChampionAsh5357: Seems to be a theme issue, so we should probably open an issue for it.

sciwhiz12: Tracked at #431.

docs/misc/config.md
```diff
@@ -63,64 +63,68 @@ The values themselves can be obtained using `ConfigValue#get`. The values are ad
...
 
-Listening for Configuration Loading/Reloading
----------------------------------------------
+Here is a list of the available configuration types:
+
+Type   | Loaded           | Synced to Client | Stored                         | Default File Suffix
+:---:  | :---:            | :---:            | :---:                          | :---
+CLIENT | Client Side Only | No               | `.minecraft/config`            | `-client`
+COMMON | On Both Sides    | No               | `<run_location>/config`            | `-common`
+SERVER | Server Side Only | Yes               | `<server_folder>/world/serverconfig` | `-server`
```
sciwhiz12: Table column separators needs to be aligned. Additionally, the `SERVER` type still needs to specify the correct path on the client side -- `.minecraft/saves/<level name>/serverconfig`.

sciwhiz12 (changes requested):

sciwhiz12 (approved):

PR #428, opened 2022-04-02
[1.18.x] PR 8527 Update: Vanilla Deferred Registers

ChampionAsh5357:
Updates the documentation to reflect the changes and renames of registering vanilla entries.

### DIFF
--- a/docs/concepts/registries.md
+++ b/docs/concepts/registries.md
@@ -45,10 +45,14 @@ public void registerBlocks(RegistryEvent.Register<Block> event) {
 
 ### Registries that aren't Forge Registries
 
-Due to some peculiarities of vanilla code, not all registries are wrapped by Forge. These can be static registries, like `RecipeType`, which are safe to use. There are also dynamic registries, like `ConfiguredFeature` and some other worldgen registries, which are typically represented in JSON. These objects should only be registered this way if there is another registry object that requires it. `DeferredRegister#create` has an overload which allows modders to specify the registry key of which vanilla registry to create a `RegistryObject` for.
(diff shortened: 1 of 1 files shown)


docs/concepts/registries.md
```diff
@@ -45,13 +45,15 @@ public void registerBlocks(RegistryEvent.Register<Block> event) {
...
-
-In this case, the value to be stored is `() -> Registry.register(...)`. For parameters, `Registry#register` takes the registry being added to, a `ResourceLocation` identifying the object to be registered, and the object itself.
+```java
+private static final DeferredRegister<RecipeType<?>> REGISTER = DeferredRegister.create(Registry.RECIPE_TYPE_REGISTRY, "examplemod");
 
-The appropriate time to populate these objects is in `FMLCommonSetupEvent`. Because vanilla registries are not guaranteed to be threadsafe, a call to `Supplier#get` should be inside a `ParallelDispatchEvent#enqueueWork` call. This evaluates the `Registry#register` call, which will register the object and allow it to be referenced freely.
+// As RecipeType is an interface, an anonymous class will be created for registering
+// Assume some recipe ExampleRecipe
+public static final RegistryObject<RecipeType<ExampleRecipe>> EXAMPLE_RECIPE_TYPE = REGISTER.register("example_recipe_type", () -> new RecipeType<>() {});
```
sciwhiz12, line 56: The anonymous subclass of `RecipeType` created by `RecipeType#register` overrides `toString` to return the name of the recipe type. I couldn't find whether this override is needed because of some registry/usage logic, but there's probably a good purpose to it. Should we change the example to override the method, or note this small difference here?

sciwhiz12 (changes requested):

docs/concepts/registries.md
ChampionAsh5357: The only real purpose of the `#toString` for recipe type is for easier debugging. It's not used for anything else afaik. I can make a note of the behavior difference though.

sciwhiz12 (approved):

PR #435, opened 2022-04-12
[1.18.x] A for Advancements

ChampionAsh5357:
Adds documentation on advancements and generating them through data providers.

This does not cover generating conditional advancements as currently the base architecture does not support it without moderate modification.


docs/datagen/server/advancements.md
```diff
@@ -0,0 +1,34 @@
...
+======================
+
+Advancements can be generated for a mod by subclassing `AdvancementProvider` and overriding `#registerAdvancements`. Advancements can either be created and supplied manually or, for convenience, created using `Advancement$Builder`.
+
+After implementation, the provider must be [added][datagen] to the `DataGenerator`.
+
+`Advancement$Builder`
+---------------------
+
+`Advancement$Builder` is a convenience implementation for creating `Advancement`s to generate. It allows the definition of the parent advancement, the display information, the rewards when the advancement has been completed, and the requirements to unlock the advancement. Only the requirements must be specified to create an `Advancement`.
```
sciwhiz12: > Only the requirements must be specified to create an `Advancement`.

The word "must" is too forceful (gives me the impresssion that one may only specify the requirements, contrary to the previous sentence), and could be replaced with "need" instead.
```suggestion
`Advancement$Builder` is a convenience implementation for creating `Advancement`s to generate. It allows the definition of the parent advancement, the display information, the rewards when the advancement has been completed, and the requirements to unlock the advancement. Only the requirements need to be specified to create an `Advancement`.
```

docs/datagen/server/advancements.md
```diff
@@ -0,0 +1,34 @@
...
+
+Although not required, there are a number of methods that are important to know of:
+
+Method | Description
+:---: | :---
+`parent` | Sets the advancement which this advancement is directly linked to. Can either specify the name of the advancement or the advancement itself if its generated by the modder.
+`display` | Sets the information to display to the chat, toast, and advancement screen.
+`rewards` | Sets the rewards obtained when this advancement is completed.
+`addCriterion` | Adds a condition to the advancement.
+`requirements` | Specifies if the conditions must all return true or at least one must return true. An additional overload can be used to mix-and-match those operations.
```
sciwhiz12: Table column separators should be aligned.
```suggestion
Method         | Description
:---:          | :---
`parent`       | Sets the advancement which this advancement is directly linked to. Can either specify the name of the advancement or the advancement itself if its generated by the modder.
`display`      | Sets the information to display to the chat, toast, and advancement screen.
`rewards`      | Sets the rewards obtained when this advancement is completed.
`addCriterion` | Adds a condition to the advancement.
`requirements` | Specifies if the conditions must all return true or at least one must return true. An additional overload can be used to mix-and-match those operations.
```

docs/datagen/server/advancements.md
```diff
@@ -0,0 +1,34 @@
+Advancement Generation
+======================
+
+Advancements can be generated for a mod by subclassing `AdvancementProvider` and overriding `#registerAdvancements`. Advancements can either be created and supplied manually or, for convenience, created using `Advancement$Builder`.
```
sciwhiz12: The first word "Advancements" should link to the page on advancements.

docs/resources/server/advancements.md
```diff
@@ -0,0 +1,173 @@
+Advancements
+============
+
+Advancements are tasks that can be achieved by the player which may advance the progress of the game. Advancements can trigger based on any action the player may be directly involved in.
+
+Data-Driven Advancements
+------------------------
```
sciwhiz12: This section header can be omitted, and the paragraph within can be included in the leading section.

docs/resources/server/advancements.md
```diff
@@ -0,0 +1,173 @@
...
+
+All advancement implementations within vanilla are data driven via JSON. This means that a mod is not necessary to create a new advancement, only a [Data pack][datapack]. A full list on how to create and put these advancements within the mod's `resources` folder can be found on the [Minecraft Wiki][wiki].
```
sciwhiz12: Capitalization of `Data pack` to `data pack`? Also, we could omit "folder" from "the mod's `resources` folder".

docs/resources/server/advancements.md
```diff
@@ -0,0 +1,173 @@
...
+```java
+// Where ID is the registry name of the trigger
+public ExampleTriggerInstance(EntityPredicate.Composite player, ItemPredicate item) {
+  super(ID, player);
+  // Store the item condition that must be met
+}
+```
+
+!!! note
+    Typically, trigger instances have a static constructor which allow these instances to be easily created for data generation.
```
sciwhiz12: It should likely be noted that another (likely) reason for the existence of these static factory methods is to allow static import of these methods in the relevant data providers.

docs/resources/server/advancements.md
```diff
@@ -0,0 +1,173 @@
...
+    // ...
+  ],
+  "function": "minecraft:example_function"
+}
+```
+
+Conditional Advancements
+------------------------
+
+Advancements can be [loaded conditionally and defaulted][conditional] depending on what information is present (mod loaded, item exists, etc.).
```
sciwhiz12: Perhaps this should be included in the lead section instead of added at the very end of the article?

sciwhiz12 (changes requested):

docs/resources/server/advancements.md
ChampionAsh5357: I don't see why not. I typically include them at the end since I order the forge added features in the way I come up with them.

sciwhiz12 (approved):

PR #436, opened 2022-04-14
[1.18.x] Generating Tags

ChampionAsh5357:
Adds the docs for creating `TagProvider`s.


docs/datagen/server/tags.md
```diff
@@ -0,0 +1,83 @@
+Tag Generation
+==============
+
+Tags can be generated for a mod by subclassing `TagsProvider` and implementing `#addTags`. After implementation, the provider must be [added][datagen] to the `DataGenerator`.
```
sciwhiz12: Perhaps `Tags` could link to the page about tags.

docs/datagen/server/tags.md
```diff
@@ -0,0 +1,83 @@
...
+!!! note
+    Typically, a provider will not call `#getOrCreateRawBuilder` directly unless a registry contains a representation of objects from a different registry (blocks have item representations to obtain the blocks in the inventory).
+
+When `#tag` is called, a `TagAppender` is created which acts as a chainable consumer of elements to add to the tag:
+
+Method           | Description
+:---:            | :---
+`add`            | Adds an object to a tag. 
+`addOptional`    | Adds an object to a tag through its name. If the object is not present, then the object will be skipped when loading.
+`addTag`         | Adds a tag to a tag. All elements within the inner tag are now a part of the outer tag.
```
sciwhiz12, line 20: Do we need to clarify that this does not add all entries from the tag in the given tag to this tag, but only adds a reference to that tag?

docs/datagen/server/tags.md
```diff
@@ -0,0 +1,83 @@
...
+`addOptionalTag` | Adds a tag to a tag through its name. If the tag is not present, then the tag will be skipped when loading.
+`replace` | When `true`, all other entries added to this tag from other datapacks will be discarded.
```
sciwhiz12: TO my knowledge, this needs clarification: all other entries from _lower priority_ datapacks, as higher priority datapacks can still add (or themselves `replace`) the contents of these tags.

docs/datagen/server/tags.md
```diff
@@ -0,0 +1,83 @@
...
+`remove` | Removes an object or tag from a tag.
```
sciwhiz12: Table column separators need alignment.
```suggestion
`addOptionalTag` | Adds a tag to a tag through its name. If the tag is not present, then the tag will be skipped when loading.
`replace`        | When `true`, all other entries added to this tag from other datapacks will be discarded.
`remove`         | Removes an object or tag from a tag.
```

docs/datagen/server/tags.md
```diff
@@ -0,0 +1,83 @@
...
+  .add(EXAMPLE_OBJECT) // Adds an object to the tag
+  .addOptional(new ResourceLocation("othermod", "other_object")) // Adds an object from another mod to the tag
+
+this.tag(EXAMPLE_TAG_2)
+  .addTag(EXAMPLE_TAG) // Adds a tag to the tag
+  .remove(EXAMPLE_OBJECT) // Removes an object from this tag
+```
+
+!!! important
+    If the mod's tags softly depends on another mod's tags, the other mods' tags should be referenced using the optional methods.
```
sciwhiz12: Perhaps it may be good to define what "softly depends" is, indirectly.
```suggestion
!!! important
    If the mod's tags softly depends on another mod's tags (the other mod may or may not be present at runtime), the other mods' tags should be referenced using the optional methods.
```

docs/datagen/server/tags.md
```diff
@@ -0,0 +1,83 @@
...
+
+```java
+public RecipeTypeTagsProvider(DataGenerator generator, ExistingFileHelper fileHelper) {
+  super(generator, Registry.RECIPE_TYPE, MOD_ID, fileHelper);
+}
+```
+
+### Forge Registry Tag Providers
+
+If a registry is wrapped or created by Forge, a provider can be created via a `ForgeRegistryTagsProvider` subclass instead:
```
sciwhiz12: This paragraph should also mention the possibility of custom mod-created registries.

sciwhiz12 (changes requested):

docs/datagen/server/tags.md
ChampionAsh5357: That was the idea of created by Forge, I just worded it badly.

docs/datagen/server/tags.md
ChampionAsh5357: I don't think so. There's no way to check if a tag is within another tag, only whether the elements of a tag are in both tags. So, mentioning the encoded format of reference won't affect the users unless they are generating it manually, which goes against the point of referencing the datagen docs.

sciwhiz12 (approved):

PR #437, opened 2022-04-14
[1.18.x] Around the World in Language Providers

ChampionAsh5357:
Adds the documentation for creating and using `LanguageProvider`s.


docs/datagen/client/localization.md
```diff
@@ -0,0 +1,27 @@
...
+this.addBlock(EXAMPLE_BLOCK, "Example Block");
+this.add("object.examplemod.example_object", "Example Object");
+```
+
+!!! tip
+    Localized names which contain alphanumeric values not in American English can be supplied as is. The provider automatically translates the characters into their unicode equivalents to be read by the game.
+
+    ```java
+    // Encdoded as 'Art\u00EDculo de ejemplo
+    this.addItem(EXAMPLE_ITEM, "Artículo de ejemplo");
```
sciwhiz12: For reference, what language is this, and from what was this translation taken?

sciwhiz12 (changes requested):

docs/datagen/client/localization.md
ChampionAsh5357: The Spanish 'es_es' locale

sciwhiz12: After a bit of consultation with @gigaherz and the official Crowdin translations for Spanish `es_es`, the proper and consistent translation for `Example item` is `Objeto de ejemplo`.

This means that we cannot use it as a example for how diacritics are serialized. However, since this is only an example, we needn't use an actual translation: we can use a dummy translation key, and fill in the value as something like `Example with a díacritic`.
```suggestion
    this.addItem("example.diacritic", "Example with a díacritic");
```

sciwhiz12 (approved):

PR #444, opened 2022-05-05
Update link from AdoptOpenJDK -> Adoptium

gdams:


ChampionAsh5357 (approved):
Making a note to keep track of adoptium/website-v2#373 for when deep link syntax is added back to make sure JDK shown in J17. Otherwise, this is fine.

PR #460, opened 2022-09-10
[1.19.x] Codecable

ChampionAsh5357:
Documentation on the basics of creating and using codecs.

### DIFF
--- a/docs/datastorage/codecs.md
+++ b/docs/datastorage/codecs.md
@@ -183,22 +183,22 @@ public class ClassB {
 }
 
 // Assume there is some codec A_CODEC
-public static final Codec<ClassB> B_CODEC = A_CODEC.xmap(ClassA:toB, ClassB:toA);
+public static final Codec<ClassB> B_CODEC = A_CODEC.xmap(ClassA::toB, ClassB::toA);
 ```
 
 If a type is partially equivalent, meaning that there are some restrictions during conversion, there are mapping functions which return a `DataResult` which can be used to return an error state whenever an exception or invalid state is reached.
 
-A => B | B => A | Transform Method
-:---:  | :---:  | :---
-Yes    | Yes    | `#xmap`
-Yes    | No     | `#flatComapMap`
-No     | Yes    | `#comapFlatMap`
-No     | No     | `#flatXMap`
+Is A Fully Equivalent to B | Is B Fully Equivalent to A | Transform Method
+:---:                      | :---:                      | :---
+Yes                        | Yes                        | `#xmap`
+Yes                        | No                         | `#flatComapMap`
+No                         | Yes                        | `#comapFlatMap`
+No                         | No                         | `#flatXMap`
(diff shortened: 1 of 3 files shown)


Commoble:
This all looks fine

docs/datastorage/codecs.md
```diff
@@ -0,0 +1,437 @@
...
+}
+
+// Another equivalent class
+public class ClassB {
+
+  public ClassA toA() { /* ... */ }
+}
+
+// Assume there is some codec A_CODEC
+public static final Codec<ClassB> B_CODEC = A_CODEC.xmap(ClassA:toB, ClassB:toA);
```
sciwhiz12: Method references are missing one more colon.
```suggestion
public static final Codec<ClassB> B_CODEC = A_CODEC.xmap(ClassA::toB, ClassB::toA);
```

docs/datastorage/codecs.md
```diff
@@ -0,0 +1,437 @@
...
+```
+
+If a type is partially equivalent, meaning that there are some restrictions during conversion, there are mapping functions which return a `DataResult` which can be used to return an error state whenever an exception or invalid state is reached.
+
+A => B | B => A | Transform Method
+:---:  | :---:  | :---
+Yes    | Yes    | `#xmap`
+Yes    | No     | `#flatComapMap`
+No     | Yes    | `#comapFlatMap`
+No     | No     | `#flatXMap`
```
sciwhiz12: This table implies (to me at least) at first glance that these methods do not allow conversion for the entries marked with `No`. Should we be concerned about this possible confusion?

docs/datastorage/codecs.md
```diff
@@ -0,0 +1,437 @@
...
+  [4, 5, 6], // BlockPos(4, 5, 6)
+  [7, 8, 9]  // BlockPos(7, 8, 9)
+]
+```
+
+List objects decoded using a list codec are stored in an **immutable** list. If a mutable list is needed, a [transformer] should be applied to the list codec.
+
+### Map
+
+A codec for a map of keys and value objects can be generated from two codecs via `Codec#unboundedMap`.
```
sciwhiz12: This only mentions `#unboundedMap` which is a map with an unbounded set of keys, but not `#simpleMap` which is a map for a bounded set of keys (as provided by a `Keyable`). This should either mention both, or at least clarify what an "unbounded map" means.

sciwhiz12 (changes requested):

docs/datastorage/codecs.md
ChampionAsh5357: `#simpleMap` is a bit complicated for users to understand since it compresses a map into a list where the number of keys are fixed and ordered. This is one of the few codecs I would rather avoid mentioning since it is likely most modders will never use it nor really understand it.

sciwhiz12 (approved):

PR #462, opened 2022-09-25
Intermission: Key Mappings

### LINKED ISSUE #461: Requesting documentation on KeyMapping.
The basics of key binding are probably a fundamental of writing any mod. Certainly every mod I have ever played (and there have been a lot) has custom keys.

I have just spent 2 and a half hours searching the internet and the forge forums for information on how to do this. Almost everything I have found refers to using a KeyBinding instance. KeyBinding has apparently been replaced by KeyMapping.

Every article and guide is incomplete, except for the very very old ones for very very old versions of forge (1.7.2).

I have just about figured out how to instantiate a KeyMapping object, although I have no idea how to associate it with user friendly text in the settings UI. The article that helped with that also mentioned that you need to register with a different event source for the key input event than you do for registering the keybinding.

I am really really surprised that the documentation has absolutely nothing on key input. I cannot imagine how anyone ever figured it out. Maybe Java works VERY differently from C++ (which I am more used to), but is there no documentation on any of the classes and interfaces? In C and C++ we use Doxygen on the header files to autogenerate documentation that often then gets published on the internet. Failing that we can go read the header files and infer how to use the functions (and see what functions are there). However Java packages are all precompiled object files and I have no idea how to get at the information on what functions are available and what parameters they take.

Many years ago I did a load of Java programming and I am pretty sure that all the classes and packages I used were properly documented.

Please add a basic keybinding to the example mod and/or to the documentation.

ChampionAsh5357:
*We interrupt the work on menu documentation to bring you an important announcement: key mappings are now documented on the official docs!*

This documents Key Mappings in it's entirety including use cases within a common environment.

Closes #461

### DIFF
--- a/docs/misc/keymappings.md
+++ b/docs/misc/keymappings.md
@@ -1,0 +1,150 @@
+# Key Mappings
+
+A key mapping, or key binding, defines a particular action that should be tied to an input: mouse click, key press, etc. Each action defined by a key mapping can be checked whenever the client can take an input. Furthermore, each key mapping can be assigned to any input through the [Controls option menu][controls].
+
+## Registering a `KeyMapping`
+
+A `KeyMapping` can be registered by listening to the `RegisterKeyMappingsEvent` on the [**mod event bus**][modbus] only on the physical client and calling `#register`.
+
+```java
+// In some physical client only class
+
+// Key mapping is lazily initialized so it doesn't exist until it is registered
+public static final Lazy<KeyMapping> EXAMPLE_MAPPING = Lazy.of(() -> /*...*/);
+
+// Event is on the mod event bus only on the physical client
+public void registerBindings(RegisterKeyBindingsEvent event) {
+  event.register(EXAMPLE_MAPPING.get());
+}
+```
+
+## Creating a `KeyMapping`
+
+A `KeyMapping` can be created using it's constructor. The `KeyMapping` takes in a [translation key][tk] defining the name of the mapping, the default input of the mapping, and the [translation key][tk] defining the category the mapping will be put within in the [Controls option menu][controls].
+
+!!! tip
+    A `KeyMapping` can be added to a custom category by providing a category [translation key][tk] not provided by vanilla. Custom category translation keys should contain the mod id.
+
+### Default Inputs
+
+Each key mapping has a default input associated with it. This is provided through `InputConstants$Key`. Each input consists of an `InputConstants$Type`, which defines what device is providing the input, and an integer, which defines the associated identifier of the input on the device.
+
+Vanilla provides three types of inputs: `KEYSYM`, which defines a keyboard through the provided `GLFW` key tokens, `SCANCODE`, which defines a keyboard through the platform-specific scancode, and `MOUSE`, which defines a mouse.
+
+!!! note
+    It is highly recommended to use `KEYSYM` over `SCANCODE` for keyboards as `GLFW` key tokens are not tied to any particular system. You can read more on the [GLFW docs][keyinput].
+
+The integer is dependent on the type provided. All input codes are defined in `GLFW`: `KEYSYM` tokens are prefixed with `GLFW_KEY_*` while `MOUSE` codes are prefixed with `GLFW_MOUSE_*`.
+
+```java
+new KeyMapping(
+  "key.examplemod.example1", // Will be localized using this translation key
+  InputConstants.Type.KEYSYM, // Default mapping is on the keyboard
+  GLFW.GLFW_KEY_P, // Default key is P
+  "key.categories.misc" // Mapping will be in the misc category
+)
+```
+
+!!! note
+    If the key mapping should not be mapped to a default, the input should be set to `InputConstants#UNKNOWN`. The vanilla constructor will require you to extract the input code via `#getValue` while the Forge constructor can be supplied the raw input field.
+
+### `IKeyConflictContext`
+
+Not all mappings are used in every context. Some mappings are only used in a GUI, while others are only used purely in game. To avoid mappings of the same key used in different contexts conflicting with each other, an `IKeyConflictContext` can be assigned.
+
+Each conflict context contains two methods: `#isActive`, which defines if the mapping can be used in the current game state, and `#conflicts`, which defines whether the mapping conflicts with a key in the same or different conflict context.
+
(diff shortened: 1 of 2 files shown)


docs/misc/keymappings.md
```diff
@@ -0,0 +1,150 @@
...
+  event.register(EXAMPLE_MAPPING.get());
+}
+```
+
+## Creating a `KeyMapping`
+
+A `KeyMapping` can be created using it's constructor. The `KeyMapping` takes in a [translation key][tk] defining the name of the mapping, the default input of the mapping, and the [translation key][tk] defining the category the mapping will be put within in the [Controls option menu][controls].
+
+!!! tip
+    A `KeyMapping` can be added to a custom category by providing a category [translation key][tk] not provided by vanilla. Custom category translation keys should contain the mod id.
```
sciwhiz12: Should we provide an example or template of a custom category translation key?

docs/misc/keymappings.md
```diff
@@ -0,0 +1,150 @@
...
+new KeyMapping(
+  "key.examplemod.example1", // Will be localized using this translation key
+  InputConstants.Type.KEYSYM, // Default mapping is on the keyboard
+  GLFW.GLFW_KEY_P, // Default key is P
+  "key.categories.misc" // Mapping will be in the misc category
+)
+```
+
+!!! note
+    If the key mapping should not be mapped to a default, the input should be set to `InputConstants#UNKNOWN`. The vanilla constructor will require you to extract the input code via `#getValue` while the Forge constructor can be supplied the raw input field.
```
sciwhiz12: `#getValue` should be qualified to `InputConstants$Key` for clarity.
```suggestion
    If the key mapping should not be mapped to a default, the input should be set to `InputConstants#UNKNOWN`. The vanilla constructor will require you to extract the input code via `InputConstants$Key#getValue` while the Forge constructor can be supplied the raw input field.
```

docs/misc/keymappings.md
```diff
@@ -0,0 +1,150 @@
...
+)
+```
+
+## Checking a `KeyMapping`
+
+A `KeyMapping` can be checked to see whether it has been clicked. Depending on when, the mapping can be used in a conditional to apply the associated logic.
+
+### Within the Game
+
+Within the game, a mapping should be checked by listening to `ClientTickEvent` on the [**forge event bus**][forgebus] and checking `KeyMapping#consumeClick` within a while loop. `#consumeClick` will return `true` only the number of times the input was performed and not already previously handled, so it won't infinitely stall the game.
```
sciwhiz12: "forge" should be capitalized. Same should be done in further references to the Forge event bus.
```suggestion
Within the game, a mapping should be checked by listening to `ClientTickEvent` on the [**Forge event bus**][forgebus] and checking `KeyMapping#consumeClick` within a while loop. `#consumeClick` will return `true` only the number of times the input was performed and not already previously handled, so it won't infinitely stall the game.
```

docs/misc/keymappings.md
```diff
@@ -0,0 +1,150 @@
...
+  if (EXAMPLE_MAPPING.get().isActiveAndMatches(InputConstants.TYPE.MOUSE.getOrCreate(button))) {
+    // Execute logic to perform on mouse click here
+    return true;
+  }
+  return super.mouseClicked(x, y, button);
+} 
+```
+
+!!! note
+    If you do not own the screen which you are trying to check a mouse for, you can listen to the `Pre` or `Post` events of `ScreenEvent$MouseButtonPressed` on the [**forge event bus**][forgebus] instead.
```
sciwhiz12: Given the proximity and similarity of the two notes here, the "mouse" in this note and the "key" in the previous note should be emphasized.

sciwhiz12 (changes requested):

sciwhiz12 (approved):

JontomXire:
Amazingly quick, and very well written, thankyou.

Only one question: What packages need to be imported for all this to work? I am guessing just the package that contains KeyMapping. Maybe also the one that contains RegisterKeyBindingsEvent?

ChampionAsh5357:
> What packages need to be imported for all this to work? I am guessing just the package that contains KeyMapping. Maybe also the one that contains RegisterKeyBindingsEvent?

It'll depend what you're attempting to do. You can import classes as you see fit (most likely using the autofill function of your IDE when calling a specific class). If you have any more questions, ask on the forums or discord after making a solid attempt at implementation yourself.

JontomXire:
I was suggesting it as an improvement to the documentation. The MSDN documentation, and the man pages always list the include file for a given class or function. In fact I don't remember seeing any C or C++ documentation that doesn't. Also, the Java documentation (for example, picking at random, https://docs.oracle.com/javase/7/docs/api/javax/swing/border/AbstractBorder.html) lists the java object which implicitly contains the package that needs to be imported.

This documentation is primarily about the KeyMapping class with mentions of other objects and functions. I was suggesting that it may be useful for other people if the import packages or full Java object path were specified, at least for the Forge stuff that isn't part of standard Java.

ChampionAsh5357:
I'm well aware of how the C/C++ documentation typically handles those cases. Additionally, I'm also aware at how Java handles the generated javadocs.

That said, this does not really fall into the traditional definition of docs. It is more of a quick reference to understanding the concepts of vanilla and it's Forge additions. Key mappings are a purely vanilla system with some Forge hooks for additional features.

In addition, while having imports are useful, they are nothing more than syntactic sugar in Java. When using a proper IDE, you can almost always search the necessary class to add without the full string. This is especially the case since most classes in vanilla are almost uniquely named.

Modders without the necessary knowhow of the language or how to do their own research will be stymied. Modders are required to have the relevant background and critical thinking skills to receive support from us, so this would be considered a necessary test of skill.

If you would like a more proper documentation on vanilla stuff, I would checkout Parchment, which adds in parameter names and javadocs. Forge documentation is typically provided as a deprecated tag to the relevant code, or documented on the class itself.

ChampionAsh5357:
I should probably clarify: I'm not trying to discredit your idea, it's purely a good way to force people to read the code and understand it's complexities themselves after getting the gist from a quick guide.

However, I will take a look into finding a tooltip system such that on hover the full path is revealed.

JontomXire:
I appreciate your feedback.

I don't actually use an IDE myself. As you say "it's purely a good way to force people to read the code and understand it's complexities themselves", so in the same way I prefer to read the documentation and fully understand what the parameters do, rather than have an IDE prompt me as to how to fill in the parameters to a function.

I also apologise for my lack of understanding. I am not new to Java but am new to modding and am having a lot of difficulty finding any proper documentation at all.

Thanks for your help. I will move to the forums now.

PR #464, opened 2022-10-10
[1.16.5] added Recipe Generator

Brandonbr1:


ChampionAsh5357:
This does not add any additional information, links a cite which is unknown and probably not good at teaching people how to make a recipe provider, and is for an outdated version.

Closed for the above reasons.

PR #465, opened 2022-10-10
[1.19] Menus and Screens: An HCI Research Paper

ChampionAsh5357:
Adds in the documentation relating to creating menus and screens.

I figured it's probably better to keep people up to date on what I'm working on since I tend not to have a lot of time for docs recently. So, here is what I currently have, and I'll keep working on it when I can.

Things to do:

- [x] Menus
- [x] Screen
- [x] GuiEventListener/ContainerEventHandler
- [x] AbstractContainerScreen
- [x] Widget


eerussianguy:
Not terribly interesting, but it's safe to store menu providers in a static final field rather than creating it each time, though that's not the vanilla pattern.

ChampionAsh5357:
It could also be a method on an instance as well. Though I don't believe there's any difference in execution speed since it gets compiled down to the same thing.

docs/gui/screens.md
```diff
@@ -0,0 +1,356 @@
...
+    );
+}
+```
+
+!!! warning
+    `MenuScreens#register` is not thread-safe, so it needs to be called inside `#enqueueWork` provided by the parallel dispatch event.
+
+[menus]: ./menus.md
+[network]: ../networking/index.md
+[screen]: #TODO
```
sciwhiz12: `#TODO` :shipit:

docs/gui/screens.md
```diff
@@ -0,0 +1,356 @@
...
+### Strings
+
+Strings are drawn through its `Font`, typically consisting of their own shaders for normal, see through, and offset mode. There are two alignment of strings that can be rendered, each with a back shadow: a left-aligned string (`#drawString`) and a center-aligned string (`#drawCenteredString`). These both take in the font the string will be rendered in, the string to draw, the x coordinate representing the left or center of the string respectively, the top y coordinate, and the color.
+
+!!! note
+    Strings should typically be passed in as [`Component`s][component] as they handle a variety of usecases, including the two other overloads of the method.
+
+### Textures
+
+Textures are drawn through blitting, hence the method name `#blit`, which, for this purpose, copies the bits of an image and draws them directly to the screen. These are drawn through a position texture shader. While there are many different `#blit` overloads, we will only discuss two: the instance `#blit` and the static `#blit` the instance method delegates to.
```
sciwhiz12:
```suggestion
Textures are drawn through blitting, hence the method name `#blit`, which, for this purpose, copies the bits of an image and draws them directly to the screen. These are drawn through a position texture shader. While there are many different `#blit` overloads, we will only discuss two: the instance `#blit` and the static `#blit` which the instance method delegates to.
```

docs/gui/screens.md
```diff
@@ -0,0 +1,356 @@
...
+
+## GuiEventListener
+
+Any screen rendered in Minecraft implements `GuiEventListener`. `GuiEventListener`s are responsible for handling user interaction with the screen. These include inputs from the mouse (movement, clicked, released, dragged, scrolled, mouseover) and keyboard (pressed, released, typed). Each method returns whether the associated action affected the screen successfully. Widgets like buttons, chat boxes, lists, etc. also implement this interface.
+
+### ContainerEventHandler
+
+Almost synonymous with `GuiEventListener`s are their subtype: `ContainerEventHandler`s. These are responsible for handling user interaction on screens which contain widgets, managing which is currently focused and how the associated interactions are applied. `ContainerEventHandler`s add three additional features: interactable children, dragging, and focusing.
+
+Event handlers held children which was used to determine the interaction order of elements. During the mouse event handlers (excluding dragging), the first child in the list that the mouse was over would have their logic executed.
```
sciwhiz12: This paragraph seems a bit... off. At the very least, the first sentence has an odd choice of verb tense. We should use either the present or future tense for verbs, like so:

```suggestion
Event handlers hold children which are used to determine the interaction order of elements. During the mouse event handlers (excluding dragging), the first child in the list that the mouse was hovering over will have their logic executed.
```

docs/gui/screens.md
```diff
@@ -0,0 +1,356 @@
...
+
+Dragging is implemented within `#mouseClicked` and `#mouseReleased` allowing for logic to be more precisely executed when the mouse was dragging the element.
+
+Focusing allowed for a specific child to be selected to execute logic. These included during keyboard events and when the mouse was being dragged. Focus of which element was typically set through `#setFocused` or, when the screen was being opened, `#setInitialFocus`. In addition, interactable children could be cycled through using `#changeFocus`, selecting the next child in the list, or the previous child when shift was held down.
```
sciwhiz12: These also seem to have some odd choices for verb tenses.

docs/gui/menus.md
```diff
@@ -0,0 +1,343 @@
...
+  this(containerId, playerInventory);
+}
+
+// Server menu constructor
+public MyMenu(int containerId, Inventory playerInventory) {
+  // ...
+}
+```
+
+Each menu implementation must implement two methods: `#stillValid` and [`#quickMoveStack`][qms]. `#stillValid` determines whether the menu should remain open for a given player. This is typically directed to the Forge deprecated `Container#stillValid`. As an alternative, a boolean function which takes in the player will work just as well. The client menu must always return `true` for this method. The vanilla implementation checks whether the player is within eight blocks of where the data storage object is located.
```
sciwhiz12: Perhaps "Forge deprecated" should be hyphenated, to signal more clearly that it is a method which Forge deprecates, rather than a Forge method which is also deprecated?
```suggestion
Each menu implementation must implement two methods: `#stillValid` and [`#quickMoveStack`][qms]. `#stillValid` determines whether the menu should remain open for a given player. This is typically directed to the Forge-deprecated `Container#stillValid`. As an alternative, a boolean function which takes in the player will work just as well. The client menu must always return `true` for this method. The vanilla implementation checks whether the player is within eight blocks of where the data storage object is located.
```

docs/gui/menus.md
```diff
@@ -0,0 +1,343 @@
...
+```
+
+### Data Synchronization
+
+Some data needs to be present on both the server and the client to display to the player. To do this, the menu implements a basic layer of data synchronization such that whenever the current data does not match the data last synced to the client. For players, this is checked every tick.
+
+Minecraft supports two forms of data synchronization by default: `ItemStack`s via `Slot`s and integers via `DataSlot`s. `Slot`s and `DataSlot`s are views which hold references to data storages that can be be modified by the player in a screen, assuming the action is valid. These can be added to a menu within the constructor through `#addSlot` and `#addDataSlot`.
+
+!!! note
+    Since `Container`s used by `Slot`s are deprecated by Forge in favor of using the [`IItemHandler` capability][cap], the rest of the explanation will revolve around using `SlotItemHandler`: the capability variant.
```
sciwhiz12: The nouns around that colon should be switched around.
```suggestion
    Since `Container`s used by `Slot`s are deprecated by Forge in favor of using the [`IItemHandler` capability][cap], the rest of the explanation will revolve around using the capability variant: `SlotItemHandler`.
```

docs/gui/menus.md
```diff
@@ -0,0 +1,343 @@
...
+
+A `SlotItemHandler` contains four parameters: the `IItemHandler` representing the inventory the stacks are within, the index of the stack this slot is specifically representing, and the x and y position of where the top-left position of the slot will render on the screen relative to `AbstractContainerScreen#leftPos` and `#topPos`. The client menu constructor should always supply an empty instance of an inventory of the same size.
+
+In most cases, any slots the menu contains is first added, followed by the player's inventory, and finally concluded with the player's hotbar. To access any individual `Slot` from the menu, the index must be calculated based upon the order of which slots were added.
+
+A `DataSlot` is an abstract class which should implement a getter and setter to reference the data stored in the data storage object.The client menu constructor should always supply a new instance via `DataSlot#standalone`.
```
sciwhiz12: Missing space.
```suggestion
A `DataSlot` is an abstract class which should implement a getter and setter to reference the data stored in the data storage object. The client menu constructor should always supply a new instance via `DataSlot#standalone`.
```

docs/gui/menus.md
```diff
@@ -0,0 +1,343 @@
...
+
+These, along with slots, should be recreated every time a new menu is initialized.
+
+!!! warning
+    Although a `DataSlot` may hold enough data for an integer, only a short is sent across the network.
```
sciwhiz12: Rephrased it a bit with a bit more emphasis to draw the eye. Also, what happens when a value larger than a short is provided in a DataSlot? Is it truncated to fit the bits of a short, or can some form of overflow/underflow occur?

```suggestion
!!! warning
    Although a `DataSlot` stores an integer, it is effectively limited to a **short** (-32768 to 32767) because of how it sends the value across the network.
```

docs/gui/menus.md
```diff
@@ -0,0 +1,343 @@
...
+  // Add data slots for handled integers
+  this.addDataSlot(dataSingle);
+
+  // ...
+}
+```
+
+#### `ContainerData`
+
+If multiple integers need to be synced to the client, a `ContainerData` can be used to reference the integers instead. This interface functions as an index lookup such that each index represents a different integer. `ContainerData`s can also be constructed in the data object itself if the `ContainerData` is added to the menu through `#addDataSlots`. The method creates a new `DataSlot` for the amount of data specified by the interface. The client menu constructor should always supply a new instance via `SimpleContainerData`.
```
sciwhiz12, line 150: These also fall to the same caveat as `DataSlot`s with regards to the effectively `short` data size, right?

docs/gui/menus.md
```diff
@@ -0,0 +1,343 @@
...
+    /*
+    The following quick move logic can be simplified to if in data inventory,
+    try to move to player inventory/hotbar and vice versa for containers
+    that cannot transform data (e.g. chests).
+    */
+
+    // If the quick move was performed on the data inventory result slot
+    if (quickMovedSlotIndex == 0) {
+      // Try to move the result slot into the player inventory/hotbar
+      if (!this.moveItemStackTo(rawStack, 5, 41, true)) {
```
sciwhiz12, line 214: `moveItemStackTo`'s slot end range is exclusive?

sciwhiz12 (changes requested):

docs/gui/menus.md
```diff
@@ -169,10 +169,15 @@ public MyMenuAccess(int containerId, Inventory playerInventory, ContainerData da
...
 ```
 
+!!! warning
+    As `ContainerData` delegates to `DataSlot`s, these are also limited to a **short** (-32768 to 32767).
+
 #### `#quickMoveStack`
 
 `#quickMoveStack` is the second method that must be implemented by any menu. This method is called whenever a stack has been shift-clicked, or quick moved, out of its current slot until the stack has been fully moved out of its previous slot or there is no other place for the stack to go. The method returns a copy of the stack in the slot being quick moved.
 
+Stacks are typically moved between slots using `#moveItemStackTo`, which moves the stack into the first available slot. It takes in the stack to be moved, the starting slot index (inclusive) to try and move the stack to, the last slot index (exclusive), and whether to check the slots in ascending (when false) or descending (when true) order.
```
sciwhiz12: Just to be clear, ascending means from the starting index to the ending index, and descending means from the ending index to the starting index? 

Also, there should be better parallelism: either "starting slot index" and "ending slot index", or "first slot index" and "last slot index".

As a matter of style, should we always monospace `true` and `false`? If not, when do we monospace them?

docs/gui/screens.md
```diff
@@ -69,11 +69,11 @@ Any screen rendered in Minecraft implements `GuiEventListener`. `GuiEventListene
...
 
-Dragging is implemented within `#mouseClicked` and `#mouseReleased` allowing for logic to be more precisely executed when the mouse was dragging the element.
+Dragging an element with the mouse, implemented via `#mouseClicked` and `#mouseReleased`, provides more precisely executed logic.
 
-Focusing allowed for a specific child to be selected to execute logic. These included during keyboard events and when the mouse was being dragged. Focus of which element was typically set through `#setFocused` or, when the screen was being opened, `#setInitialFocus`. In addition, interactable children could be cycled through using `#changeFocus`, selecting the next child in the list, or the previous child when shift was held down.
+Focusing allows for a specific child to be checked first and handled during an event's execution, such during keyboard events or dragging the mouse. Focus is typically set through `#setFocused` or, when the screen is being opened, `#setInitialFocus`. In addition, interactable children can be cycled using `#changeFocus`, selecting the next child in the list, or the previous child if the shift key is held down.
```
sciwhiz12: Missing `as` in `such during`:
```suggestion
Focusing allows for a specific child to be checked first and handled during an event's execution, such as during keyboard events or dragging the mouse. Focus is typically set through `#setFocused` or, when the screen is being opened, `#setInitialFocus`. In addition, interactable children can be cycled using `#changeFocus`, selecting the next child in the list, or the previous child if the shift key is held down.
```

sciwhiz12 (changes requested):

docs/gui/menus.md
ChampionAsh5357: I try to always monospace them, just forgot this time.

sciwhiz12 (approved):

PR #470, opened 2022-12-23
[1.19.x] Fix KeyMapping registration code

Maaxed:
This PR fixes the code example for registering a KeyMapping, by replacing ``RegisterKeyBindingsEvent`` with ``RegisterKeyMappingsEvent`` and adds the ``@SubscribeEvent`` annotation


ChampionAsh5357 (approved):
I will mention that the `SubscribeEvent` annotation is wholly unnecessary depending on the method of registration. I typically use different methods to give an understanding that it's not needed and depends on the registry method, but it has no effect on the outcome anyways, so that's fine.

PR #476, opened 2023-02-01
[1.19.x] Datapack Registry Object Providers

ChampionAsh5357:
Adds the documentation on how to register datapack registry objects. Not much else to say.

### DIFF
--- a/docs/datagen/server/dynamicregistries.md
+++ b/docs/datagen/server/dynamicregistries.md
@@ -1,0 +1,128 @@
+Dynamic Registry Object Generation
+==================================
+
+Dynamic registry objects can be generated for a mod by constructing a new `DatapackBuiltinEntriesProvider` and providing a `RegistrySetBuilder` with the new objects to register. The provider must be [added][datagen] to the `DataGenerator`.
+
+!!! note
+    `DatapackBuiltinEntriesProvider` is a Forge extension on top of `RegistriesDatapackGenerator` which properly handles referencing existing dynamic registry objects without exploding the entry. So, this documentation will use `DatapackBuiltinEntriesProvider`.
+
+```java
+// On the MOD event bus
+@SubscribeEvent
+public void gatherData(GatherDataEvent event) {
+    event.getGenerator().addProvider(
+        // Tell generator to run only when server data are generating
+        event.includeServer(),
+        output -> new DatapackBuiltinEntriesProvider(
+          output,
+          event.getLookupProvider(),
+          // The builder containing the dynamic registry objects to generate
+          new RegistrySetBuilder().add(/* ... */),
+          // Set of mod ids to generate the dynamic registry objects of
+          Set.of(MOD_ID)
+        )
+    );
+}
+```
+
+`RegistrySetBuilder`
+--------------------
+
+A `RegistrySetBuilder` is responsible for building all dynamic registry objects to be used within the game. The builder can add a new entry for a registry, which can then register objects to that registry.
+
+First, a new instance of a `RegistrySetBuilder` can be initialized by calling the constructor. Then, the `#add` method can be called to handle a specific registry for registration. The `#add` method takes in the `ResourceKey` of the registry, a `RegistryBootstrap` consumer containing the `BootstapContext` to register the objects, and an optional `Lifecycle` argument to indicate the registry's current lifecycle status.
+
+```java
+new RegistrySetBuilder()
+  // Create configured features
+  .add(Registries.CONFIGURED_FEATURE, bootstrap -> {
+    // Register configured features here
+  })
+  // Create placed features
+  .add(Registries.PLACED_FEATURE, bootstrap -> {
+    // Register placed features here
+  });
+```
+
+!!! note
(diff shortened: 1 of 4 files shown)


docs/datagen/server/dynamicregistries.md
```diff
@@ -0,0 +1,128 @@
...
+    );
+}
+```
+
+`RegistrySetBuilder`
+--------------------
+
+A `RegistrySetBuilder` is responsible for building all dynamic registry objects to be used within the game. The builder can add a new entry for a registry, which can then register objects to that registry.
+
+First, a new instance of a `RegistrySetBuilder` can be initialized by calling the constructor. Then, the `#add` method can be called to handle a specific registry for registration. The `#add` method takes in the `ResourceKey` of the registry, a `RegistryBootstrap` consumer containing the `BootstapContext` to register the objects, and an optional `Lifecycle` argument to indicate the registry's current lifecycle status.
```
Matyrobbrt:
```suggestion
First, a new instance of a `RegistrySetBuilder` can be initialized by calling the constructor. Then, the `#add` method (which takes in the `ResourceKey` of the registry, a `RegistryBootstrap` consumer containing the `BootstapContext` to register the objects, and an optional `Lifecycle` argument to indicate the registry's current lifecycle status) can be called to handle a specific registry for registration.
```

docs/datagen/server/dynamicregistries.md
```diff
@@ -0,0 +1,128 @@
...
+  });
+```
+
+!!! note
+    Dynamic registries created through Forge can also generate their objects using this builder by also passing in the associated `ResourceKey`.
+
+Registering with `BootstapContext`
+----------------------------------
+
+The `BootstapContext` provided by the builder can be used to register an object by calling the `#register` method. The `#register` method takes in the `ResourceKey` representing the registry name of the object, the object to register, and an optional `Lifecycle` argument to indicate the registry object's current lifecycle status.
```
Matyrobbrt:
```suggestion
The `#register` method in the `BootstapContext` provided by the builder can be used to register objects. It takes in the `ResourceKey` representing the registry name of the object, the object to register, and an optional `Lifecycle` argument to indicate the registry object's current lifecycle status. 
```

docs/datagen/server/dynamicregistries.md
```diff
@@ -0,0 +1,128 @@
...
+  // Create configured features
+  .add(Registries.CONFIGURED_FEATURE, bootstrap -> {
+    // Register configured features here
+    bootstrap.register(
+      // The resource key for the configured feature
+      EXAMPLE_CONFIGURED_FEATURE,
+      new ConfiguredFeature<>(
+        Feature.ORE, // Create an ore feature
+        new OreConfiguration(
+          Collections.emptyList(), // Does nothing
```
Matyrobbrt: I'd recommend using `List#of` instead.

docs/datagen/server/dynamicregistries.md
```diff
@@ -0,0 +1,128 @@
...
+  })
+  // Create placed features
+  .add(Registries.PLACED_FEATURE, bootstrap -> {
+    // Register placed features here
+  });
+```
+
+### Dynamic Registry Object Lookup
+
+Sometimes dynamic registry objects may want to use other dynamic registry objects or tags containing dynamic registry objects. In those cases, you can look up another dynamic registry using `BootstapContext#lookup` to get a `HolderGetter`. From there, you can get a `Holder$Reference` to the dynamic registry object or a `HolderSet$Named` for the tag via `#getOrThrow` by passing in the associated key.
```
Matyrobbrt: Could the repetition of `dynamic registry objects` in the first phrase be avoided?

docs/datagen/server/dynamicregistries.md
```diff
@@ -0,0 +1,128 @@
...
+
+    // Get configured feature registry
+    HolderGetter<ConfiguredFeature<?, ?>> configured = bootstrap.lookup(Registries.CONFIGURED_FEATURE);
+
+    bootstrap.register(
+      // The resource key for the placed feature
+      EXAMPLE_PLACED_FEATURE,
+      new PlacedFeature(
+        configured.getOrThrow(EXAMPLE_CONFIGURED_FEATURE), // Get the configured feature
+        Collections.emptyList() // and do nothing to the placement location
```
Matyrobbrt: `List#of`

Matyrobbrt (changes requested):
Should this guide reference dynamic regs as "Datapack Registries" instead of "Dynamic Registries", for the sake of consistency with class names?

docs/datagen/server/dynamicregistries.md
ChampionAsh5357: Sure, doesn't matter

docs/datagen/server/dynamicregistries.md
ChampionAsh5357: I don't think so since otherwise a reader would likely think this would apply to static registry objects as well, which is not the case. I think explicit here is best.

Matyrobbrt (approved):

sciwhiz12 (approved):

PR #484, opened 2023-03-11
Clarify src folder required to get started

paylett:
Explicitly stating to include the example src folder, or the test run step later does not work.

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -1,6 +1,8 @@
 Contributing to This Documentation
 ==================================
 
+Contribute to this documentation at [GitHub].
+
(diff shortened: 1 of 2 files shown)


ChampionAsh5357:
> Explicitly stating to include the example src folder, or the test run step later does not work.

I mean...the example src folder doesn't need to be included. It could be any organizational structure for the `src/<sourceSet>/java|resources` folders. The text is not well worded, but the solution here would be to explain it better rather than say copy the example src folder, which I plan to do in #486 unless you want to take up the task of doing so for this page.

ChampionAsh5357:
Superseded by #490

Issue #486, opened 2023-03-16
Improvements Towards Explanations for New Modders


ChampionAsh5357:
This is something I've seen for a while on the forums and on discord to which no one creates an issue on the docs so that I remember to address this. As such, since I can't really use my computer right now anyways, I'm creating the necessary issue so I can remember to address it once I get back.

This will be done after the 1.19.4 update.

AdamRaichu:
I'm considering starting to learn forge (I've done basics with fabric), and something like this would be super helpful to me.

PR #490, opened 2023-03-29
[1.19.x] Rewrite Getting Started

ChampionAsh5357:
Rewrites the getting started section to be more friendly to new modders or those who do not have the relevant experience in minecraft modding. Adds explicit mentions against those who do not know the languages or tools they need to use.

Making this a draft since maty told me to work on FG5 first, so...yeah. Here's just keeping things open on how far I've gotten.

Part of #486

Supersedes #484

* [x] index.md
* [x] contributing.md
* gettingstarted
    * [x] index.md
    * [x] structuring.md
    * [x] versioning.md


ChampionAsh5357:
Ok, this should be ready for review. I've restructured it to be more like the wiki without any of the extra redundancies.

docs/gettingstarted/index.md
```diff
@@ -1,60 +1,113 @@
...
+```text
+com
+- example (group package)
+  - mymod (the mod id)
+    - MyMod.java (renamed ExampleMod.java)
+```
+
+#### Version
+
+Set the `version` property to the current version of your mod. We recommend using a [variation of semantic versioning][semver].
```
Matyrobbrt: Forge uses maven versioning and not semantic versioning. While the two are similar in basic examples, they have some substantial differences.

docs/gettingstarted/index.md
```diff
@@ -1,60 +1,113 @@
...
+
+```gradle
+// In some build.gradle
+group = 'com.example'
+```
+
+The packages within your java source (`src/main/java`) should also now conform to this structure, with an inner package representing the mod id:
+
+```text
+com
```
Matyrobbrt, line 91: Should maybe use TLD and domain. Additionally, the guide should recommend using a domain you own, like a github pages one.

docs/gettingstarted/modfiles.md
```diff
@@ -0,0 +1,160 @@
...
+`loaderVersion`      | string  | **mandatory** | The acceptable version range of the language loader, expressed as a [Maven Version Range][mvr]. For `javafml` and `lowcodefml`, the version is the major version of the Forge version. | `"[45,)"`
+`license`            | string  | **mandatory** | The license the mod(s) in this JAR are provided under. It is suggested that this is set to the [SPDX identifier][spdx] you are using and/or a link to the license. You can visit https://choosealicense.com/ to help pick the license you want to use. | `"MIT"`
+`showAsResourcePack` | boolean | `false`       | When `true`, the mod(s)'s resources will be displayed as a separate resource pack on the 'Resource Packs' menu, rather than being combined with the 'Mod resources' pack. | `true`
+`services`           | array   | `[]`          | An array of services your mod uses. This is consumed as part of the created module for the mod from Forge's implementation of the Java Platform Module System. | `["net.minecraftforge.forgespi.language.IModLanguageProvider"]`
```
Matyrobbrt: The services are detected from the declared service files, usage of this property should be avoided.

docs/gettingstarted/structuring.md
```diff
@@ -1,124 +1,81 @@
...
+
+```
+a.jar
+  - com.example.ExampleClass
+b.jar
+  - com.example.ExampleClass // This class will not normally be loaded
 ```
 
-If any string is specified as `${file.jarVersion}`, Forge will replace the string with the **Implementation Version** specified in your jar manifest at runtime. Since the user development environment has no jar manifest to pull from, it will be `NONE` instead. As such, it is usually recommended to leave the `version` field alone. Here is a table of attributes that may be given to a mod, where `mandatory` means there is no default and the absence of the property causes an error.
+This is even more relevant when it comes to loading modules. If there are class files in two packages under the same name in separate modules, then only the first package will be loaded.
```
Matyrobbrt: This is false, duplicate packages across modules will result in a hard failure at startup, as mod modules are exported to the game and each other.

Matyrobbrt (changes requested):

docs/gettingstarted/index.md
ChampionAsh5357: I moved that over to the packaging section in structuring. It recommends all this information. I just didn't want to repeat it twice, so I put a link to that section instead. Would you recommend a different structure?

docs/gettingstarted/index.md
ChampionAsh5357: What does this have to do with mods? Afaik, we have always recommended using semantic versioning for mods: https://docs.minecraftforge.net/en/1.19.x/gettingstarted/versioning/

Do you want me to rewrite this to reference maven versioning instead because of the checker?

docs/gettingstarted/modfiles.md
ChampionAsh5357: Why do we have the property if it should be avoided regardless?

docs/gettingstarted/index.md
Matyrobbrt: My main problem is the discoverability. Using domains you don't own is common, and I'd rather the recommendation is obvious.

ChampionAsh5357: Ok, I copied the table over from the packaging section and made the text more explicit.

docs/gettingstarted/index.md
```diff
@@ -1,60 +1,119 @@
...
-This is a simple guide to get you from nothing to a basic mod. The rest of this documentation is about where to go from here.
+If you have never made a Forge mod before, this section will provide the minimum amount of information needed to setup a Forge development environment. The rest of the documentation is about where to go from here.
+
+Prerequisites
+-------------
+
+* An installation of the Java 17 Development Kit (JDK) and 64-bit Java Virtual Machine (JVM). Forge recommends and officially supports [Eclipse Temurin][jdk].
+
+!!! warning
+    Using a 32-bit JVM will cause some problems when using [ForgeGradle]. Make sure your computer supports a 64-bit architecture along with an associated JVM.
```
PaintNinja: Might be better to reword this to something along the lines of:
> Make sure you're using a 64-bit JVM. One way of checking is to run `java -version` in a terminal

Pretty much all computers used for MC dev these days have a 64bit architecture. JEP 449 proposes dropping 32bit Windows support in Java, Ubuntu and macOS dropped 32bit support years ago and Microsoft dropped it in Win11.

docs/gettingstarted/modfiles.md
Matyrobbrt: It seems like this property (which has a misleading name) is used for the [`uses`](https://docs.oracle.com/javase/specs/jls/se17/html/jls-7.html#jls-7.7.3) module declaration, allowing you to *load* a service.

docs/gettingstarted/index.md
```diff
@@ -1,60 +1,119 @@
...
-2. To test run your mod, the easiest way is to use the run configs that were generated when you set up your project. Otherwise, you can run `gradlew runClient`. This will launch Minecraft from the `<runDir>` location along with your mod's code in any source sets specified within your run configurations. The default MDK includes the `main` source set, so any code written within `src/main/java` will be applied.
-3. You can also run a dedicated server using the server run config or via `gradlew runServer`. This will launch the Minecraft server with its GUI. After the first run, the server will shut down immediately until the Minecraft EULA is accepted by editing `run/eula.txt`. Once accepted, the server will load and can be accessed via a direct connect to `localhost`.
+1. To build your mod, run `gradlew build`. This will output a file in `build/libs` with the name `[archivesBaseName]-[version].jar`. This file can be placed in the `mods` folder of a Forge-enabled Minecraft setup or distributed.
```
Matyrobbrt:
```suggestion
1. To build your mod, run `gradlew build`. This will output a file in `build/libs` with the name `[archivesBaseName]-[version].jar`, by default. This file can be placed in the `mods` folder of a Forge-enabled Minecraft setup or distributed.
```

docs/gettingstarted/index.md
```diff
@@ -1,60 +1,119 @@
...
+1. To run your mod in a test environment, you can either use the generated run configurations or use the associated tasks (e.g. `gradlew runClient`). This will launch Minecraft from the run directory (default 'run') along with any source sets specified. The default MDK includes the `main` source set, so any code written in `src/main/java` will be applied.
+1. If you are running a dedicated server, whether through the run configuration or `gradlew runServer`, the server will initially shut down immediately. You will need to accept the Minecraft EULA by editing the `eula.txt` file in the run directory. Once accepted, the server will load, which can then be accessed via a direct connect to `localhost`.
 
 !!! note
-    It is always advisable to test your mod in a dedicated server environment if it is intended to run there.
-    
+    You should always test your mod in a dedicated server environment. This includes [client-only mods][client] as they should not do anything when loaded on the server.
```
Matyrobbrt, line 118: This could maybe explain best practices for client-only mods (such as client-only EBS or early constructor exit with a different method). Could be in a different page or [here](https://github.com/MinecraftForge/Documentation/pull/490/files#diff-2099cf892b927b6c29c6c1190c139f16521d5b87764b74ad0bb6b1457060dc8dR54).

docs/gettingstarted/modfiles.md
```diff
@@ -0,0 +1,160 @@
...
+=========
+
+The mod files are responsible for determining what mods are packaged into your JAR, what information to display within the 'Mods' menu, and how your mod should be loaded in the game.
+
+mods.toml
+---------
+
+The `mods.toml` file defines the metadata of your mod(s). It also contains additional information that is displayed within the 'Mods' menu and how your mod(s) should be loaded into the game.
+
+The file uses the [Tom's Obvious Minimal Language, or TOML,][toml] format. The file must be stored under the `META-INF` folder in the resource directory of the source set you are using (`src/main/resources/META-INF/mods.toml` for the `main` source set). A `mods.toml` file may look something like this:
```
Matyrobbrt:
```suggestion
The file uses the [Tom's Obvious Minimal Language, or TOML][toml], format. The file must be stored under the `META-INF` folder in the resource directory of the source set you are using (`src/main/resources/META-INF/mods.toml` for the `main` source set). A `mods.toml` file may look something like this:
```

docs/gettingstarted/modfiles.md
```diff
@@ -0,0 +1,160 @@
...
+`modproperties` | table   | `{}`                    | A table of key/values associated with this mod. Currently unused by Forge, but is mainly for use by mods. | `{ example = "value" }` 
+`modUrl`        | string  | *nothing*               | A URL to the download page of the mod. Currently unused. | `"https://files.minecraftforge.net/"`
+`credits`       | string  | *nothing*               | Credits and acknowledges for the mod shown on the mod list screen. | `"The person over here and there."`
+`authors`       | string  | *nothing*               | The authors of the mod shown on the mod list screen. | `"Example Person"`
+`displayURL`    | string  | *nothing*               | A URL to the display page of the mod shown on the mod list screen. | `"https://minecraftforge.net/"`
+`displayTest`   | string  | `"MATCH_VERSION"`       | See '[sides]'. | `"NONE"`
+
+#### Features
+
+The features system allows mods to demand that certain settings, software, or hardware are available when loading the system. Currently, only the following features can be specified:
```
Matyrobbrt:
```suggestion
The features system allows mods to demand that certain settings, software, or hardware are available when loading the system. When a feature is not satisfied, mod loading will fail, informing the user about the requirement. Currently, Forge provides the following features:
```

docs/gettingstarted/modfiles.md
```diff
@@ -0,0 +1,160 @@
...
+
+Mods can specify their dependencies, which are checked by Forge before loading the mods. These configurations are created using the [array of tables][array] `[[dependencies.<modid>]]` where `modid` is the identifier of the mod the dependency is for.
+
+Property       | Type    | Default       | Description | Example
+:---           | :---:   | :---:         | :---:       | :---
+`modId`        | string  | **mandatory** | The identifier of the mod added as a dependency. | `"example_library"`
+`mandatory`    | boolean | **mandatory** | Whether the game should crash when this dependency is not met. | `true`
+`versionRange` | string  | `""`          | The acceptable version range of the language loader, expressed as a [Maven Version Range][mvr]. An empty string matches any version. | `"[1, 2)"`
+`ordering`     | string  | `"NONE"`      | Defines if the mod must load before (`"BEFORE"`) or after (`"AFTER"`) this dependency. If the ordering does not matter, return `"NONE"` | `"AFTER"`
+`side`         | string  | `"BOTH"`      | The [physical side][dist] the most must be present on: `"CLIENT"`, `"SERVER"`, or `"BOTH"`.| `"CLIENT"`
```
Matyrobbrt:
```suggestion
`side`         | string  | `"BOTH"`      | The [physical side][dist] the dependency must be present on: `"CLIENT"`, `"SERVER"`, or `"BOTH"`.| `"CLIENT"`
```

docs/gettingstarted/structuring.md
```diff
@@ -1,124 +1,81 @@
 Structuring Your Mod
 ====================
 
-Let us look at how to organize your mod into different files and what those files should do.
+Structured mods are beneficial for maintenance, making contributions, and providing a clearer understanding of the underlying codebase. Some of the recommendations from Java, Minecraft, and Forge are listed below.
+
+!!! note
+    You do not have to follow the advice below; you can structure your mod anyway you set fit. However, it is still highly recommended to do so.
```
Matyrobbrt:
```suggestion
    You do not have to follow the advice below; you can structure your mod any way you set fit. However, it is still highly recommended to do so.
```

docs/gettingstarted/structuring.md
```diff
@@ -1,124 +1,81 @@
...
-The Mod File
-------------
+Type      | Value             | Top-Level Package
+:---:     | :---:             | :---
+Domain    | example.com       | `com.example`
+Subdomain | example.github.io | `io.github.example`
+Email     | example@gmail.com | `com.gmail.example`
 
-Generally, we will start with a file named after your mod and put into your package. This is the *entry point* to your mod and will contain some special indicators marking it as such.
+The next level package should then be your mod's id (e.g. `com.example.examplemod` where `examplemod` is the mod id). This will guarantee that, unless you have two mods with the same mod id (which should never be the case), your packages should not have any issues loading.
```
Matyrobbrt:
```suggestion
The next level package should then be your mod's id (e.g. `com.example.examplemod` where `examplemod` is the mod id). This will guarantee that, unless you have two mods with the same id (which should never be the case), your packages should not have any issues loading.
```

docs/gettingstarted/structuring.md
```diff
@@ -1,124 +1,81 @@
...
 
-This is an annotation indicating to the Forge Mod Loader that the class is a Mod entry point. The `@Mod` annotation's value should match a mod id in the `src/main/resources/META-INF/mods.toml` file.
+### Sub-package Organization
 
-Keeping Your Code Clean Using Sub-packages
-------------------------------------------
+In addition to the top-level package, it is highly recommend to break your mod's classes between subpackages. There are numerous methods on how to do so:
 
-Rather than clutter up a single class and package with everything, it is recommended that you break your mod into subpackages.
+* **Group By Function**: Make subpackages for classes with a common purpose. For example, blocks can be under `block` or `blocks`, entities under `entity` or `entities`, etc. Minecraft uses this structure with the singular version of the word.
```
Matyrobbrt: Given it is names and we're talking about official mappings.
```suggestion
* **Group By Function**: Make subpackages for classes with a common purpose. For example, blocks can be under `block` or `blocks`, entities under `entity` or `entities`, etc. Mojang uses this structure with the singular version of the word.
```

docs/gettingstarted/modfiles.md
```diff
@@ -0,0 +1,160 @@
...
+    The `ordering` of two mods may cause a crash due to a cyclic dependency: for example, mod A must load `"BEFORE"` mod B and mod B `"BEFORE"` mod A.
+
+Mod Entrypoints
+---------------
+
+Now that the `mods.toml` is filled out, we need to provide an entrypoint to being programming the mod. Entrypoints are essentially the starting point for executing the mod. The entrypoint itself is determined by the language loader used in the `mods.toml`.
+
+### `javafml` and `@Mod`
+
+`javafml` is a language loader provided by Forge for the Java programming language. The entrypoint is defined using a public class with the `@Mod` annotation. The value of `@Mod` must contain one of the modids specified within the `mods.toml`. From there, all initialization logic (e.g., [registering events][events], [adding `DeferredRegister`s][registration]) can be specified within the constructor of the class. The mod bus can be obtained from `FMLJavaModLoadingContext`.
```
Matyrobbrt:
```suggestion
`javafml` is a language loader provided by Forge for the Java programming language. The entrypoint is defined using a public class with the `@Mod` annotation. The value of `@Mod` must contain one of the mod ids specified within the `mods.toml`. From there, all initialization logic (e.g., [registering events][events], [adding `DeferredRegister`s][registration]) can be specified within the constructor of the class. The mod bus can be obtained from `FMLJavaModLoadingContext`.
```

docs/gettingstarted/structuring.md
```diff
@@ -1,124 +1,81 @@
...
+* **Group By Logic**: Make subpackages for classes with a common logic. For example, if you were creating a new type of crafting table, you would put its block, menu, item, and more under `feature/crafting_table`.
```
Matyrobbrt:
```suggestion
* **Group By Logic**: Make subpackages for classes with a common logic. For example, if you were creating a new type of crafting table, you would put its block, menu, item, and more under `feature.crafting_table`.
```

docs/gettingstarted/structuring.md
```diff
@@ -1,124 +1,81 @@
...
+
+Classes are commonly suffixed with its type, for example:
 
-For Example:
+* An `Item` called `PowerRing` -> `PowerRingItem`.
+* A `Block` called `NotDirt` -> `NotDirtBlock`.
+* A menu for an `Oven` -> `OvenMenu`.
+
+!!! note
+    Minecraft typically follows a similar structure for all classes except entities. Those are represented by just their names (e.g. `Pig`, `Zombie`, etc.).
```
Matyrobbrt:
```suggestion
    Mojang typically follows a similar structure for all classes except entities. Those are represented by just their names (e.g. `Pig`, `Zombie`, etc.).
```

Matyrobbrt (changes requested):

docs/gettingstarted/index.md
ChampionAsh5357: Replaced all references with maven versioning and added a note on the versioning page that `ComparableVersion` is not fully compatible with semver 2.

docs/gettingstarted/index.md
ChampionAsh5357: I'm not sure how to best handle the organizational structure here since we have a 'sides' page which should be dedicated to these question and is linked to from this page. I think the best idea is to put that on the backburner until I rewrite sides with more examples, which I should be able to tackle in a separate PR relatively quickly.

docs/concepts/resources.md
```diff
@@ -14,7 +14,7 @@ All resources should have snake case paths and filenames (lowercase, using "_" f
...
+Minecraft identifies resources using `ResourceLocation`s. A `ResourceLocation` contains two parts: a namespace and a path. It generally points to the resource at `assets/<namespace>/<ctx>/<path>`, where `ctx` is a context-specific path fragment that depends on how the `ResourceLocation` is being used. When a `ResourceLocation` is written/read as from a string, it is seen as `<namespace>:<path>`. If the namespace and the colon are left out, then when the string is read into an `ResourceLocation` the namespace will always default to `"minecraft"`. A mod should put its resources into a namespace with the same name as its mod id (e.g. a mod with id `examplemod` should place its resources in `assets/examplemod` and `data/examplemod` respectively, and `ResourceLocation`s pointing to those files would look like `examplemod:<path>`.). This is not a requirement, and in some cases it can be desirable to use a different (or even more than one) namespace. `ResourceLocation`s are used outside the resource system, too, as they happen to be a great way to uniquely identify objects (e.g. [registries][]).
```
Matyrobbrt:
```suggestion
Minecraft identifies resources using `ResourceLocation`s. A `ResourceLocation` contains two parts: a namespace and a path. It generally points to the resource at `assets/<namespace>/<ctx>/<path>`, where `ctx` is a context-specific path fragment that depends on how the `ResourceLocation` is being used. When a `ResourceLocation` is written/read as from a string, it is seen as `<namespace>:<path>`. If the namespace and the colon are left out, then when the string is read into an `ResourceLocation` the namespace will always default to `"minecraft"`. A mod should put its resources into a namespace with the same name as its mod id (e.g. a mod with the id `examplemod` should place its resources in `assets/examplemod` and `data/examplemod` respectively, and `ResourceLocation`s pointing to those files would look like `examplemod:<path>`.). This is not a requirement, and in some cases it can be desirable to use a different (or even more than one) namespace. `ResourceLocation`s are used outside the resource system, too, as they happen to be a great way to uniquely identify objects (e.g. [registries][]).
```

docs/contributing.md
```diff
@@ -1,6 +1,8 @@
 Contributing to This Documentation
 ==================================
 
+You can make a contribution via a PR on [Github].
```
Matyrobbrt:
```suggestion
You can make a contribution via a PR on [GitHub].
```

docs/gettingstarted/index.md
```diff
@@ -1,60 +1,119 @@
...
-    * For IntelliJ, run the `genIntellijRuns` gradle task (`gradlew genIntellijRuns`). This will generate the Run Configurations and download any required assets for the game to run. If you encounter an error saying "module not specified", you can either edit the configuration to select your "main" module or specify it through the `ideaModule` property.
-    * For VSCode, run the `genVSCodeRuns` gradle task (`gradlew genVSCodeRuns`). This will generate the Launch Configurations and download any required assets for the game to run.
+
+    The `src` subdirectory does not need to be copied across workspaces; however, you may need to refresh the Gradle project if the java (`src/main/java`) and resource (`src/main/resources`) are created later.
+
+3. Open your selected IDE:
+    * Forge only explicitly supports development on Eclipse, but there are additional run configurations for IntelliJ IDEA and Visual Studio Code. Regardless, any environment, from Apache NetBeans to Vim / Emacs, can be used.
```
Matyrobbrt: Official support is not only provided for Eclipse, but for IntelliJ as well

docs/gettingstarted/index.md
```diff
@@ -1,60 +1,119 @@
...
+
+```gradle
+// In some build.gradle
+group = 'com.example'
+```
+
+The packages within your java source (`src/main/java`) should also now conform to this structure, with an inner package representing the mod id:
+
+```text
+com
```
Matyrobbrt, line 91:
```suggestion
com (TLD) 
```

docs/gettingstarted/versioning.md
```diff
@@ -1,7 +1,10 @@
 Versioning
 ==========
 
-In general projects, [Semantic Versioning][semver] is often used (which has the format `MAJOR.MINOR.PATCH`). However, in the case of modding it may be more beneficial to use the format `MCVERSION-MAJORMOD.MAJORAPI.MINOR.PATCH` to be able to differentiate between world-breaking and API-breaking changes of a mod.
+In general projects, [semantic versioning][semver] is often used (which has the format `MAJOR.MINOR.PATCH`). However, in the case of modding it may be more beneficial to use the format `MCVERSION-MAJORMOD.MAJORAPI.MINOR.PATCH` to be able to differentiate between world-breaking and API-breaking changes of a mod.
+
+!!! important
+    Forge uses `ComparableVersion` to compare version strings, which is not fully compatible with the Semantic Versioning 2.0.0 spec, such as the 'prerelease' tag.
```
Matyrobbrt:
```suggestion
    Forge uses Maven Version ranges to compare version strings, which is not fully compatible with the Semantic Versioning 2.0.0 spec, such as the 'prerelease' tag.
```

Matyrobbrt (changes requested):

docs/gettingstarted/index.md
ChampionAsh5357: I think this should go on the `example` section where it says `(top-level package (group))`

docs/gettingstarted/versioning.md
ChampionAsh5357: Why are we being less specific here? Forge uses `ComparableVersion` to compare version strings.

Matyrobbrt: We should specify the specification, not implementation

Matyrobbrt (approved):

docs/gettingstarted/index.md
```diff
@@ -1,60 +1,119 @@
...
-This is a simple guide to get you from nothing to a basic mod. The rest of this documentation is about where to go from here.
+If you have never made a Forge mod before, this section will provide the minimum amount of information needed to setup a Forge development environment. The rest of the documentation is about where to go from here.
+
+Prerequisites
+-------------
+
+* An installation of the Java 17 Development Kit (JDK) and 64-bit Java Virtual Machine (JVM). Forge recommends and officially supports [Eclipse Temurin][jdk].
+
+!!! warning
+    Make sure you are using a 64-bit JVM. One way of checking is to run `java -version` in a terminal. Using a 32-bit JVM will cause some problems when using [ForgeGradle].
```
sciwhiz12: Admonitions that are 'part' of a list item should be indented to follow.
```suggestion
    !!! warning
        Make sure you are using a 64-bit JVM. One way of checking is to run `java -version` in a terminal. Using a 32-bit JVM will cause some problems when using [ForgeGradle].
```

docs/gettingstarted/index.md
```diff
@@ -1,60 +1,119 @@
...
 From Zero to Modding
 --------------------
 
-1. Obtain a Java 17 Development Kit (JDK) and a 64-bit Java Virtual Machine (JVM). Minecraft and MinecraftForge both compile against Java 17 and as such should be used for development. Using a 32-bit JVM will result in some problems when running the below gradle tasks. You can obtain one from [Eclipse Adoptium][jdk].
-2. Obtain the Mod Development Kit (MDK) from Forge's [files][] site.
-3. Extract the downloaded MDK into an empty directory. You should see a bunch of files along with an example mod placed in `src/main/java` for you to look at. Only a few of these files are strictly necessary for mod development, and you may reuse these files for all your projects. These files are:
+1. Download the Mod Developer Kit (MDK) from the [Forge file site][files] by clicking 'Mdk' followed by the 'Skip' button in the top right after waiting for a period of time. It is recommended to download the latest version of Forge whenever possible.
+1. Extract the downloaded MDK into an empty directory. This will be your mod's directory, which should now contain some gradle files and a `src` subdirectory containing the example mod.
+
+!!! note
```
sciwhiz12: Same comment as above with regards to indenting adomitions that form a 'part' of a list item. (I can't leave a suggestion for the whole admonition because of the removals in the way.)

docs/gettingstarted/index.md
```diff
@@ -1,60 +1,119 @@
...
-6. Generating IDE Launch/Run Configurations:
-    * For Eclipse, run the `genEclipseRuns` gradle task (`gradlew genEclipseRuns`). This will generate the Launch Configurations and download any required assets for the game to run. After this has finished, refresh your project.
-    * For IntelliJ, run the `genIntellijRuns` gradle task (`gradlew genIntellijRuns`). This will generate the Run Configurations and download any required assets for the game to run. If you encounter an error saying "module not specified", you can either edit the configuration to select your "main" module or specify it through the `ideaModule` property.
-    * For VSCode, run the `genVSCodeRuns` gradle task (`gradlew genVSCodeRuns`). This will generate the Launch Configurations and download any required assets for the game to run.
+
+    The `src` subdirectory does not need to be copied across workspaces; however, you may need to refresh the Gradle project if the java (`src/main/java`) and resource (`src/main/resources`) are created later.
+
+3. Open your selected IDE:
```
sciwhiz12: Numbering: either it's all `1.`, or it's numbered sequentially.
```suggestion
1. Open your selected IDE:
```

docs/gettingstarted/index.md
```diff
@@ -1,60 +1,119 @@
...
+    * Forge only explicitly supports development on Eclipse and IntelliJ IDEA, but there are additional run configurations for Visual Studio Code. Regardless, any environment, from Apache NetBeans to Vim / Emacs, can be used.
+    * Eclipse and IntelliJ IDEA's Gradle integration, both installed and enabled by default, will handle the rest of the initial workspace setup on import or open. This includes downloading the necessary packages from Mojang, MinecraftForge, etc. The 'Gradle for Java' plugin is needed for Visual Studio Code to do the same.
+    * Gradle will need to be invoked to re-evaluate the project for almost all changes to its associated files (e.g., `build.gradle`, `settings.gradle`, etc.). Some IDEs come with 'Refresh' buttons to do this; however, it can be done through the terminal via `gradlew build --refresh-dependencies`.
+3. Run configurations can be generated for Eclipse, IntelliJ IDEA, or Visual Studio Code:
```
sciwhiz12: Numbering: either it's all `1.`, or it's numbered sequentially.

Additionally, to be in parallel with the structure of the other list items, this should be imperative/have active voice.

```suggestion
1. Generate run configurations for your selected IDE:
```

docs/gettingstarted/index.md
```diff
@@ -1,60 +1,119 @@
...
+
+    The `src` subdirectory does not need to be copied across workspaces; however, you may need to refresh the Gradle project if the java (`src/main/java`) and resource (`src/main/resources`) are created later.
+
+3. Open your selected IDE:
+    * Forge only explicitly supports development on Eclipse and IntelliJ IDEA, but there are additional run configurations for Visual Studio Code. Regardless, any environment, from Apache NetBeans to Vim / Emacs, can be used.
+    * Eclipse and IntelliJ IDEA's Gradle integration, both installed and enabled by default, will handle the rest of the initial workspace setup on import or open. This includes downloading the necessary packages from Mojang, MinecraftForge, etc. The 'Gradle for Java' plugin is needed for Visual Studio Code to do the same.
+    * Gradle will need to be invoked to re-evaluate the project for almost all changes to its associated files (e.g., `build.gradle`, `settings.gradle`, etc.). Some IDEs come with 'Refresh' buttons to do this; however, it can be done through the terminal via `gradlew build --refresh-dependencies`.
```
sciwhiz12: I believe that the 'Refresh' IDE buttons do the equivalent of a simple `gradlew` (and the IDE uses the Gradle Tooling API to import the project models afterwards). Also, it may be unwise to suggest `--refresh-dependencies` for each refresh, as that means refetching all dependencies from remote repositories and may cause substantial delay to the user.

docs/gettingstarted/index.md
```diff
@@ -1,60 +1,119 @@
...
-These customizations are highly recommended for all projects.
+Replace all occurrences of `examplemod`, including [`mods.toml` and the main mod file][modfiles] with the mod id of your mod. This also includes changing the name of the file you build by setting `archivesBaseName` (this is typically set to your mod id).
 
-* To change the name of the file you build - edit the value of `archivesBaseName` to suit.
-* To change your "maven coordinates" - edit the value of `group` as well.
-* To change the version number - edit the value of `version`.
-* To update the run configurations - replace all occurrences of `examplemod` to the mod id of your mod.
+```gradle
+// In some build.gradle
+archivesBaseName = 'mymod'
```
sciwhiz12: As a note, I believe Gradle is looking to move away from conventions (which is what allows us to define `archivesBaseName` in the very root of the buildscript, per [`BasePluginConvention`](https://docs.gradle.org/current/dsl/org.gradle.api.plugins.BasePluginConvention.html)) to extensions. 

Specifically, we will likely need to transition this to `base { archivesName = 'mymod' }`. CCing @MinecraftForge/library-maintainers, for future reference.

docs/gettingstarted/modfiles.md
```diff
@@ -0,0 +1,165 @@
...
+`displayName`   | string  | value of `modId`        | The pretty name of the mod. Used when representing the mod on a screen (e.g., mod list, mod mismatch). | `"Example Mod"`
+`description`   | string  | `"MISSING DESCRIPTION"` | The description of the mod shown in the mod list screen. It is recommended to use a [multiline literal string][multiline]. | `"This is an example."`
+`logoFile`      | string  | *nothing*               | The name and extension of an image file used on the mods list screen. The logo must be in the root of the JAR or directly in the root of the source set (e.g., `src/main/resources` for the main source set). | `"example_logo.png"`
+`logoBlur`      | boolean | `true`                  | Whether to use `GL_LINEAR*` (true) or `GL_NEAREST*` (false) to render the `logoFile`. | `false`
+`updateJSONURL` | string  | *nothing*               | A URL to a JSON used by the [update checker][update] to make sure the mod you are playing is the current version. | `"https://files.minecraftforge.net/net/minecraftforge/forge/promotions_slim.json"`
```
sciwhiz12: "current version" ought to be "latest version" instead.
```suggestion
`updateJSONURL` | string  | *nothing*               | A URL to a JSON used by the [update checker][update] to make sure the mod you are playing is the latest version. | `"https://files.minecraftforge.net/net/minecraftforge/forge/promotions_slim.json"`
```

docs/gettingstarted/modfiles.md
```diff
@@ -0,0 +1,165 @@
...
+# Properties for examplemod2
+[[mods]]
+modId = "examplemod2"
+```
+
+Property        | Type    | Default                 | Description | Example
+:---            | :---:   | :---:                   | :---:       | :---
+`modId`         | string  | **mandatory**           | The unique identifier representing this mod. The id must match `^[a-z][a-z0-9_]{1,63}$` (a string 2-64 characters; starts with a lowercase letter; made up of lowercase letters, numbers, or underscores). | `"examplemod"`
+`namespace`     | string  | value of `modId`        | An override namespace for the mod. The namespace much match `^[a-z][a-z0-9_.-]{1,63}$` (a string 2-64 characters; starts with a lowercase letter; made up of lowercase letters, numbers, underscores, dots, or dashes). Currently unused. | `"example"`
+`version`       | string  | `"1"`                   | The version of the mod, preferably in a [variation of maven versioning][mvnver]. When set to `${file.jarVersion}`, it will be replaced with the value of the `Implementation-Version` property in the JAR's manifest. | `"1.19.4-1.0.0.0"`
```
sciwhiz12: Should we mention the special behavior of `0.0NONE` which IIRC is when the `Implementation-Version` isn't filled up due to there not being a JAR manifest available (as in dev time)?

docs/gettingstarted/structuring.md
```diff
@@ -1,124 +1,81 @@
...
+
+```
+a.jar
+  - com.example.ExampleClass
+b.jar
+  - com.example.ExampleClass // This class will not normally be loaded
 ```
 
-If any string is specified as `${file.jarVersion}`, Forge will replace the string with the **Implementation Version** specified in your jar manifest at runtime. Since the user development environment has no jar manifest to pull from, it will be `NONE` instead. As such, it is usually recommended to leave the `version` field alone. Here is a table of attributes that may be given to a mod, where `mandatory` means there is no default and the absence of the property causes an error.
+This is even more relevant when it comes to loading modules. If there are class files in two packages under the same name in separate modules, this will cause the mod loader to crash on startup since mod modules are exported to the game and other mods.
```
sciwhiz12: Perhaps emphasis should be added on the "to crash on startup" part of the paragraph.
```suggestion
This is even more relevant when it comes to loading modules. If there are class files in two packages under the same name in separate modules, this will cause the mod loader **to crash on startup** since mod modules are exported to the game and other mods.
```

docs/gettingstarted/structuring.md
```diff
@@ -1,124 +1,81 @@
...
-|  description  |  string  | mandatory         | A description of this mod. |
-| displayTest   | string   | `"MATCH_VERSION"` | Controls the display of the mod in the server connection screen.
-| dependencies  | [list]   |   `[]`            | A list of dependencies of this mod. |
+```
+module A
+  - package X
+    - class I
+    - class J
+module B
+  - package X // This package will not normally be loaded, preventing usage of R, S, T
```
sciwhiz12: As a bit of nuance: By my understanding, Java doesn't crash by the mere presence of two modules have the same package. The part that causes the crash is when two modules co-exist and try to export that same package out to another module (or all other modules, in the case of `open` modules).

As Java provides APIs to build module layers programmatically, an application or framework can load multiple modules that have the same package as long as each module does not co-exist with another on the same layer and export the same package.

docs/gettingstarted/structuring.md
```diff
@@ -1,124 +1,81 @@
...
-What is `@Mod`?
--------------
+You can find some additional naming conventions on [Oracle's tutorial page][naming].
 
-This is an annotation indicating to the Forge Mod Loader that the class is a Mod entry point. The `@Mod` annotation's value should match a mod id in the `src/main/resources/META-INF/mods.toml` file.
+### Sub-package Organization
 
-Keeping Your Code Clean Using Sub-packages
-------------------------------------------
+In addition to the top-level package, it is highly recommend to break your mod's classes between subpackages. There are numerous methods on how to do so:
```
sciwhiz12: Re-word the last sentence, since we only provide two methods here (and usually these are the major two methods that are used):
```suggestion
In addition to the top-level package, it is highly recommend to break your mod's classes between subpackages. There are two major methods on how to do so:
```

docs/gettingstarted/structuring.md
```diff
@@ -1,124 +1,81 @@
...
+    Mojang typically follows a similar structure for all classes except entities. Those are represented by just their names (e.g. `Pig`, `Zombie`, etc.).
 
-* An `Item` called `PowerRing` would be in an `item` package, with a class name of `PowerRingItem`.
-* A `Block` called `NotDirt` would be in a `block` package, with a class name of `NotDirtBlock`.
-* Finally, a `BlockEntity` for a block called `SuperChewer` would be a `block.entity` or `blockentity` package, with a class name of `SuperChewerBlockEntity`.
+Choose One Method from Many
+---------------------------
 
-Appending your class names with what *kind* of object they are makes it easier to figure out what a class is or guess the class for an object.
+There are many methods for performing a certain task: registering an object, listening for events, etc. It's generally recommended to only use a single method to accomplish a given task. While this does improve code formatting, it also avoid any weird interactions or redundancies that may occur (e.g. your event listener executing twice).
```
sciwhiz12: Perhaps this should mention "consistency" as well; stick to one way to do it and be consistent throughout your code.

docs/gettingstarted/versioning.md
```diff
@@ -1,7 +1,10 @@
 Versioning
 ==========
 
-In general projects, [Semantic Versioning][semver] is often used (which has the format `MAJOR.MINOR.PATCH`). However, in the case of modding it may be more beneficial to use the format `MCVERSION-MAJORMOD.MAJORAPI.MINOR.PATCH` to be able to differentiate between world-breaking and API-breaking changes of a mod.
+In general projects, [semantic versioning][semver] is often used (which has the format `MAJOR.MINOR.PATCH`). However, in the case of modding it may be more beneficial to use the format `MCVERSION-MAJORMOD.MAJORAPI.MINOR.PATCH` to be able to differentiate between world-breaking and API-breaking changes of a mod.
+
+!!! important
+    Forge uses [Maven version ranges][cmpver] to compare version strings, which is not fully compatible with the Semantic Versioning 2.0.0 spec, such as the 'prerelease' tag.
```
sciwhiz12, line 7: Looking at the diffs in other files, they seem to use `maven` (lowercase) instead of `Maven` (capitalized). One or the other should be picked and then all uses be made consistent.

sciwhiz12 (changes requested):

Icicles0:
I think this should just get merged already

docs/gettingstarted/structuring.md
ChampionAsh5357: I'm not sure whether that should be stated, but I fixed the message to reflect the text above mentioning the crash.

docs/gettingstarted/structuring.md
ChampionAsh5357: It would help if I remembered that consistency was a word when I originally wrote this.

ChampionAsh5357:
> I think this should just get merged already

It only makes sense to merge something once we consider it to be accurate. Especially since we are rewriting this to be more explanatory and user friendly to new modders.

docs/gettingstarted/index.md
```diff
@@ -53,13 +53,23 @@ Edit the `build.gradle` file to customize how your mod is built (e.g., file name
...
 
+!!! note
+    The Forge MDK currently uses `archivesBaseName` to set the artifact name instead of `base.archivesName`. We recommend using `base.archivesName` instead as `archivesBaseName` is deprecated for removal in Gradle 9, which a future version of ForgeGradle will support.
+
+    You can still use `archivesBaseName` by setting the following:
+
+    ```gradle
+    // In some build.gradle
+    base.archivesName = 'mymod'
+    ```
```
sciwhiz12: I was more thinking on doing this change when we change the MDK as well, but it works too. :shipit:

sciwhiz12 (approved):

PR #493, opened 2023-04-02
[1.19.x] Document Forge extensions of the model JSON format

XFactHD:
Documents the various extensions Forge adds to the model JSON format.
Opening this as a draft as I intend to also add docs for the other model loaders provided by Forge (only the OBJ loader is documented at the moment).

@amadornes I would especially appreciate your feedback on this since your client reworks affected a bunch of the extensions documented here.


docs/rendering/modelloaders/index.md
```diff
@@ -3,6 +3,8 @@ Custom Model Loaders
 
 A "model" is simply a shape. It can be a simple cube, it can be several cubes, it can be a truncated icosidodecahedron, or anything in between. Most models you'll see will be in the vanilla JSON format. Models in other formats are loaded into `IUnbakedGeometry`s by an `IGeometryLoader` at runtime. Forge provides default implementations for WaveFront OBJ files, buckets, composite models, models in different render layers, and a reimplementation of Vanilla's `builtin/generated` item model. Most things do not care about what loaded the model or what format it's in as they are all eventually represented by an `BakedModel` in code.
 
+Specifying a custom model loader through the top-level `loader` entry in a model JSON will cause the `elements` entry to be ignored if specified.
```
ChampionAsh5357: This should probably be wrapped in a `!!! note` or `!!! warning` tag. Although, I will mention that it depends on if the model loader consumes the `elements` entry since the same can be said with any of the base tags for the vanilla model.

docs/rendering/modelextensions/index.md
```diff
@@ -0,0 +1,16 @@
...
+- [Root Transforms][transforms]
+- [Render Types][rendertypes]
+- [Part Visibility][visibility]
+- [Face Data][facedata]
+
+[modelloaders]: ../modelloaders/index.md
+[transforms]: ./transforms.md
+[rendertypes]: ./rendertypes.md
+[visibility]: ./visibility.md
+[facedata]: ./facedata.md
```
ChampionAsh5357: This introduction section doesn't feel as though it is introducing anything. This should either hold the docs for something, like vanilla extensions, or just shouldn't exist at all since it can be seen through the sidebar.

docs/rendering/modelextensions/transforms.md
```diff
@@ -0,0 +1,32 @@
+Root Transforms
+===============
+
+Adding the `transform` entry at the top level of a model JSON allows specifying a transformation that will be baked into the vertex data of the quads. These transforms apply before the transforms specified in the [blockstate][blockstate] file of a block model and the [display transforms][displaytransform] of an item model.
```
ChampionAsh5357: You can simplify `[blockstate][blockstate]` to just `[blockstate]`.

Also, you should talk about transformations as to what they are applied to. For example, the line is ambiguous for whether the transformation is applied to the model as a whole or onto an individual face.

docs/rendering/modelextensions/transforms.md
```diff
@@ -0,0 +1,32 @@
...
+
+The root transforms can be specified in four formats:
+
+- A string referring to a default value (currently only supports the value `"identity"`)
+- A raw transformation matrix in the form of a nested JSON array with the last row ommited (3*4 matrix, row major order)
+- A JSON object containing a singular `matrix` entry with the same format as above
+- A JSON object containing any combination of the following optional entries:
```
ChampionAsh5357: Provide examples with comments. You could infer what this means, but always assume that they don't have the relevant knowledge or context.

docs/rendering/modelextensions/transforms.md
```diff
@@ -0,0 +1,32 @@
...
+    - `origin`: origin point used for the rotations and scaling, defaults to `opposing-corner`
+        - Array with 3 floating point values for x, y, z
+        - One of the default values: `"corner"` (0, 0, 0), `"center"` (.5, .5, .5) or `"opposing-corner"` (1, 1, 1)
```
ChampionAsh5357: Unclear how the interplay between the two options work. From what I gather, these can be combined together, but it seems as though that it should be one or the other, not both.

docs/rendering/modelextensions/transforms.md
```diff
@@ -0,0 +1,32 @@
...
+        - Array with 3 floating point values for x, y, z
+    - `rotation` or `left_rotation`: rotation around the translated origin to be applied before scaling, defaults to no rotation
+        - Single JSON object with a single axis -> rotation degree mapping
+        - Array of an arbitrary amount of JSON objects with the above format
+        - Array of 3 floating point values specifying the rotation in degrees around each axis
+        - Array of 4 floating point values specifying a quaternion directly
+    - `scale`: scale relative to the translated origin, defaults to (1, 1, 1)
+        - Single floating point value
+        - Array with 3 floating point values for x, y, z
+    - `right_rotation` or `post_rotation`: rotation around the translated origin to be applied after scaling, defaults to no rotation
```
ChampionAsh5357: Are the transformations performed in the order provided?

docs/rendering/modelextensions/rendertypes.md
```diff
@@ -0,0 +1,31 @@
+Render Types
+============
+
+Adding the `render_type` entry at the top level allows specifying the render type this model will render in. If this is not specified, the model will fall back to the render types returned by `ItemBlockRenderTypes#getRenderLayers()`.
```
ChampionAsh5357: Provide an example.

docs/rendering/modelextensions/rendertypes.md
```diff
@@ -0,0 +1,31 @@
...
+- `minecraft:cutout_mipped_all`
+    - Chunk render type: `RenderType#cutoutMipped()`
+    - Entity render type: `ForgeRenderTypes#ITEM_LAYERED_CUTOUT_MIPPED`
+- `minecraft:translucent`
+    - Chunk render type: `RenderType#translucent()`
+    - Entity render type: `ForgeRenderTypes#ITEM_LAYERED_TRANSLUCENT`
+- `minecraft:tripwire`
+    - Chunk render type: `RenderType#tripwire()`
+    - Entity render type: `ForgeRenderTypes#ITEM_LAYERED_TRANSLUCENT`
+    - Chunk and entity render type differ due to the tripwire render type not being feasible as an entity render type
```
ChampionAsh5357: Provide examples on when you want to use these render types (e.g. when should I use `cutout_mipped` vs `cutout_mipped_all`)?

docs/rendering/modelextensions/rendertypes.md
```diff
@@ -0,0 +1,31 @@
+Render Types
+============
+
+Adding the `render_type` entry at the top level allows specifying the render type this model will render in. If this is not specified, the model will fall back to the render types returned by `ItemBlockRenderTypes#getRenderLayers()`.
+
+This replaces the deprecated method of setting the applicable render type(s) via `ItemBlockRenderTypes#setRenderLayer()` for blocks.
```
ChampionAsh5357: Probably better as a `!!! note` and provide what version did this. Also mention that this is **only** for blocks and not fluids.

docs/rendering/modelextensions/rendertypes.md
```diff
@@ -0,0 +1,31 @@
...
+    - Entity render type: `ForgeRenderTypes#ITEM_LAYERED_CUTOUT_MIPPED`
+- `minecraft:translucent`
+    - Chunk render type: `RenderType#translucent()`
+    - Entity render type: `ForgeRenderTypes#ITEM_LAYERED_TRANSLUCENT`
+- `minecraft:tripwire`
+    - Chunk render type: `RenderType#tripwire()`
+    - Entity render type: `ForgeRenderTypes#ITEM_LAYERED_TRANSLUCENT`
+    - Chunk and entity render type differ due to the tripwire render type not being feasible as an entity render type
+
+Custom named render types to be specified in a model can be registered in the `RegisterNamedRenderTypesEvent`.
```
ChampionAsh5357: Mentioned with no explanation or example on what this means and how to use the event on which bus.

docs/rendering/modelextensions/visibility.md
```diff
@@ -0,0 +1,12 @@
+Part Visibility
+===============
+
+Adding the `visibility` entry at the top level of a model JSON allows control over the visibility of different parts of the model to decide whether they should be baked into the final [`BakedModel`][bakedmodel]. The definition of a "part" is dependent on the model loader loading this model. Out of the model loaders provided by Forge only the [composite model loader][composite] and the [OBJ model loader][obj] make use of this functionality.
```
ChampionAsh5357: Provide an example.

docs/rendering/modelextensions/visibility.md
```diff
@@ -0,0 +1,12 @@
...
+
+The visibility entries are specified as `"part name": boolean` entries. The check for the visibility of a given part traverses the full dependency chain until it finds an entry or reaches the end of the chain such that if a model doesn't specify the visibility of a part its parent will be interrogated for the visibility of that part. If no model in the dependency chain specifies a visibility for a part the visibility will default to true.
```
ChampionAsh5357: Is this just saying it finds the visibility of the part by checking whether its specified and recursively checks the parent until either it finds an instance or is not specified, so it defaults to true? If so, I would say the language is not modder-friendly and probably should be simplified or given an example.

docs/rendering/modelextensions/facedata.md
```diff
@@ -0,0 +1,74 @@
+Face Data
+=========
+
+Additional data about an element's faces can be specified at either the element level or the face level in a vanilla "elements" model and in generated item models loaded through the `forge:item_layers` model loader. Faces which do not specify their own face data will fall back to the element's face data or a default if no face data is specified at the element level. All values of the face data are optional.
```
ChampionAsh5357: This is very misleading since it seems to talk about an extension to vanilla which then tells you to use Forge's model loader.

docs/rendering/modelextensions/facedata.md
```diff
@@ -0,0 +1,74 @@
...
+            "color": "0xFFFF0000",
+            "block_light": 15,
+            "sky_light": 15,
+            "ambient_occlusion": false
+          },
+          ...
+        },
+        ...
+      },
+      ...
```
ChampionAsh5357: Use comments to explain things.

Also `// ...` to make valid JavaScript.

docs/rendering/modelextensions/facedata.md
```diff
@@ -0,0 +1,74 @@
...
+      "color": "0xFFFF0000",
+      "block_light": 15,
+      "sky_light": 15,
+      "ambient_occlusion": false
+    },
+    ...
+  },
+  ...
+}
+```
```
ChampionAsh5357: I do not understand how this differs from the above example as there is no relation to the `elements` tag and no comments are provided.

docs/rendering/modelextensions/facedata.md
```diff
@@ -0,0 +1,74 @@
...
+
+In vanilla "elements" models the face data applies to the face it is specified in or all faces of the element it is specified in which don't have their own face data.  
+In generated item models loaded through the `forge:item_layers` loader the face data applies to all faces generated from the texture specified for the layer the respective entry is associated with.
```
ChampionAsh5357: Shouldn't this go before the examples since it explains what they mean?

docs/rendering/modelextensions/facedata.md
```diff
@@ -0,0 +1,74 @@
...
+
+Color
+-----
+
+Specifying a color value with the `color` entry will apply that color as a tint to the quads. Defaults to `0xFFFFFFFF` (white, fully opaque). The color must be in the `ARGB` format packed into a 32bit integer and can be specified as either a hexadecimal string (`"0xAARRGGBB"`) or as a decimal integer literal (JSON does not support hexadecimal integer literals).
```
ChampionAsh5357: 32bit -> 32-bit.

Also should explain that only specifying a 24-bit integer will make the color fully transparent since that seems like a common issue.

docs/rendering/modelextensions/facedata.md
```diff
@@ -0,0 +1,74 @@
...
+
+This can be used as a replacement for tinting with [`BlockColor` and `ItemColor`][tinting] if the color values are constant.
+
+Block and Sky Light
+-------------------
+
+Specifying a block and/or sky light value with the `block_light` and `sky_light` entry respectively will override the respective light value of the quads. Both values default to 0. The values must be in the range 0-15 (inclusive) and are treated as a minimum value for the respective light type when the face is rendered meaning that a higher in-world value of the respective light type will override the specified value.
```
ChampionAsh5357: This should probably be more specific whether the light affects the server or is purely client side. I'm thinking of how this would play with the `lightLevel` block property.

docs/rendering/modelextensions/facedata.md
```diff
@@ -0,0 +1,74 @@
...
+
+Ambient Occlusion
+-----------------
+
+Specifying the ambient occlusion flag will configure AO for the quads. Defaults to `true`. The behaviour of this flag is equivalent to the top-level `ambientocclusion` flag of the vanilla format.
```
ChampionAsh5357: Link to AO wiki page would probably be helpful. Or maybe an example with an image.

docs/rendering/modelextensions/facedata.md
```diff
@@ -0,0 +1,74 @@
...
+
+!!! note
+    If the top-level AO flag is set to false, specifying this flag as true on an element or face won't be able to override the top-level flag
```
ChampionAsh5357: Give an example.

ChampionAsh5357 (changes requested):
Mainly just need examples and more specific information since you tend to make the assumption that people already know about these systems by default. The docs don't need to be exhaustive, but they do need to be friendly enough that a user could come in and get a general idea of the information by reading it.

docs/rendering/modelloaders/index.md
XFactHD: The reason I mentioned the `elements` entry specifically is that Forge's extended model deserializer explicitely nukes the loaded elements when a custom loader is present. All other vanilla attributes are still loaded and available in the resulting `BlockModel` instance and may be used by vanilla and mod code. I have appended the note as such.

docs/rendering/modelextensions/index.md
XFactHD: I'm not satisfied with the introduction section because, yes, it's basically empty. But for some reason it felt weird to not have an introduction with a short overview. I'll have to find a way to meaningfully expand on this.

docs/rendering/modelextensions/transforms.md
XFactHD: Made it a bit clearer that it's an either-or

docs/rendering/modelextensions/transforms.md
XFactHD: Made it clearer that it applies to the full model

docs/rendering/modelextensions/transforms.md
XFactHD: I've separated the elementwise specification into a separate section to provide some examples

docs/rendering/modelextensions/transforms.md
XFactHD: Added an explicit comment about the order of operation

docs/rendering/modelextensions/rendertypes.md
XFactHD: Added a very simple example

docs/rendering/modelextensions/rendertypes.md
XFactHD: Added typical use cases, though I'll want to have amadornes' opinion on these, especially the tripwire one as that is a really weird one I'm having a hard time getting the point of.

docs/rendering/modelextensions/rendertypes.md
XFactHD: Converted to a note, not sure how I can make it more specific to blocks since the sentence ends as such.

docs/rendering/modelextensions/rendertypes.md
XFactHD: Expanded on custom named render types to give a rough idea what they are about.

docs/rendering/modelextensions/visibility.md
XFactHD: Added a simple example

docs/rendering/modelextensions/visibility.md
XFactHD: Simplified the explanation, I hope it makes more sense this way

docs/rendering/modelextensions/facedata.md
XFactHD: Separated the mention of the Forge loader into a separate paragraph and elaborated on why that loader is needed

docs/rendering/modelextensions/facedata.md
XFactHD: Expanded the example to make it more obvious

docs/rendering/modelextensions/facedata.md
XFactHD: Note added

docs/rendering/modelextensions/facedata.md
XFactHD: Moved to more fitting places

docs/rendering/modelextensions/facedata.md
XFactHD: Added a note about the non-existent effect on the actual light level and surrounding blocks

docs/rendering/modelextensions/facedata.md
XFactHD: Added a link to the AO wiki page and an example image comparing AO on vs. off

docs/rendering/modelextensions/facedata.md
XFactHD: Added an example

docs/rendering/modelextensions/facedata.md
XFactHD: I'm not sure what kind of comments to place within the example since the parameters are explained below

docs/rendering/modelextensions/facedata.md
```diff
@@ -0,0 +1,109 @@
...
+  ]
+}
+```
+
+Generated Item Model
+--------------------
+
+In generated item models loaded through the `forge:item_layers` loader the face data applies to all faces generated from the texture specified for the layer the respective entry is associated with.
+
+The additional data has to be specified at the top level with the entry names in the `forge_data` object being the layer index the respective entry belongs to (in the example layer 1 will be tinted red and glow at full brightness):
```
Matyrobbrt:
```suggestion
The additional data has to be specified at the top level with the entry names in the `forge_data` object being the layer index the respective entry belongs to (in the example, layer 1 will be tinted red and glow at full brightness):
```

docs/rendering/modelextensions/transforms.md
```diff
@@ -0,0 +1,67 @@
+Root Transforms
+===============
+
+Adding the `transform` entry at the top level of a model JSON allows specifying a transformation that will be baked into the vertex data of all quads of the model. These transforms apply before the transforms specified in the [blockstate] file of a block model and the [display transforms][displaytransform] of an item model.
+
+The root transforms can be specified in four formats:
+
+1. A string referring to a default value (currently only supports the value `"identity"`)
+2. A raw transformation matrix in the form of a nested JSON array with the last row ommited (3*4 matrix, row major order)
```
Matyrobbrt:
```suggestion
2. A raw transformation matrix in the form of a nested JSON array with the last row omitted (3*4 matrix, row major order)
```

docs/rendering/modelextensions/transforms.md
```diff
@@ -0,0 +1,67 @@
...
+3. A JSON object containing a singular `matrix` entry with the same format as above
+4. A JSON object containing any combination of the following optional entries:
+    - `origin`: origin point used for the rotations and scaling
+    - `translation`: relative translation
+    - `rotation` or `left_rotation`: rotation around the translated origin to be applied before scaling
+    - `scale`: scale relative to the translated origin
+    - `right_rotation` or `post_rotation`: rotation around the translated origin to be applied after scaling
+
+Elementwise specification
```
Matyrobbrt:
```suggestion
Element wise specification
```

docs/rendering/modelextensions/transforms.md
```diff
@@ -0,0 +1,67 @@
...
+    },
+    // ...
+}
+```
+
+The elements are expected to be defined as follows:
+
+### Origin
+
+The origin can be specified either as an array of 3 floating point values representing a three dimensional vector: `[ x, y, z ]` or as one of the three default values:
```
Matyrobbrt:
```suggestion
The origin can be specified either as an array of 3 floating point values representing a three-dimensional vector: `[ x, y, z ]` or as one of the three default values:
```

docs/rendering/modelextensions/transforms.md
```diff
@@ -0,0 +1,67 @@
...
+
+### Translation
+
+The translation must be specified as an array of 3 floating point values representing a three dimensional vector: `[ x, y, z ]` and defaults to (0, 0, 0) if not present.
+
+### Left and Right Rotation
+
+The rotations can be specified in any one of the following four ways:
+
+- Single JSON object with a single axis -> rotation degree mapping: `{ "X": 90 }`
```
Matyrobbrt: This could use an actual arrow, such as `⇒`

docs/rendering/modelextensions/transforms.md
```diff
@@ -0,0 +1,67 @@
...
+- Array of an arbitrary amount of JSON objects with the above format (applied in the order they are specified in): `[ { "x": 90 }, { "y": 45 }, { "x": -22.5 } ]`
+- Array of 3 floating point values specifying the rotation in degrees around each axis: `[ 90, 180, 45 ]`
+- Array of 4 floating point values specifying a quaternion directly: `[ 0.38268346, 0, 0, 0.9238795 ]` (example equals 45 degrees around the X axis)
+
+If the respective rotation is not specified, it will default to no rotation.
+
+### Scale
+
+The scale must be specified as an array of 3 floating point values representing a three dimensional vector: `[ x, y, z ]` and defaults to (1, 1, 1) if not present.
```
Matyrobbrt:
```suggestion
The scale must be specified as an array of 3 floating point values representing a three-dimensional vector: `[ x, y, z ]` and defaults to (1, 1, 1) if not present.
```

docs/rendering/modelextensions/visibility.md
```diff
@@ -0,0 +1,30 @@
...
+  },
+  "visibility": {
+    "part_two": false
+  }
+}
+```
+
+The visibility of a given part is determined by checking whether the model specifies a visibility for this part and, if not present, recursively checking the model's parent until either an entry is found or there is no further parent to check, in which case it defaults to true.
+
+This allows things like a composite model with multiple parts which multiple models use as their parent and then specify different visibility settings for the parts.
```
Matyrobbrt:
```suggestion
This allows things like a composite model with multiple parts which multiple models use as their parent, and then specify different visibility settings for the parts.
```

docs/rendering/modelextensions/rendertypes.md
```diff
@@ -0,0 +1,74 @@
...
+    - Entity render type: `ForgeRenderTypes#ITEM_LAYERED_TRANSLUCENT`
+    - Chunk and entity render type differ due to the tripwire render type not being feasible as an entity render type
+    - Used for blocks with the special requirement of being rendered to the weather render target (i.e. Tripwire)
+
+Custom Values
+-------------
+
+Custom named render types to be specified in a model can be registered in the `RegisterNamedRenderTypesEvent`. This event is fired on the mod event bus.
+
+A custom named render type consists two or three components:
```
Matyrobbrt:
```suggestion
A custom named render type consists of two or three components:
```

docs/rendering/modelextensions/facedata.md
```diff
@@ -0,0 +1,109 @@
...
+In a vanilla "elements" model, additional data about an element's faces can be specified at either the element level or the face level. Faces which do not specify their own face data will fall back to the element's face data or a default if no face data is specified at the element level.
+
+To use this extension for a generated item model, the model must be loaded through the `forge:item_layers` model loader due to the vanilla item model generator not being extended to read this additional data.
+
+All values of the face data are optional.
+
+Elements Model
+--------------
+
+In vanilla "elements" models the face data applies to the face it is specified in or all faces of the element it is specified in which don't have their own face data.
```
Matyrobbrt:
```suggestion
In vanilla "elements" models, the face data applies to the face it is specified in or all faces of the element it is specified in which don't have their own face data.
```

docs/rendering/modelextensions/facedata.md
```diff
@@ -0,0 +1,109 @@
...
+      // ...
+    }
+  ]
+}
+```
+
+Generated Item Model
+--------------------
+
+In generated item models loaded through the `forge:item_layers` loader the face data applies to all faces generated from the texture specified for the layer the respective entry is associated with.
```
Matyrobbrt:
```suggestion
In generated item models loaded through the `forge:item_layers` loader, the face data applies to all faces generated from the texture specified for the layer the respective entry is associated with.
```

docs/rendering/modelextensions/facedata.md
```diff
@@ -0,0 +1,109 @@
...
+### Color
+
+Specifying a color value with the `color` entry will apply that color as a tint to the quads. Defaults to `0xFFFFFFFF` (white, fully opaque). The color must be in the `ARGB` format packed into a 32-bit integer and can be specified as either a hexadecimal string (`"0xAARRGGBB"`) or as a decimal integer literal (JSON does not support hexadecimal integer literals).
+
+!!! warning
+    If the color is specified as a 24-bit integer (no matter the representation), the resulting color will be fully transparent due to the alpha component being 0.
+
+This can be used as a replacement for tinting with [`BlockColor` and `ItemColor`][tinting] if the color values are constant.
+
+### Block and Sky Light
```
Matyrobbrt, line 84: Skylight is usually spelled as the same word.

docs/rendering/modelextensions/facedata.md
```diff
@@ -0,0 +1,109 @@
...
+
+Specifying a block and/or sky light value with the `block_light` and `sky_light` entry respectively will override the respective light value of the quads. Both values default to 0. The values must be in the range 0-15 (inclusive) and are treated as a minimum value for the respective light type when the face is rendered meaning that a higher in-world value of the respective light type will override the specified value.
```
Matyrobbrt:
```suggestion
Specifying a block and/or sky light value with the `block_light` and `sky_light` entry respectively will override the respective light value of the quads. Both values default to 0. The values must be in the range 0-15 (inclusive) and are treated as a minimum value for the respective light type when the face is rendered, meaning that a higher in-world value of the respective light type will override the specified value.
```

Matyrobbrt (changes requested):
Mainly grammar and punctuation nitpicks.

docs/rendering/modelextensions/visibility.md
XFactHD: I'm not sure about this change. The point is that the models using the composite model as their parent are specifying the visibilities, not the composite model itself. To me, that additional comma would mean that the composite model itself specifies the visibilities. Maybe the way that sentence is written at the moment is just confusing.

docs/rendering/modelextensions/facedata.md
XFactHD: Everything I can find in the game code writes it as two words. It would also look weird to have two different ways of spelling the two light types.

docs/rendering/modelextensions/transforms.md
XFactHD: Opted for a dash instead of a space

docs/rendering/modelextensions/facedata.md
Matyrobbrt: Fair enough

Matyrobbrt (approved):

amadornes (changes requested):
I think it's important to explicitly note that the `render_type`, `visibility`, and really even `transform` are just hints, and can be ignored by the loader (for example, `render_type` isn't supported in composite models because it makes no sense).

I would also reword the warning about ARGB in face tinting. Talking about 24-bit integers feels oddly indirect when the goal is to say "be warned that the alpha bits are multiplied, so if they're zero, the face will be transparent".

It might also be good to briefly explain what ambient occlusion is and why you might want to disable it instead of just linking to the Wikipedia page and letting the reader figure it out on their own.

Other than that, looking pretty good.

docs/rendering/modelextensions/ambientocclusion.png
ChampionAsh5357, line 1: Image should probably be more clear which side has AO and which side doesn't

docs/rendering/modelextensions/rendertypes.md
```diff
@@ -0,0 +1,74 @@
...
+Adding the `render_type` entry at the top level allows specifying the render type this model will render in.  
+If this is not specified, the model will fall back to the render types returned by `ItemBlockRenderTypes#getRenderLayers()`.
+
+!!! note
+    Since 1.19 this replaces the deprecated method of setting the applicable render type(s) via `ItemBlockRenderTypes#setRenderLayer()` for blocks.
+
+Example of a cutout block (i.e. a Glass Block)
+
+```js
+{
```
ChampionAsh5357, line 14: I would probably add a comment here saying something like `// For some model JSON`

docs/rendering/modelextensions/facedata.md
```diff
@@ -0,0 +1,109 @@
...
+          "ambient_occlusion": false
+      },
+      "faces": {
+        "north": {
+          "forge_data": {
+            "color": "0xFFFF0000",
+            "block_light": 15,
+            "sky_light": 15,
+            "ambient_occlusion": false
+          },
```
ChampionAsh5357: Are the face and top model specification an overwrite or is it merged from the parent if a field is not defined?

docs/rendering/modelextensions/facedata.md
```diff
@@ -0,0 +1,109 @@
...
+Specifying a block and/or sky light value with the `block_light` and `sky_light` entry respectively will override the respective light value of the quads. Both values default to 0. The values must be in the range 0-15 (inclusive) and are treated as a minimum value for the respective light type when the face is rendered, meaning that a higher in-world value of the respective light type will override the specified value.
+
+The specified light values are purely client-side and affect neither the server's light level nor the brightness of surrounding blocks.
+
+### Ambient Occlusion
+
+Specifying the `ambient_occlusion` flag will configure [AO] for the quads. Defaults to `true`. The behaviour of this flag is equivalent to the top-level `ambientocclusion` flag of the vanilla format.
+
+![Ambient occlusion in action](./ambientocclusion.png)  
+*Ambient occlusion enabled on the left and disabled on the right, demonstrated with the Smooth Lighting graphics setting*
```
ChampionAsh5357: I see you have the comment here, but directly in the image would be better.

docs/rendering/modelextensions/facedata.md
```diff
@@ -0,0 +1,109 @@
...
+
+!!! note
+    If the top-level AO flag is set to false, specifying this flag as true on an element or face won't be able to override the top-level flag.
+    ```js
+    {
+      "ambientocclusion": false,
```
ChampionAsh5357, line 101: I'm guessing this is supposed to be `ambient_occlusion`

docs/rendering/modelextensions/transforms.md
```diff
@@ -0,0 +1,67 @@
+Root Transforms
+===============
+
+Adding the `transform` entry at the top level of a model JSON allows specifying a transformation that will be baked into the vertex data of all quads of the model. These transforms apply before the transforms specified in the [blockstate] file of a block model and the [display transforms][displaytransform] of an item model.
+
+The root transforms can be specified in four formats:
+
+1. A string referring to a default value (currently only supports the value `"identity"`)
+2. A raw transformation matrix in the form of a nested JSON array with the last row omitted (3*4 matrix, row major order)
+3. A JSON object containing a singular `matrix` entry with the same format as above
```
ChampionAsh5357: I would still like an example for these as it not generally clear how this would look:

```js
{
    "transform": [
       [4, 5, 6], // What do these numbers represent?
       // ...
    ]
}
```

docs/rendering/modelextensions/transforms.md
```diff
@@ -0,0 +1,67 @@
...
+
+### Translation
+
+The translation must be specified as an array of 3 floating point values representing a three-dimensional vector: `[ x, y, z ]` and defaults to (0, 0, 0) if not present.
+
+### Left and Right Rotation
+
+The rotations can be specified in any one of the following four ways:
+
+- Single JSON object with a single axis => rotation degree mapping: `{ "X": 90 }`
```
ChampionAsh5357: Is this supposed to be capitalized `X`

docs/rendering/modelextensions/visibility.md
```diff
@@ -0,0 +1,34 @@
...
+      "parent": "mymod:mypartmodel_two"
+    }
+  },
+  "visibility": {
+    "part_two": false
+  }
+}
+```
+
+The visibility of a given part is determined by checking whether the model specifies a visibility for this part and, if not present, recursively checking the model's parent until either an entry is found or there is no further parent to check, in which case it defaults to true.
```
ChampionAsh5357, line 42: I would probably provide an example here where the first part is disabled, the second child is enabled, and the third child does not specify anything.

ChampionAsh5357 (changes requested):

docs/rendering/modelextensions/ambientocclusion.png
XFactHD: Added annotations to the image itself (I've kept the original image as-is in case I need to go back to it)

docs/rendering/modelextensions/rendertypes.md
XFactHD: I've opted to make the example a simple but fully functional model

docs/rendering/modelextensions/facedata.md
XFactHD: Added a comment detailing the override behaviour

docs/rendering/modelextensions/facedata.md
XFactHD: Nope, the vanilla key is written without an underscore: https://minecraft.fandom.com/wiki/Tutorials/Models#Block_models

docs/rendering/modelextensions/transforms.md
XFactHD: Nice catch, supposed to be lowercase

docs/rendering/modelextensions/visibility.md
XFactHD: Expanded the example above, let me know if that's what you were looking for

docs/rendering/modelextensions/transforms.md
```diff
@@ -0,0 +1,76 @@
...
+1. A JSON object containing a singular `matrix` entry containing a raw transformation matrix in the form of a nested JSON array with the last row omitted (3*4 matrix, row major order). Example to demonstrate the structure:
+   ```js
+   "transform": {
+       "matrix": [
+           [ 0, 0, 0, 0 ],
+           [ 0, 0, 0, 0 ],
+           [ 0, 0, 0, 0 ]
+       ]
+   }
+   ```
```
ChampionAsh5357: Should be indented 4 spaces

docs/rendering/modelextensions/transforms.md
```diff
@@ -0,0 +1,76 @@
...
+2. A JSON object containing any combination of the following optional entries:
+    - `origin`: origin point used for the rotations and scaling
+    - `translation`: relative translation
+    - `rotation` or `left_rotation`: rotation around the translated origin to be applied before scaling
+    - `scale`: scale relative to the translated origin
+    - `right_rotation` or `post_rotation`: rotation around the translated origin to be applied after scaling
```
ChampionAsh5357: Use `*` instead of `-`

ChampionAsh5357 (changes requested):
I think this looks good now. Took a quick double check on the formats and it's basically just make sure list indents are 4 spaces and that unordered lists use `*` instead of `-`.

Other than that, I'm gonna give @amadornes until Monday, the 15th to see if they have anything else to add. Otherwise, I'm just gonna merge it.

docs/rendering/modelextensions/transforms.md
XFactHD: I've added an example purely to demonstrate the expected structure. At the moment, I can't explain what the individual values in the matrix represent as I am failing to even find a simple matrix that works at all. While searching one, I may have found a severe bug in this implementation (apart from two of the transform formats turning out to not be usable at the moment) but I'll need to confirm that with further tests.

docs/rendering/modelextensions/facedata.md
```diff
@@ -0,0 +1,112 @@
...
+
+All values of the face data are optional.
+
+Elements Model
+--------------
+
+In vanilla "elements" models, the face data applies to the face it is specified in or all faces of the element it is specified in which don't have their own face data.
+
+!!!note
+    The `forge_data` specified on a specific face completely overrides the `forge_data` specified on the whole element for that face and will not inherit any parameters from the element-level declaration.
```
amadornes: I think the following would be simpler and clearer, and avoids unnecessary repetition:

```suggestion
    If `forge_data` is specified on a face, it will not inherit any parameters from the element-level `forge_data` declaration.
```

docs/rendering/modelextensions/facedata.md
```diff
@@ -0,0 +1,112 @@
...
+  ]
+}
+```
+
+Generated Item Model
+--------------------
+
+In generated item models loaded through the `forge:item_layers` loader, the face data applies to all faces generated from the texture specified for the layer the respective entry is associated with.
+
+The additional data has to be specified at the top level with the entry names in the `forge_data` object being the layer index the respective entry belongs to (in the example, layer 1 will be tinted red and glow at full brightness):
```
amadornes:
```suggestion
In item models generated using the `forge:item_layers` loader, face data is specified for each texture layer and applies to all of the geometry (front/back facing quads and edge quads).

The `forge_data` field must be located at the top level of the model JSON, with each key-value pair associating a face data object to a layer index.

In the following example, layer 1 will be tinted red and glow at full brightness:
```

docs/rendering/modelextensions/index.md
```diff
@@ -0,0 +1,16 @@
+Model Extensions
+================
+
+Forge provides several extensions to the model JSON format to allow fine tuning the way JSON models are loaded, baked and rendered:
```
amadornes:
```suggestion
Forge provides several extensions to vanilla's JSON model format to allow fine tuning of loading, baking and rendering behavior:
```

docs/rendering/modelextensions/rendertypes.md
```diff
@@ -0,0 +1,78 @@
+Render Types
+============
+
+Adding the `render_type` entry at the top level allows specifying the render type this model will render in.  
+If this is not specified, the model will fall back to the render types returned by `ItemBlockRenderTypes#getRenderLayers()`.  
+This is merely a hint to the model loader and custom model loaders are free to ignore this entry.
```
amadornes:
```suggestion
Adding the `render_type` entry at the top level of the JSON suggests to the loader what render type the model should use. If not specified, the loader gets to pick the render type(s) used, often falling back to the render types returned by `ItemBlockRenderTypes#getRenderLayers()`.

Custom model loaders may ignore this field entirely.
```

docs/rendering/modelextensions/rendertypes.md
```diff
@@ -0,0 +1,78 @@
...
+    * Entity render type: `ForgeRenderTypes#ITEM_LAYERED_CUTOUT_MIPPED`
+    * Used in similar cases as `minecraft:cutout_mipped` when the item representation should also have mipmapping applied
+* `minecraft:translucent`
+    * Chunk render type: `RenderType#translucent()`
+    * Entity render type: `ForgeRenderTypes#ITEM_LAYERED_TRANSLUCENT`
+    * Used for blocks where any given pixel may be partially transparent (i.e. Stained Glass)
+* `minecraft:tripwire`
+    * Chunk render type: `RenderType#tripwire()`
+    * Entity render type: `ForgeRenderTypes#ITEM_LAYERED_TRANSLUCENT`
+    * Chunk and entity render type differ due to the tripwire render type not being feasible as an entity render type
```
amadornes, line 52: Why specify it here but nowhere else? 🤔  
It's the same reason for all of them at the end of the day.

docs/rendering/modelextensions/rendertypes.md
```diff
@@ -0,0 +1,78 @@
...
+The optional entity render type is used the same way as the required entity render type when the *Fabulous!* graphics mode is selected. This render type is needed in cases where the required entity render type does not work in the *Fabulous!* graphics mode (typically only applies to translucent render types).
+
+```java
+public static void onRegisterNamedRenderTypes(RegisterNamedRenderTypesEvent event)
+{
+  event.register("special_cutout", RenderType.cutout(), Sheets.cutoutBlockSheet());
+  event.register("special_translucent", RenderType.translucent(), Sheets.translucentCullBlockSheet(), Sheets.translucentItemSheet());
+}
+```
+
```
amadornes, line 77:
```suggestion

These can then be addressed in JSON as `<your_mod_id>:special_cutout` and `<your_mod_id>:special_translucent`.

```

docs/rendering/modelextensions/transforms.md
```diff
@@ -0,0 +1,76 @@
+Root Transforms
+===============
+
+Adding the `transform` entry at the top level of a model JSON allows specifying a transformation that will be baked into the vertex data of all quads of the model. These transforms apply before the transforms specified in the [blockstate] file of a block model and the [display transforms][displaytransform] of an item model.
+
+Custom model loaders are given these transforms combined with the [blockstate] transforms or [display transforms][displaytransform] in the form of a `ModelState` in `IGeometryLoader#bake()` and are free to ignore them.
```
amadornes:
```suggestion
Adding the `transform` entry at the top level of a model JSON suggests to the loader that a transformation should be applied to all geometry right before the rotations in the [blockstate] file in the case of a block model, and before the [display transforms][displaytransform] in the case of an item model.

Custom model loaders may ignore this field entirely.
```

amadornes (changes requested):

docs/rendering/modelextensions/rendertypes.md
XFactHD: In all cases except the two I mention it specifically (the referenced one being one of those), the entity render type is the direct entity-equivalent of the chunk render type, but in these two cases it's a different one for a specific reason in case of the cutout_mipped render type and for a reason I have to assume in case of the tripwire

docs/rendering/modelextensions/ambientocclusion.png
ChampionAsh5357, line 1: You can delete this if it goes unused.

docs/rendering/modelextensions/facedata.md
```diff
@@ -0,0 +1,114 @@
...
+
+Specifying a block and/or sky light value with the `block_light` and `sky_light` entry respectively will override the respective light value of the quads. Both values default to 0. The values must be in the range 0-15 (inclusive) and are treated as a minimum value for the respective light type when the face is rendered, meaning that a higher in-world value of the respective light type will override the specified value.
+
+The specified light values are purely client-side and affect neither the server's light level nor the brightness of surrounding blocks.
+
+### Ambient Occlusion
+
+Specifying the `ambient_occlusion` flag will configure [AO] for the quads. Defaults to `true`. The behaviour of this flag is equivalent to the top-level `ambientocclusion` flag of the vanilla format.
+
+![Ambient occlusion in action](./ambientocclusion_annotated.png)
```
ChampionAsh5357: Reference already in `[ao_img]` below but not associated.

ChampionAsh5357 (changes requested):
More random fixes found, but should be the last if @amadornes has no more comments.

docs/rendering/modelextensions/facedata.md
```diff
@@ -0,0 +1,114 @@
...
+The additional data can be specified in the two ways shown in this example:
+```js
+{
+  "elements": [
+    {
+      "forge_data": {
+        "color": "0xFFFF0000",
+          "block_light": 15,
+          "sky_light": 15,
+          "ambient_occlusion": false
```
Matyrobbrt:
```suggestion
        "block_light": 15,
        "sky_light": 15,
        "ambient_occlusion": false
```

docs/rendering/modelextensions/facedata.md
```diff
@@ -0,0 +1,114 @@
...
+
+Parameters
+----------
+
+### Color
+
+Specifying a color value with the `color` entry will apply that color as a tint to the quads. Defaults to `0xFFFFFFFF` (white, fully opaque). The color must be in the `ARGB` format packed into a 32-bit integer and can be specified as either a hexadecimal string (`"0xAARRGGBB"`) or as a decimal integer literal (JSON does not support hexadecimal integer literals).
+
+!!! warning
+    The four color components are multiplied with the texture's pixels. If the alpha component is ommited, the quad will render fully transparent.
```
amadornes:
```suggestion
    The four color components are multiplied with the texture's pixels. Omitting the alpha component is equivalent to making it 0, which will make the geometry fully transparent.
```

amadornes (approved):
Other than that minor nitpick about alpha multiplication, lgtm

ChampionAsh5357 (approved):

PR #499, opened 2023-06-07
[1.19.x] Change Versions Displayed on Docs

ChampionAsh5357:
This *would* change the version to latest and LTS, but as we don't know what those are yet, I'm just assuming 1.19.x and 1.20.x until there is a decision.


ChampionAsh5357:
Handled in doc fixes.

PR #504, opened 2023-07-15
[1.19.x] improve an expression within items/index.md

srcres258:
The expression "An item(s)" is invalid on grammar when it refers to the plural meaning at which it turns into "An items". To avoid the possibly vague thoughts for some readers about this, change this into "An item or multiple items", towards which readers are able to comprehend both grammatically and misunderstanding-avoidingly.

### DIFF
--- a/docs/items/index.md
+++ b/docs/items/index.md
@@ -26,7 +26,7 @@ Setting the properties of an item as above only works for simple items. If you w
 
 ## `CreativeModeTabEvent`
 
-An item can be added to a `CreativeModeTab` via `CreativeModeTabEvent$BuildContents` on the [mod event bus][modbus]. An item(s) can be added without any additional configurations via `#accept`.
+An item can be added to a `CreativeModeTab` via `CreativeModeTabEvent$BuildContents` on the [mod event bus][modbus]. An item or multiple items can be added without any additional configurations via `#accept`.
 
 ```java
 // Registered on the MOD event bus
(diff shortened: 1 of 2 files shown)


docs/items/index.md
```diff
@@ -26,7 +26,7 @@ Setting the properties of an item as above only works for simple items. If you w
 
 ## `CreativeModeTabEvent`
 
-An item can be added to a `CreativeModeTab` via `CreativeModeTabEvent$BuildContents` on the [mod event bus][modbus]. An item(s) can be added without any additional configurations via `#accept`.
+An item can be added to a `CreativeModeTab` via `CreativeModeTabEvent$BuildContents` on the [mod event bus][modbus]. An item or multiple items can be added without any additional configurations via `#accept`.
```
ChampionAsh5357:
```suggestion
An item can be added to a `CreativeModeTab` via `CreativeModeTabEvent$BuildContents` on the [mod event bus][modbus]. Items can be added without any additional configurations via `#accept`.
```

It makes more sense to use the plural without the article. Otherwise, the sentence flows strangely.

ChampionAsh5357 (changes requested):

docs/items/index.md
srcres258: Applied. Thanks.

srcres258 (commented):
Applied.

ChampionAsh5357 (approved):

PR #508, opened 2023-07-19
[1.20.x] fix a numeric typo

srcres258:
The count of the events behind the number is three instead of two. Hence turn the figure "two" into "three".


ChampionAsh5357 (approved):
Counting is a useful skill.

PR #510, opened 2023-07-22
1.19.x: fix a typo

srcres258:
The "width" here should be the "height" as the field's name is "imageHeight" rather than "imageWidth".


ChampionAsh5357 (approved):
Whoops, thanks again.

PR #511, opened 2023-07-22
1.19.x: Append the missing full point at the end of some sentences

srcres258:
There were some sentences whose full stops were ignored, consequently add them back.


ChampionAsh5357 (approved):
I'm running out of witty approvals, so I'll just approve.

PR #513, opened 2023-08-06
Added information related to creating a BakedQuad.

NickWrecks:


NickWrecks:
Made a mistake by typing: "Any value greater than 1 will position a given vertex outside of the block space". Obviously this is nonsense. It was meant to say "Giving a vertex coordonate any value greater than 1 will position it outside of the block space"

PR #528, opened 2024-02-14
Replace refrences to removed NetworkHooks#openMenu with ServerPlayer#openMenu

JulianVennen:
The documentation on menus still mentions that you should use NetworkHooks#openMenu.
The NetworkHooks class however no longer exists in Forge 1.20.2+ and ServerPlayer#openMenu should be used instead.


PaintNinja (approved):

ChampionAsh5357 (approved):
Thanks!

PR #541, opened 2025-09-09
Add info about jarJar task and -all jar

Asbestosstar:


LexManos:
Id say add something to the top about it not being built when you run "build" and that you want to use the -all jar

Jonathing:
Right, the `jarJar` task really *should be* a dependent of the `assemble` task, since that's how that works. Run the `help` task on any Java project to see for yourself.
<https://docs.gradle.org/current/userguide/tutorial_using_tasks.html#task_classification>

My bet is that people are already using weird setups for FG6's Jar-in-Jar, so I'm not going to change anything with that. But for the new plugin I'm working on: the `jarJar` task will be a dependent of `assemble`, which itself is a dependent of `build`.

LexManos:
I _think_ the reason the jarJar task isn't a dependency of assemble is because it defaults to being disabled. Is gradle fine with having disabled dependents?

But either way this is outside the scope of this PR.
My request was to make it clear that the jarJar task needed to be run in the documentation to reflect the current state. So ya, please make it more clear in the first line or two at the top of the page.

Jonathing:
Even when enabled, it isn't a dependent of assemble.

Asbestosstar:
I have made changes to add the content to the top.

LexManos:
Moved it to the correct place on my phone.

